PR #3, opened 2023-04-01
single-arm64 image.

jacksonrnewhouse:
This brings us to parity between x86 and arm64. I disabled wasm-opt as it doesn't support arm64, per this ticket:  https://github.com/WebAssembly/binaryen/issues/5337.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@ you need to get a test Arroyo system up and running.
 ```bash
 $ docker run -p 8000:8000 -p 8001:8001 ghcr.io/arroyosystems/arroyo:single
 ```
-
+If you're on an arm64 machine, instead use ghcr.io/arroyosystems/arroyo:single-arm64.
 This will run all of the core arroyo services and a temporary Postgres database. The web UI will be exposed at
 http://localhost:8000.
 
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,1 +1,3 @@
 target
+docker/single/Dockerfile.*
+arroyo-console/node_modules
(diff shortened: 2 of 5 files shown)


README.md
```diff
@@ -56,7 +56,7 @@ $ docker run -p 8000:8000 -p 8001:8001 ghcr.io/arroyosystems/arroyo:single
 ```
 
 This will run all of the core arroyo services and a temporary Postgres database. The web UI will be exposed at
-http://localhost:8000.
+http://localhost:8000. If you're on an arm64 machine, instead use ghcr.io/arroyosystems/arroyo:single-arm64.
```
mwylde, line 59: Can you move this before the previous line, so that it's right after the bash statement?

This is great for now, but we'll probably want to do a proper multi-arch image (https://www.docker.com/blog/multi-arch-build-and-images-the-simple-way/) so that users don't need to pick the right one.

docker/single/Dockerfile.arm64
```diff
@@ -0,0 +1,55 @@
+FROM ubuntu:22.04
```
mwylde, line 1: Is it necessary to duplicate all of this, or would it be possible to use a build arg to combine both into a single Dockerfile?

jacksonrnewhouse: There are two packages we download that are architecture specific. I think there might be a way to do it with build args.

mwylde (dismissed):

docker/single/Dockerfile.arm64
jacksonrnewhouse: Fixed this.

mwylde (approved):

mikeyasaservice:
Already exists - 23 Rust crates in /crates/

PR #6, opened 2023-04-03
Package upgrades and Clippy suggestions

jacksonrnewhouse:


arroyo-controller/Cargo.toml
```diff
@@ -16,10 +16,10 @@ tonic = "0.8"
 tonic-reflection = "0.5"
 prost = "0.11"
 tokio = { version = "1", features = ["full"] }
-tokio-stream = "0.1.11"
+tokio-stream = "0.1.12"
 tonic-web = "0.4.0"
 rand = "0.8"
-bincode = { version = "2.0.0-rc.1", features = ["serde"]}
+bincode = { version = "2.0.0-rc.3", features = ["serde"]}
```
mwylde, line 22: Did you test that this is safe? We've hit issues before upgrading RCs of bincode

mwylde (approved):

arroyo-controller/Cargo.toml
jacksonrnewhouse: Yeah, I ran a copy locally. That's why I had to bump the versions in the files generated by compiler.rs.

mikeyasaservice:
Already exists - /docker/ directory

PR #7, opened 2023-04-03
Add sql expression testing via macros.

jacksonrnewhouse:
This lets us make assertions on the compiled code. The basic inputs are an expression string, a TestStruct, and a result.

Unfortunately I needed to introduce two more crates to get this all working. If there's some way to avoid that I'd love to hear.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -601,6 +601,25 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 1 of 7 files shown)


mwylde (approved):
Unfortunately I think the two crates are necessary unless there's a way to implement the macro without needing any arroyo-sql code.

PR #11, opened 2023-04-04
add arroyo-compiler-service and build multi-arch

jacksonrnewhouse:


docker/single/Dockerfile
```diff
@@ -37,21 +37,48 @@ RUN  curl -fsSL https://deb.nodesource.com/setup_19.x | sudo -E bash - && \
     sudo apt-get install -y nodejs && \
     curl -fsSL https://get.pnpm.io/install.sh | SHELL=$(which bash) bash -
 
-COPY . .
+COPY arroyo-types arroyo-types
```
mwylde, line 40: Is there a reason we need to list out all of the crates here?

jacksonrnewhouse: If you just do `COPY . .` the entire repo ends up in your docker container, which can make it hard to iterate as you rebuild everything after that each time. I mainly changed it for iteration speed reasons, so fine with going back to `COPY . .`.

mwylde (approved):

PR #16, opened 2023-04-04
Add prometheus and pushgateway into the docker image to get metrics working

mwylde:


docker/single/Dockerfile
```diff
@@ -4,6 +4,7 @@ ENV DEBIAN_FRONTEND=noninteractive
 WORKDIR /opt/arroyo/src
 ARG MOLD_ARCH
 ARG PROTO_ARCH
+ARG PROM_ARCH
```
jacksonrnewhouse, line 7: Might be worth finding some way to have one architecture arg that we convert to the various names.

jacksonrnewhouse (approved):

docker/single/Dockerfile
mwylde: Agreed, should be possible with a small bash script?

PR #20, opened 2023-04-10
Add SASL authentication support to kafka connections.

jacksonrnewhouse:
Would appreciate feedback on 
1. My TypeScript.
2. Which structs/types to use for passing these parameters. The conversion to a list of pairs is kinda ugly, but makes codegen easier.

I've tested this by successfully reading and writing to a SASL_SSL cluster.


arroyo-console/src/routes/connections/CreateConnection.tsx
```diff
@@ -88,6 +159,57 @@ function ConfigureKafka({
...
       </FormControl>
+      <FormControl isRequired>
+  <FormLabel>Authentication Type</FormLabel>
+  <Select value={config.authConfig?.authType.case || ""} onChange={onChangeAuthType}>
+    <option value="">Select an authentication type</option>
+    <option value="noAuth">No Auth</option>
+    <option value="saslAuth">SASL Auth</option>
+  </Select>
+  {config.authConfig?.authType.case === "saslAuth" && (
+  <Stack spacing={5}>
```
mwylde, line 170: This should be indented a level so that it's clear it's part of the conditional

mwylde: This is also incorrectly nested under the FormControl; a FormControl should have only one child (visually this leads to the spacing being off)

mwylde (changes requested):
Overall looks good. Some comments about the JSX -- it might help to take a look at some of the other complex forms like the ConfigureSources stuff.

Agree that the list of connection config options is kind of ugly, but I think we'll appreciate the flexibility of that approach going forward, as it means we don't need to change the pipeline representation or code generation to add new options.

arroyo-api/Cargo.toml
```diff
@@ -55,7 +55,7 @@ syn = {version = "1", features = ["full"]}
 quote = "1"
 proc-macro2 = "1"
 # source testing
-rdkafka = { version = "0.28", features = ["cmake-build"] }
+rdkafka = { version = "0.28", features = ["cmake-build",  "ssl-vendored", "sasl"] }
```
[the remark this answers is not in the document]
jacksonrnewhouse: Did this in a second commit, let me know if you think it is worth the effort.

arroyo-console/src/routes/connections/CreateConnection.tsx
jacksonrnewhouse: I think I fixed this.

arroyo-console/src/routes/connections/CreateConnection.tsx
```diff
@@ -88,6 +159,57 @@ function ConfigureKafka({
         <Input type="text" value={config.bootstrapServers} onChange={onChange} />
         <FormHelperText>Comma-separated list of kafka brokers to connect to</FormHelperText>
       </FormControl>
+      <FormControl isRequired>
+  <FormLabel>Authentication Type</FormLabel>
```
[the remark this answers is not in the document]
jacksonrnewhouse: fixed

mwylde (approved):

PR #38, opened 2023-04-13
better S3 environment variable names.

jacksonrnewhouse:
compiler and state should share a region, but have different bucket variables.


mwylde:
Is there a reason to use separate variables? You can still configure them indepdently because they are different services. It might also be nice to move all of these env vars to arroyo-types so that we're not duplicating them, and it's easy for users to see the full set of available vars.

PR #49, opened 2023-04-17
Implement Hashing Functions

### LINKED ISSUE #30: Implement DataFusion Hashing Functions
DataFusion [supports](https://arrow.apache.org/datafusion/user-guide/sql/scalar_functions.html#hashing-functions) md5, sha224, sha256, sha384, sha512.

akennedy4155:
Closes #30 

---

Implements functions for hashing:
* md5
* sha224
* sha256
* sha384
* sha512

---

* Added `HashExpression` and `HashFunction` struct to `expressions.rs`.  
* Added the hash and encoding libs to `arroyo-types`
  * base64 (for encoding the hash digest to str)
  * md-5 (rust-crypto)
  * sha2 (rust-crypto)
* Added module `arroyo-types::functions`
  * mv `functions.rs` (used to contain str functions) to `arroyo-types::functions::strings`
  * Add `arroyo-types::functions::hash`


arroyo-sql/src/expressions.rs
```diff
@@ -1195,7 +1202,77 @@ pub enum StringFunction {
     Rpad(Box<Expression>, Box<Expression>, Option<Box<Expression>>),
     Rtrim(Box<Expression>, Option<Box<Expression>>),
 }
+// postgres has to be active for build for some reason?
```
akennedy4155, line 1205: rm this comment

arroyo-sql/src/expressions.rs
```diff
@@ -1195,7 +1202,77 @@ pub enum StringFunction {
...
+            }
+        })
+    }
+
+    fn return_type(&self) -> TypeDef {
+        // this *can* be null because in SQL - MD5(NULL) = NULL
+        TypeDef::DataType(DataType::Utf8, self.input.nullable())
+    }
+}
+// I'm going to start adding comments as I go
```
akennedy4155, line 1275: rm comment

arroyo-types/src/functions/hash.rs
```diff
@@ -0,0 +1,45 @@
+use base64::{engine::general_purpose::STANDARD_NO_PAD as encoder, Engine};
+use md5::{Digest, Md5};
+
+// would like to make this parameterized based on the hasher, since everything else looks the same
+// but I couldn't quite figure out what the common trait was between all of them...
```
akennedy4155, line 5: I thought about how to parametrize this:

```
fn hash(argument: String, hasher: ?) -> String {
    hasher.update(argument.as_bytes());
    let res = hasher.finalize();
    encoder.encode(&res)
}

pub fn md5(argument: String) -> String {
  hash(argument, Md5::new())
}

pub fn sha224(argument:String) -> String {
  hash(argument, Sha224::new())
}

etc.
```

But I couldn't quite figure out the right type for the hasher argument.  The `rust-crypto` hashing libs use something like `CoreWrapper<...>` and after messing around with some dyn/impl types, I just decided to have some repeated code for the sake of getting this PR out and gathering feedback.

arroyo-types/src/functions/hash.rs
```diff
@@ -0,0 +1,45 @@
...
+// I'll ask later if anyone has ideas.
+pub fn md5(argument: String) -> String {
+    let mut hasher = Md5::new();
+    hasher.update(argument.as_bytes());
+    let res = hasher.finalize();
+    // chose to encode as base64 to encode bytes to string, but not sure if that's correct
```
akennedy4155, line 11: Is `base64` str encoding the right choice here? utf-8 didn't make sense to me because it's not guaranteed that the hash is going to be a utf char.

akennedy4155 (commented):
Is there a way to document this as I go?  Is it worth going through all of my changes and adding docs just to get that process moving?

arroyo-types/src/functions/hash.rs
jacksonrnewhouse, line 11: From the [postgres docs](https://www.postgresql.org/docs/11/functions-binarystring.html), md5 is a hex-encoded text representation, while the shaXXX functions return a byte array.

> Note that for historic reasons, the function md5 returns a hex-encoded value of type text whereas the SHA-2 functions return type bytea. Use the functions encode and decode to convert between the two, for example encode(sha256('abc'), 'hex') to get a hex-encoded text representation.

DataFusion seems to follow this convention, so it would probably be good if we did as well.

arroyo-sql/src/expressions.rs
```diff
@@ -1195,7 +1202,77 @@ pub enum StringFunction {
...
+            input,
+        }))
+    }
+}
+
+impl ExpressionGenerator for HashExpression {
+    fn to_syn_expression(&self) -> syn::Expr {
+        let input = self.input.to_syn_expression();
+        let hash_fn = format_ident!("{}", self.function.to_string());
+        parse_quote!({
```
jacksonrnewhouse, line 1262: This is only correct if self.input.is_nullable(). Take a look at some of the other single-argument methods for how to handle the two cases.

arroyo-types/src/functions/hash.rs
jacksonrnewhouse: I answered this in-line, but it seems like md5 should be a hex-coded String and the others should return a DataType::Binary, which I'd presume will be a Vec<u8>.

jacksonrnewhouse:
For testing, https://github.com/ArroyoSystems/arroyo/blob/c430f37abf73774204699284ff9792b3e7b1fc26/arroyo-sql-testing/src/lib.rs#L7 lets you write a macro-based test. You can use psql to get your assertions, e.g. 

```
    single_test_codegen!(
        "md5",
        "md5(non_nullable_string)",
        arroyo_sql::TestStruct {
            non_nullable_string: "test".into(),
            ..Default::default()
        },
        "098f6bcd4621d373cade4e832627b4f6".to_string()
    );

File: arroyo-types/src/functions/hash.rs (hunk 4 above)
Comment (akennedy4155) [reply arroyo-types/src/functions/hash.rs]: Yep, looks like the argument type for sha functions is also not right according to postgres docs.  DataType::Binary should be the argument and the return, I believe.



sha256(bytea) | bytea | SHA-256 hash | sha256('abc') | \xba7816bf8f01cfea414140de5dae2223​b00361a396177a9cb410ff61f20015ad
-- | -- | -- | -- | --

File: arroyo-types/Cargo.toml [hunk 6]
```diff
@@ -5,4 +5,7 @@ edition = "2021"
 
 [dependencies]
 bincode = "2.0.0-rc.3"
+base64 = "0.21"
+md-5 = "0.10"
+sha2 = "0.10"
```
akennedy4155, line 10: Usually I leave these unpinned on the minor version, in order to stay up to date with patches, but I've seen it both ways.  What do we prefer here?

akennedy4155:
Any preferences on unit testing vs speed of changes?  I could write tests for the hash functions, but it might take longer to get merged.  I lean toward lower velocity, improved correctness.

arroyo-types/Cargo.toml
jacksonrnewhouse: Yeah, that's fine.

arroyo-sql/src/expressions.rs
akennedy4155: Should be better now with the check on `nullable()`

jacksonrnewhouse:
The unit tests in arroyo-sql-testing are meant to be pretty light-weight, both in terms of effort to write them and how quickly they run, so when we add new functions to the SQL integration I'd like to see a test or two per function.

arroyo-types/src/functions/hash.rs
akennedy4155: Adjusted

jacksonrnewhouse:
Looking pretty good, just opened a small PR with some touchups. https://github.com/akennedy4155/arroyo/pull/1

akennedy4155:
Good to go!

Issue #22, opened 2023-04-11
Controller continues in degraded state when psql connection fails


mwylde:
On startup, the controller spawns a tokio task that watches the postgres job_configs table. This is what allows it to respond to changes in configuration, starting, stopping and modifying pipelines.

However, if that connection fails (because of a configuration issue or because the DB is not available) the task will panic, but the overall controller process will continue on. However, because that task is gone it will not do any useful work.

This should be fixed by:
1. Failing controller startup if we're unable to talk to postgres
2. Adding error handling to the config watcher task so that intermittent issues with postgres don't lead to a broken controller

mwylde:
Fixed with #50

PR #50, opened 2023-04-18
Support deploying Arroyo to a nomad cluster

mwylde:
This PR adds a number of improvements that allow an arroyo cluster to be deployed to Nomad. There is a complementary docs PR (https://github.com/ArroyoSystems/arroyo-docs/pull/7) and a new repo (https://github.com/ArroyoSystems/arroyo-nomad-pack) that contains the Nomad orchestration code.

The changes in this PR include:
* Added the ability to inject the API endpoint into the HTML for the frontend, which is needed to host the API on arbitrary hosts/ports as occurs in a cluster scheduler like Nomad
* The controller now checks for a proper postgres connection on startup, and fails if it cannot connect (addresses #22)
* The database port is now configurable
* Simplifies S3 configuration; the env vars `S3_REGION` and `S3_BUCKET` now control all usages of S3
* Richer configuration of ports for services; all services support `ADMIN_PORT` and `GRPC_PORT` configurations, and this can be overridden for each services with `{SERVICE}_{ADMIN, GRPC}_PORT`, for example `API_ADMIN_PORT`
* Added Dockerfiles for arroyo-services (which runs arroyo-api and arroyo-controller) and arroyo-compiler
* arroyo-compiler now has a batch mode that can be used to compile jobs without requiring a persistent service
*  Added [just](https://github.com/casey/just) commands for building docker images

There is also a breaking change. In order to more seamlessly support using a single bucket for artifacts and checkpoints, checkpoint are now stored in `{BUCKET}/{job_id}/checkpoints`. This will break existing jobs that expect checkpoints to be in the former location.


arroyo-compiler-service/src/main.rs
```diff
@@ -74,9 +78,54 @@ pub async fn main() {
...
+                .await
+                .expect("Failed to read file from object store")
+                .bytes()
+                .await
+                .expect("Failed to read file from object store");
+
+            let query = CompileQueryReq::decode(&*query).expect("Failed to decode query request");
+
+            let resp = service.compile(query).await.unwrap();
+            println!(
```
jacksonrnewhouse, line 106: should this be an info!()?

arroyo-controller/src/lib.rs
```diff
@@ -447,6 +448,18 @@ impl ControllerServer {
             .create_pool(Some(deadpool_postgres::Runtime::Tokio1), NoTls)
             .unwrap();
 
+        // test that the DB connection is valid
+        let _ = pool.get().await.unwrap_or_else(|e| {
+            panic!(
```
jacksonrnewhouse, line 453: Is it standard to panic in these cases, rather than either returning an error or exiting?

jacksonrnewhouse (approved):
Looks great!

arroyo-compiler-service/src/main.rs
mwylde: It's writing json to std out so that it can be consumed by whatever is calling it. If it were an info!, that would send it to the logging system instead.

arroyo-controller/src/lib.rs
mwylde: Panicing will produce an error and exit. Eventually we'll probably want clearer error handling for users, but while most of our users are going to be developers working on the system I think it's more helpful to have the context that a panic gives you (like the line).

PR #60, opened 2023-04-19
Add fn to homogenize float and integer operands in `BinaryMathExpression`

### LINKED ISSUE #59: Math operations between float (`f64`) and integer `{integer}` gives error
This does not give an error:

```
select 101.0 / 100 from nexmark_50000;
```

While this does:

```
select round(bid.price * 100) / 100 from nexmark_50000;
```

> `no implementation for f64 / {integer}`

akennedy4155:
Added a fn that turns right operand into the left's type if left is a float32 or float64, and the opposite for left if right is a float32 or float64.

Doesn't do any validation on the types that we're casting from so that the error that would come anyway from an operation like `string + float` wouldn't be polluted with casts.

---

Closes #59

### DIFF
--- a/arroyo-sql/src/expressions.rs
+++ b/arroyo-sql/src/expressions.rs
@@ -675,13 +675,41 @@ pub struct BinaryMathExpression {
 }
 
 impl BinaryMathExpression {
+    /// Coerces the operands to the same type.
+    /// 
+    /// This is needed because Rust doesn't allow math between float and integer types.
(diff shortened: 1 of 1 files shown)


jacksonrnewhouse:
I'm not sure this is the problem, actually. Still investigating.

jacksonrnewhouse:
I believe #64 will solve these issues.

akennedy4155:
Awesome, just reviewed and looks great.

PR #61, opened 2023-04-19
Windowless Joins

jacksonrnewhouse:
This changes how we handle joins whose inputs don't contain a preceding window operator. Now each record is joined against all of the existing records on the other side, emitting pairs. This only supports inner joins. I've also made several changes/improvements, including

- Standardize how Durations are quoted into a single function `duration_to_syn_expr()`.
- Fix support for impulses with a specified intervalMicros.
- Add some out-of-orderness to Nexmark.
- Set the default allowed-lateness for SQL pipelines to 1 second, rather than 0.

### DIFF
--- a/arroyo-worker/src/operators/join_with_expiration.rs
+++ b/arroyo-worker/src/operators/join_with_expiration.rs
@@ -1,0 +1,124 @@
+use std::{marker::PhantomData, time::Duration};
+
+use arroyo_macro::{co_process_fn, StreamNode};
+use arroyo_rpc::grpc::{TableDeleteBehavior, TableDescriptor, TableType, TableWriteBehavior};
+use arroyo_state::tables::KeyTimeMultiMap;
+use arroyo_types::*;
+
+use crate::engine::Context;
+
+#[derive(StreamNode)]
+pub struct JoinWithExpiration<K: Key, T1: Data, T2: Data> {
+    left_expiration: Duration,
+    right_expiration: Duration,
+    _t: PhantomData<(K, T1, T2)>,
+}
+
+#[co_process_fn(in_k1=K, in_t1=T1, in_k2=K, in_t2=T2, out_k=K, out_t=(T1,T2))]
+impl<K: Key, T1: Data, T2: Data> JoinWithExpiration<K, T1, T2> {
+    fn name(&self) -> String {
+        "JoinWithExperiation".to_string()
+    }
+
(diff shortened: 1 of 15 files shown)


akennedy4155:
Looks like this addresses #65 as well.

arroyo-worker/src/operators/join_with_expiration.rs
```diff
@@ -0,0 +1,124 @@
...
+            if record.timestamp < watermark {
+                return;
+            }
+        };
+        let mut right_state: KeyTimeMultiMap<K, T2, _> =
+            ctx.state.get_key_time_multi_map('r').await;
+        let mut key = record.key.clone().unwrap();
+        let value = record.value.clone();
+        let records = {
+            let right_rows = right_state.get_all_values_with_timestamps(&mut key).await;
```
mwylde, line 63: get_all_values_with_timestamps materializes the data into a vec, which is somewhat wasteful since all we're going to do is iterate over it. Instead it could return an iterator to avoid that extra allocation.

arroyo-worker/src/operators/join_with_expiration.rs
```diff
@@ -0,0 +1,124 @@
...
+            let mut records = vec![];
+            for (timestamp, right_value) in right_rows {
+                records.push(Record {
+                    timestamp: record.timestamp.max(timestamp),
+                    key: Some(key.clone()),
+                    value: (value.clone(), right_value.clone()),
```
mwylde, line 69: do we need to clone right_value?

arroyo-worker/src/operators/join_with_expiration.rs
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    async fn process_right(&mut self, record: &Record<K, T2>, ctx: &mut Context<K, (T1, T2)>) {
+        if let Some(watermark) = ctx.watermark() {
+            if record.timestamp < watermark {
+                return;
+            }
+        };
+
+        let mut left_state: KeyTimeMultiMap<K, T1, _> = ctx.state.get_key_time_multi_map('l').await;
```
mwylde, line 88: Is there any way to share this logic between left/right?

mwylde (approved):

arroyo-worker/src/operators/join_with_expiration.rs
jacksonrnewhouse: Yeah, it's a pointer at this point.

Issue #63, opened 2023-04-19
Float Literals that are integer-valued generated as integers


jacksonrnewhouse:
This test passes, but shouldn't
```
    single_test_codegen!(
        "mixed_type_addition",
        "100.0",
        arroyo_sql::TestStruct {
            ..Default::default()
        },
        100u64
    );
```

And the correct version where the right term is 100f64 fails. This is the root cause of #59. @alex-astronomer.

PR #64, opened 2023-04-19
fix the quoting of types in the sql package.

jacksonrnewhouse:
This addresses #59/ #63. We were relying on Rust to figure out the types of numerical literals, which isn't actually reliable. Now the code generated for literals includes their types, so that `100.0` becomes `100f64` instead of `100`. This meant some of the string functions needed their signatures to be changed on either the arguments or return type. This should be consistent with the signatures defined in [datafusion](https://github.com/apache/arrow-datafusion/blob/9798fbcab1e04123f05e7692475e104310e0473a/datafusion/expr/src/function.rs). DataFusion splits the determination into two parts:

The arguments: DataFusion refers to this generically as the "Signature". There are possibly multiple valid signatures for a given function. StrPos's is [here](https://github.com/apache/arrow-datafusion/blob/9798fbcab1e04123f05e7692475e104310e0473a/datafusion/expr/src/function.rs#L552).

The return type: This is a single Result<DataType> based on the actual types of the input. StrPos's is [here](https://github.com/apache/arrow-datafusion/blob/9798fbcab1e04123f05e7692475e104310e0473a/datafusion/expr/src/function.rs#L207).

Many of the string methods are valid for both DataType::Utf8 and DataType:LargeUtf8. Arguments and return types related to these strings will flip between i32 and i64 depending on whether it is large or not. Currently we only support DataType::Utf8, so i32 is used everywhere.

@alex-astronomer


akennedy4155:
LGTM!  I don't see any faults here looking through quickly, however I can imagine that there might be other places than strpos etc that need changes, but I wouldn't know where to start looking for that.  Just want to make sure that every spot that needs the type suffix is taken into account.

akennedy4155:
Does this need a Closes directive for the Issues related?

mwylde (approved):

PR #68, opened 2023-04-21
New Feature - Inspect Source

alex-astronomer:
# Feature Description

User can "Inspect" a source and get a sample of the raw output coming from that source.

image

When the "Eye" (Inspect) icon is clicked, a page opens with a button to refresh the sample, and a table that views the sample after refreshing.

image

image

image

# Implementation

When a source is created, a passthrough `SELECT * FROM <source name>` pipeline is created.

Once this is compiled, when the user clicks the Inspect icon and refreshes on the source detail view, the pipeline starts for a short period of time.  When the pipeline runs, a small subset (20 hardcoded for now) of records is captured and sent to the UI.  Then the pipeline is stopped.

# Feature Todo

- [ ] Make a loading icon for the icon button when compile in progress
- [ ] How does this behave when there are multiple users on the same pipeline both hitting refresh?
- [ ] Eye icon replaced with magnifying glass instead to reflect "inspect" terminology
- [ ] Cache the sample in the DB
- [ ] 404 if navigating to a detail page that doesn't exist
- [ ] More source details, similar to what pipeline view has


arroyo-api/migrations/V2__add_raw_pipeline_to_source.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE sources
+ADD COLUMN raw_pipeline_job_id VARCHAR(8);
+REFERENCES pipelines(id);
```
alex-astronomer, line 3: This migration will run automatically?

arroyo-api/src/main.rs
```diff
@@ -603,17 +577,39 @@ impl ApiGrpc for ApiServer {
...
+        transaction
+            .execute("SET TRANSACTION ISOLATION LEVEL SERIALIZABLE", &[])
             .await
+            .map_err(log_and_map)?;
+
+        let resp = start_or_preview(request.into_inner(), false, auth, &transaction).await;
+
+        transaction.commit().await.map_err(log_and_map)?;
+
+        resp
```
alex-astronomer, line 592: TODO: This pattern is common, could this be turned into a decorator / wrapper for other functions?

arroyo-api/src/main.rs
```diff
@@ -603,17 +577,39 @@ impl ApiGrpc for ApiServer {
...
+        transaction
+            .execute("SET TRANSACTION ISOLATION LEVEL SERIALIZABLE", &[])
             .await
+            .map_err(log_and_map)?;
+
+        let resp = start_or_preview(request.into_inner(), true, auth, &transaction).await;
+
+        transaction.commit().await.map_err(log_and_map)?;
+
+        resp
```
alex-astronomer, line 612: TODO: same logic as above, with the exception of the boolean "preview" parameter.  Refactor into one function.

arroyo-api/src/main.rs
```diff
@@ -750,6 +746,101 @@ impl ApiGrpc for ApiServer {
...
+
+        info!("connected to controller");
+
+        let mut stream = controller
+            .subscribe_to_output(Request::new(grpc::GrpcOutputSubscription {
+                job_id: job_id.clone(),
+            }))
+            .await
+            .map_err(log_and_map)?
+            .into_inner();
```
alex-astronomer, line 800: TODO: this subscription logic is similar to `subscribe_to_output` function.  Could be combined into a single function that is called by both?

arroyo-api/src/main.rs
```diff
@@ -750,6 +746,101 @@ impl ApiGrpc for ApiServer {
...
+            });
+
+            captured.push(v);
+
+            if captured.len() == 20 {
+                break;
+            }
+        }
+        // print and debug
+        info!("captured sample: {:?}", captured);
```
alex-astronomer, line 825: TODO: update logging and make more informative.

arroyo-api/src/main.rs
```diff
@@ -750,6 +746,101 @@ impl ApiGrpc for ApiServer {
...
+            job_id: job_id.clone(),
+            stop: Some(StopType::Immediate.into()),
+            checkpoint_interval_micros: None,
+            parallelism: None,
+        }))
+        .await?;
+
+        // return a bunch of random numbers to test
+        Ok(Response::new(RefreshSampleResp {
+            rando: captured
```
alex-astronomer, line 837: TODO: rando?  needs a new var name.... this was a placeholder

arroyo-api/src/main.rs
```diff
@@ -791,6 +882,30 @@ impl ApiGrpc for ApiServer {
...
+                key: d.key,
+                value: d.value,
+            });
+
+            captured.push(v);
+
+            if captured.len() == 20 {
+                break;
+            }
+        }
```
alex-astronomer, line 905: TODO: can remove this, also a placeholder / test.

arroyo-api/src/pipelines.rs
```diff
@@ -164,7 +166,7 @@ pub(crate) async fn create_pipeline<'a>(
     tx: &Transaction<'a>,
 ) -> Result<i64, Status> {
     let pipeline_type;
-    let mut program;
+    let mut program: Program;
```
alex-astronomer, line 169: TODO: remove annotation.

arroyo-api/src/pipelines.rs
```diff
@@ -303,6 +305,24 @@ impl TryInto<PipelineDef> for DbPipeline {
     }
 }
 
+pub(crate) async fn start_or_preview<'a>(
```
alex-astronomer, line 308: TODO: Can this function be named something like create_or_preview?  "start" implies what we call an "update" which is just adjusting the state of the job.  Here though, start means to create a new pipeline.

arroyo-api/src/sources.rs
```diff
@@ -10,39 +10,48 @@ use arrow::datatypes::TimeUnit;
...
+        TestSourceMessage,
+    },
+    StopMode,
 };
 use arroyo_sql::{
-    types::{StructDef, StructField, TypeDef},
+    types::{self, StructDef, StructField, TypeDef},
     ArroyoSchemaProvider,
 };
+use chrono::Utc;
```
alex-astronomer, line 33: TODO: not used.

arroyo-api/src/sources.rs
```diff
@@ -720,10 +729,48 @@ pub(crate) async fn create_source(
...
+            &"alex",
+            &Some(public::StopMode::immediate),
+            &None,
+            &<Option<serde_json::Value>>::None,
+            &job_resp.get_ref().job_id,
+            &"org",
+        )
+        .await
+        .map_err(|err| handle_db_error("pipeline", err))?;
+
```
alex-astronomer, line 771: TODO: refactor this into its own function

arroyo-console/src/routes/pipelines/JobOutputs.tsx
```diff
@@ -1,8 +1,8 @@
-import { Th, Td, Tr, Table, Thead, Tbody } from "@chakra-ui/react";
+import { Th, Td, Tr, Table, Thead, Tbody, TableContainer, Box, Container } from "@chakra-ui/react";
```
alex-astronomer, line 1: TODO: some of these imports are now unneeded.

arroyo-console/src/routes/sources/SourcePreview.tsx
```diff
@@ -0,0 +1,47 @@
...
+                        </Button>
+                    </Box>
+                </Flex>
+                <Flex>
+                    {!loading && outputs.length > 0 ? <PipelineOutputs outputs={outputs} /> : <Box />}
+                </Flex>
+            </VStack>
+        </Box>
+    )
+}
```
alex-astronomer, line 47: TODO: This should be renamed to SourceDetail

arroyo-console/src/routes/sources/Sources.tsx
```diff
@@ -164,6 +165,20 @@ function SourceTable({ client }: { client: ApiClient }) {
               ))}
 
               <Td textAlign="right">
+                {/* IconButton that navigates to the source preview view with a preview icon */}
```
alex-astronomer, line 168: TODO: remove comment

arroyo-console/vite.config.ts.timestamp-1682048020433.mjs
```diff
@@ -0,0 +1,10 @@
...
+var vite_config_default = defineConfig({
+  plugins: [react()]
+});
+export {
+  vite_config_default as default
+};
+//# sourceMappingURL=data:application/json;base64,ewogICJ2ZXJzaW9uIjogMywKICAic291cmNlcyI6IFsidml0ZS5jb25maWcudHMiXSwKICAic291cmNlc0NvbnRlbnQiOiBbImNvbnN0IF9fdml0ZV9pbmplY3RlZF9vcmlnaW5hbF9kaXJuYW1lID0gXCIvVXNlcnMvYWxleGtlbm5lZHkvRGV2ZWxvcGVyL2Fycm95by9hcnJveW8tY29uc29sZVwiO2NvbnN0IF9fdml0ZV9pbmplY3RlZF9vcmlnaW5hbF9maWxlbmFtZSA9IFwiL1VzZXJzL2FsZXhrZW5uZWR5L0RldmVsb3Blci9hcnJveW8vYXJyb3lvLWNvbnNvbGUvdml0ZS5jb25maWcudHNcIjtjb25zdCBfX3ZpdGVfaW5qZWN0ZWRfb3JpZ2luYWxfaW1wb3J0X21ldGFfdXJsID0gXCJmaWxlOi8vL1VzZXJzL2FsZXhrZW5uZWR5L0RldmVsb3Blci9hcnJveW8vYXJyb3lvLWNvbnNvbGUvdml0ZS5jb25maWcudHNcIjtpbXBvcnQgeyBkZWZpbmVDb25maWcgfSBmcm9tIFwidml0ZVwiO1xuaW1wb3J0IHJlYWN0IGZyb20gXCJAdml0ZWpzL3BsdWdpbi1yZWFjdFwiO1xuXG4vLyBodHRwczovL3ZpdGVqcy5kZXYvY29uZmlnL1xuZXhwb3J0IGRlZmF1bHQgZGVmaW5lQ29uZmlnKHtcbiAgcGx1Z2luczogW3JlYWN0KCldLFxufSk7XG4iXSwKICAibWFwcGluZ3MiOiAiO0FBQXdVLFNBQVMsb0JBQW9CO0FBQ3JXLE9BQU8sV0FBVztBQUdsQixJQUFPLHNCQUFRLGFBQWE7QUFBQSxFQUMxQixTQUFTLENBQUMsTUFBTSxDQUFDO0FBQ25CLENBQUM7IiwKICAibmFtZXMiOiBbXQp9Cg==
```
alex-astronomer, line 10: TODO: remove this file altogether

arroyo-rpc/proto/api.proto
```diff
@@ -745,6 +746,14 @@ message DeleteSinkResp {
...
 // outputs
 
+message RefreshSampleReq {
+  uint64 source_id = 1;
+}
+
+message RefreshSampleResp {
+  repeated OutputData rando = 1;
+}
+
```
alex-astronomer, line 756: Q: what is the difference between api.proto & rpc.proto?  Is this needed both places?

arroyo-api/src/main.rs
alex-astronomer: Needs more thought.  See #69.

arroyo-api/src/main.rs
alex-astronomer: See #69 for more.

RNHTTR:
Consider automatically loading the sample data upon clicking the eye/inspect icon instead of also having to click refresh.

mwylde (commented):
Excited to see this contribution and for the feature!

I have a couple of high-level thoughts about this.

The first is, for the feature as it exists in this PR, I wonder if the existing preview APIs would be sufficient to build this as a purely (or mostly) front-end feature. The preview API lets you send it a query, which it will run for a certain amount of time. You could spin up a preview pipeline with `select * from source` whenever the user clicks the preview button, which would avoid a lot of the need to track the pipeline / restart it / refresh the outputs, etc. And since it uses existing functionality it would reduce the complexity of this feature significantly.

The second thought is that having a source preview feature would be especially useful in the source creation flow, when a user needs to create a schema and validate it against the data.

However, currently you need a source (with a schema) in order to create a pipeline so this approach won't work.

I see a couple of paths to address this:
* Rely on a separate mechanism to tail sources that doesn't go through the whole pipeline machinery (you can see a potential start of this approach in the source testing machinery in https://github.com/ArroyoSystems/arroyo/blob/master/arroyo-api/src/testers.rs). The main downside here that it requires duplicating some of the source logic, unless it's possible to reuse it in some way.
* @jacksonrnewhouse is going to be working on adding support for raw schemas for sources, with `JSONB` functions to query them. Once that's available, we might be able to rely on that to tail the data.

At a higher level, for these initial PRs it will be easier to review if the first pass is a pretty direct implementation of the feature without a lot of refactoring.

Happy to jump on a call as well to talk through this stuff if that would be helpful.

arroyo-api/src/main.rs
```diff
@@ -603,17 +570,39 @@ impl ApiGrpc for ApiServer {
         request: Request<CreatePipelineReq>,
     ) -> Result<Response<CreateJobResp>, Status> {
         let (request, auth) = self.authenticate(request).await?;
-        self.start_or_preview(request.into_inner(), false, auth)
+
+        let mut client = self.client().await?;
```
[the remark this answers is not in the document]
alex-astronomer: > Excited to see this contribution and for the feature!
> 
> I have a couple of high-level thoughts about this.
> 
> The first is, for the feature as it exists in this PR, I wonder if the existing preview APIs would be sufficient to build this as a purely (or mostly) front-end feature. 

I don't believe that the preview function is going to accomplish what we're looking for.  Here are some reasons why:

Preview generates a pipeline with a TTL limit.  I would like something that doesn't have that bound on it.  The primary purpose of the source passthrough (raw) pipeline is to avoid compile times.  If this pipeline doesn't live indefinitely, bounded to the lifetime of the source, then you may have to compile this every once in a while.  I would rather have the overhead of a permanent pipeline with the trade off being that you don't *ever* need to worry about compiling this.

My biggest gripe so far is the compile times are really long for all queries, and this interrupts my flow and my ability to prototype and try new things out quickly.  Shorter iteration times lead to better solutions, at least from my experience and my style of data processing and writing these types of pipelines.

By having this as a first-class pipeline, compiled and ready to go, it lets me immediately see the output of a source.

> The preview API lets you send it a query, which it will run for a certain amount of time. You could spin up a preview pipeline with `select * from source` whenever the user clicks the preview button, which would avoid a lot of the need to track the pipeline / restart it / refresh the outputs, etc. And since it uses existing functionality it would reduce the complexity of this feature significantly.

Again, the compile times here are long to the point that it would (in my opinion) defeat the purpose of having this feature in the first place.  I want to be able to click the preview button and get immediate results, rather than having to wait for *anything* to compile, at *any* point other than when generating the pipeline and compiling when the source is created.

> The second thought is that having a source preview feature would be especially useful in the source creation flow, when a user needs to create a schema and validate it against the data.
> 
> However, currently you need a source (with a schema) in order to create a pipeline so this approach won't work.
> 
> I see a couple of paths to address this:
> 
> * Rely on a separate mechanism to tail sources that doesn't go through the whole pipeline machinery (you can see a potential start of this approach in the source testing machinery in https://github.com/ArroyoSystems/arroyo/blob/master/arroyo-api/src/testers.rs). The main downside here that it requires duplicating some of the source logic, unless it's possible to reuse it in some way.

I thought about this actually!  I thought the fastest way forward would be to implement something utilizing existing structs and code like we have for first-class pipelines, but I agree: logically it makes the most sense to me to just "tail" a "source", meaning we just visualize *actual raw stream output* instead of generating an entire pipeline for the same.

> * @jacksonrnewhouse is going to be working on adding support for raw schemas for sources, with `JSONB` functions to query them. Once that's available, we might be able to rely on that to tail the data.

Looking forward to seeing this.

> At a higher level, for these initial PRs it will be easier to review if the first pass is a pretty direct implementation of the feature without a lot of refactoring.

Totally.  I got carried away and should have separated it out.  Refactor should exist in a separate PR.  I'll try to go through and pare this down to just the essential changes.

> Happy to jump on a call as well to talk through this stuff if that would be helpful.

Let's find some time to chat about it!  Would love to introduce myself.

---

For now, I'm going to continue down this path, until we get some other logic for tailing streams without having an official "source". I do strongly believe that this is the fastest, most efficient way to get this feature into the product and due to the nature of the change, it could be refactored and these raw pipelines are decoupled enough from the rest of the system that they could be removed and revamped in favor of a "tail" output for a "stream" (not an official "source").

Let's

alex-astronomer:
> ame.
> 
> > * @jacksonrnewhouse is going to be working on adding support for raw schemas for sources, with `JSONB` functions to query them. Once that's available, we might be able to rely on that to tail the data.
> 
> Looking forward to seeing this.
> 
> > At a higher level, for these initial PRs it will be easier to review if the first pass is a pretty direct implementation of the feature without a lot of refactoring.
> 
> Totally. I got carried away and should have separated it out. Refactor should exist in a separate PR. I'll try to go through and pare this down to just the essential changes.
> 
> > Happy to jump on a call as well to talk through this stuff if that would be helpful.
> 
> Let's find some time to chat about it! Would love to introduce myself.
> 
> For now, I'm going to continue down this path, until we get some other logic for tailing streams without having an official "source". I do strongly believe that this is the fastest, most efficient way to get this feature into the product and due to the nature of the change, it could be refactored and these raw pipelines are decoupled enough from the rest of the system that they could be removed and revamped in favor of a "tail" output for a "stream" (not an official "source").

arroyo-api/migrations/V2__add_raw_pipeline_to_source.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE sources
```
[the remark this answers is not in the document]
alex-astronomer: See comment below for why I don't want to use preview feature.

arroyo-api/src/connections.rs
```diff
@@ -12,38 +13,41 @@ use crate::queries::api_queries;
 use crate::queries::api_queries::DbConnection;
 use crate::types::public;
 use crate::{handle_db_error, log_and_map, required_field, testers::KafkaTester, AuthData};
+use futures::future::BoxFuture;
 
-pub(crate) async fn create_connection(
+pub(crate) fn create_connection<'a>(
```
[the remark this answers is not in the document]
alex-astronomer: Going to remove all of the refactors to get to the core of this change.

Issue #74, opened 2023-04-24
The UI should allow users to change parallelism


mwylde:
Pipelines have an associated parallelism configuration that controls how many parallel subtasks we run for each operator. (Inside the dataflow itself, we support operators have different parallelism, but in the current API we only allow setting a single parallelism across the entire job).

This parallelism is set to an inferred value at pipeline creation, however that may be too high or low depending on the actual data volume and complexity of the query.

There is a gRPC API (`UpdateJob`) that allows users to change the parallelism of a running job, but it is not currently exposed on the Web UI.

This issue covers adding the ability to change the parallelism from the job details page (http://localhost:8000/jobs/{job_id}).

Note that because we do not currently support dynamic rescaling of pipelines, changing the parallelism triggers this sequence in the controller:
1. The job is stopped with a final checkpoint
2. The existing workers are shut down
3. A new job is scheduled with the new parallelism

This can take several seconds, and the UI should reflect that the change is rolling out.

PR #77, opened 2023-04-25
Add pipeline edit modal with parallelism control

jbeisen:
Add an "Edit" button on the job details page that opens up a modal. The modal contains a form that can be used to change the parallelism.

Resolves: #74

---

This is a work in progress. Still to do:

- [x] add more validation to the form
- [ ] ~~indicate on the job details page that the pipeline is restarting~~
  I guess it already does this?
- [ ] maybe change the design of the restart warning?

---

The modal:
image

If you change the parallelism value:

image

### DIFF
--- a/arroyo-console/src/routes/pipelines/PipelineConfigModal.tsx
+++ b/arroyo-console/src/routes/pipelines/PipelineConfigModal.tsx
@@ -1,0 +1,133 @@
+import React, { useState } from "react";
+import {
+  Alert,
+  AlertDescription,
+  AlertIcon,
+  Button,
+  FormControl,
+  FormErrorMessage,
+  FormHelperText,
+  FormLabel,
+  Modal,
+  ModalBody,
+  ModalCloseButton,
+  ModalContent,
+  ModalFooter,
+  ModalHeader,
+  ModalOverlay,
+  NumberDecrementStepper,
+  NumberIncrementStepper,
+  NumberInput,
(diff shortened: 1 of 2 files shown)


arroyo-console/src/routes/pipelines/PipelineConfigModal.tsx
```diff
@@ -0,0 +1,101 @@
...
+    changeAlert = (
+      <Alert status="warning">
+        <AlertIcon />
+        <AlertDescription>These changes require a pipeline restart.</AlertDescription>
+      </Alert>
+    );
+  }
+
+  const parallelismInput = (
+    <Input
```
mwylde, line 57: Should this be a [number input](https://chakra-ui.com/docs/components/number-input)?

mwylde (commented):
This is great!

The only issue I noticed is that if you edit the parallelism and hit cancel, re-opening the modal shows the edited parallelism rather than the actual current parallelism (in other words, cancel doesn't reset the state of the modal).

arroyo-console/src/routes/pipelines/PipelineConfigModal.tsx
jbeisen: Yes, done. I had to manually set the background color, though. Maybe there's a bug with the dark theme.

mwylde (approved):
Ship it!

PR #80, opened 2023-04-26
sql: add a SQL IR and factor out optimizations.

jacksonrnewhouse:
This introduces arroyo-sql/src/plan_graph.rs, which is a new intermediate representation for our SQL planning. Prior to this change we were converting from DataFusion's logical plan to a SqlOperator and then the SqlOperator produced an `arroyo_datastream::Program`. However, because the SqlOperator is not in a graph it was challenging to perform optimization passes. In order to capture some optimizations they would happen in-line with the conversions from LogicalPlan to SqlOperator. As we added more logic like this it was getting unwieldy. The new sequence for compiling SQL to a pipeline is.

- Parse the request using DataFusion's `PostgresSqlDialect`.
- Have data fusion translate this to a logical plan.
- Run DataFusion optimization over the logical plan.
- Convert the logical plan to a SqlOperator.
- Convert the `SqlOperator` to a `PlanGraph`.
- Run arroyo-sql optimizations over PlanGraph.
- Convert to `arroyo_datastream::Program`. 
- Run arroyo-api optimizations over that program.
- Produce the crate for the pipeline.
- Compile the crate.

It's a lot of steps, but firming up what happens where should allow us to continue to improve our engine.


mwylde:
Getting a compilation error for this query:

```sql
WITH bids as (SELECT bid.auction as auction, bid.datetime as datetime
        FROM (select bid from nexmark) where bid is not null)
        SELECT AuctionBids.auction as auction, AuctionBids.num as count
        FROM (
          SELECT
            B1.auction,
            HOP(INTERVAL '2' SECOND, INTERVAL '10' SECOND) as window,
            count(*) AS num
          FROM bids B1
          GROUP BY 1,2) AS AuctionBids
        JOIN (
          SELECT max(num) AS maxn, window
          FROM (
            SELECT count(*) AS num, HOP(INTERVAL '2' SECOND, INTERVAL '10' SECOND) AS window
            FROM bids B2
            GROUP BY B2.auction, 2
        ) AS CountBids
          GROUP BY 2
        ) AS MaxBids
        ON
           AuctionBids.num = MaxBids.maxn
           and AuctionBids.window = MaxBids.window;
```

mwylde:
This query compiles (interestingly enough, it fails on master) but doesn't produce any output: 

```sql
SELECT * FROM (
    SELECT *, ROW_NUMBER() OVER (
        PARTITION BY window 
        ORDER BY bids DESC) as row_num   
    FROM (SELECT count(*) as bids,
        bid.auction as auction_id,
        hop(interval '2 seconds', interval '1 minute') as window
            FROM nexmark_200k
            WHERE bid is not null
            group by auction_id, window)) WHERE row_num <= 3```

jacksonrnewhouse:
Looking into those.

jacksonrnewhouse:
This is working now.

mwylde (approved):

Issue #73, opened 2023-04-24
Catalog in web ui should show SQL types to the user


mwylde:
Users can view their available sources by looking at the catalog, on the left sidebar of the query editor.

![image](https://user-images.githubusercontent.com/100920/233875351-c6ae6381-704e-4efe-b3d5-187a7d233b76.png)

Currently the catalog returns types from our internal naming schema, like `Int64` and `UnixMillis`. These are returned by the `GetSources` gRPC call, which returns a list of `SourceDef` messages, each of which contains a list of `SourceField`. Ultimately, you end up returning various `PrimitiveType`s with names like `Int64`, etc.

This API should instead return SQL types that will be familiar to SQL users, like `bigint` and `timestamp`.

PR #84, opened 2023-04-27
Show SQL names instead of primitive types in catalog

jbeisen:
Define a mapping of primitive types to SQL types. Add a field 'sql_name' to the SourceFieldType message, and populate it for primitive types. Display this value in the catalog.

Resolves: #73

---

image


arroyo-api/src/sources.rs
```diff
@@ -196,6 +196,21 @@ pub enum PrimitiveType {
     UnixMillis,
 }
 
+fn primitive_to_sql(primitive_type: &PrimitiveType) -> &'static str {
+    match primitive_type {
+        PrimitiveType::Int32 => "INTEGER",
+        PrimitiveType::Int64 => "BIGINT",
+        PrimitiveType::UInt32 => "UNSIGNED INTEGER", // not really sql
+        PrimitiveType::UInt64 => "UNSIGNED BIGINT",  // not really sql
```
jbeisen, line 204: These types don't have a sql equivalent, right?

arroyo-api/src/sources.rs
```diff
@@ -353,9 +368,17 @@ impl TryFrom<SchemaField> for SourceField {
             }
         };
 
+        let sql_name = match value.typ {
+            SchemaFieldType::Primitive(p) => primitive_to_sql(&p),
+            _ => ""
```
jbeisen, line 373: I wasn't so sure about this case, thoughts?

arroyo-api/src/sources.rs
jacksonrnewhouse: I'd look at https://arrow.apache.org/datafusion/user-guide/sql/data_types.html. In which case `INTEGER UNSIGNED` and `BIGINT UNSIGNED`.

arroyo-api/src/sources.rs
```diff
@@ -196,6 +196,21 @@ pub enum PrimitiveType {
...
+fn primitive_to_sql(primitive_type: &PrimitiveType) -> &'static str {
+    match primitive_type {
+        PrimitiveType::Int32 => "INTEGER",
+        PrimitiveType::Int64 => "BIGINT",
+        PrimitiveType::UInt32 => "UNSIGNED INTEGER", // not really sql
+        PrimitiveType::UInt64 => "UNSIGNED BIGINT",  // not really sql
+        PrimitiveType::F32 => "FLOAT",
+        PrimitiveType::F64 => "DOUBLE",
+        PrimitiveType::Bool => "BOOLEAN",
+        PrimitiveType::String => "VARCHAR",
```
jacksonrnewhouse, line 208: I kinda prefer `TEXT` to `VARCHAR`, if we're gonna pick one.

arroyo-api/src/sources.rs
```diff
@@ -196,6 +196,21 @@ pub enum PrimitiveType {
...
+        PrimitiveType::Bytes => "BLOB",
```
jacksonrnewhouse, line 209: The datafusion page calls this `BYTEA`, but we can kinda pick. Could also go with BINARY

arroyo-api/src/sources.rs
jacksonrnewhouse: We already don't show the current type, so I think this is fine.

arroyo-rpc/proto/api.proto
```diff
@@ -581,6 +581,9 @@ message SourceFieldType {
     PrimitiveType primitive = 1;
     StructType struct = 2;
   }
+
+  // include a sql name if the type is PrimitiveType
```
jacksonrnewhouse, line 585: Rather than showing `""`, maybe make it optional?

jacksonrnewhouse (approved):

arroyo-api/src/sources.rs
mwylde: Agreed (should have said that in the original issue)

arroyo-api/src/sources.rs
mwylde: I'd go with BINARY or BYTEA over BLOB. BLOB is "binary large object" and is used for storing files in the DB (like images). BYTEA/BINARY and used for smaller binary strings.

mwylde (approved):

arroyo-api/src/sources.rs
mwylde, line 373: I would list out the other cases here instead of using `_` so that it's more clear how we're handling each of them, and so that if add another SchemaFieldType we'll have to come here and hand that case.

arroyo-api/src/sources.rs
```diff
@@ -329,7 +344,7 @@ impl TryFrom<SchemaField> for SourceField {
     type Error = String;
 
     fn try_from(value: SchemaField) -> Result<Self, Self::Error> {
-        let typ = match value.typ {
+        let typ = match value.clone().typ {
```
mwylde, line 347: The clone is not necessary, just need to take a reference to value.typ instead of taking it by value

```suggestion
        let typ = match &value.typ {
```

arroyo-rpc/proto/api.proto
jbeisen: Sure, done.

arroyo-api/src/sources.rs
jbeisen: Done, made it `BINARY`

arroyo-api/src/sources.rs
mwylde: You would still need to clone that (as you try_into() is implemented on the owned type) but it's preferable to cloning the entire struct.

mwylde (approved):

PR #86, opened 2023-04-27
sql: Introduce initial JSON functions and raw Kafka Source

jacksonrnewhouse:
This adds three json-related UDFs that leverage JSONPath, an emerging standard for querying data out of JSON. The syntax is sort of similar to what is used by `jq`, and mirrors the [get_json_object](https://spark.apache.org/docs/3.1.2/api/python/reference/api/pyspark.sql.functions.get_json_object.html) in pyspark sql. I've started a discussion at #85.

- `get_json_objects`: Returns a list of Strings corresponding to the results of the json path.
- `git_first_json_object`: Returns the first match, if there are any. I'd expect this to mainly get used when you expect there's only one match.
- `extract_json_string`: Because JSON strings are internally quoted, this gives the actual string, e.g. `"a"` versus `"\"a\""`.


arroyo-api/src/sources.rs
```diff
@@ -29,7 +30,7 @@ use deadpool_postgres::Pool;
 use http::StatusCode;
 use tokio::sync::mpsc::{channel, Receiver, Sender};
 use tonic::Status;
-use tracing::warn;
+use tracing::{info, warn};
```
mwylde, line 33: the info import isn't used

arroyo-api/src/sources.rs
```diff
@@ -373,13 +385,14 @@ pub struct SourceSchema {
 
     // set to true to use the kafka schema serialization wire format
     // (https://docs.confluent.io/platform/current/schema-registry/serdes-develop/index.html#wire-format)
-    kafka_schema: bool,
+    kafka_serialization_mode: Option<KafkaSerializationMode>,
```
mwylde, line 388: update the comment?

arroyo-api/src/sources.rs
```diff
@@ -373,13 +385,14 @@ pub struct SourceSchema {
...
 }
 
 fn builtin_for_name(name: &str) -> Result<SourceSchema, String> {
     match name {
         "nexmark" => Ok(nexmark_schema()),
         "impulse" => Ok(impulse_schema()),
+        "raw_kafka" => Ok(raw_kafka_schema()),
```
mwylde, line 395: Should this be something more generic than "raw_kafka"? It seems like any other stream source could similarly have a raw format.

Separately, should this be a builtin or a separate SourceSchema? The existing builtins are used solely for demo sources that have particular schemas.

arroyo-rpc/proto/api.proto
```diff
@@ -636,6 +642,7 @@ message CreateSourceReq {
     ImpulseSourceConfig impulse = 4;
     FileSourceConfig file = 5;
     NexmarkSourceConfig nexmark = 6;
+    KafkaSourceConfig raw_kafka = 7;
```
mwylde, line 645: Why is it necessary to have raw_kafka be separate from kafka here? Would you expect to also have a proto_kafka and avro_kafka, and proto_pulsar, etc.?

arroyo-rpc/proto/api.proto
```diff
@@ -167,11 +167,17 @@ message KafkaSource {
   string topic = 1;
   repeated string bootstrap_servers = 2;
   OffsetMode offset_mode = 3;
-  bool schema_registry = 4;
+  KafkaSerializationMode kafka_serialization_mode = 4;
```
mwylde, line 170: Note that this is a breaking change for existing pipelines

mwylde (commented):
Overall this looks good. I think the area where it still needs a bit of work is how schemas interact with sources.

The goal of the schema/source distinction is that these are meant to be independent: a source (like kafka or pulsar) may have data in a variety of types of schemas (json schema, raw json, protobuf, etc.). But the way that raw json is implemented in this PR means that instead of implementing e.g., raw json once in the API, we will need to implement it per-source.

I would focus on how we can maintain the existing abstraction.

arroyo-api/src/sources.rs
```diff
@@ -380,6 +393,7 @@ fn builtin_for_name(name: &str) -> Result<SourceSchema, String> {
     match name {
         "nexmark" => Ok(nexmark_schema()),
         "impulse" => Ok(impulse_schema()),
+        "raw" => Ok(raw_schema()),
```
mwylde, line 396: is this still used?

mwylde (approved):

arroyo-api/src/sources.rs
jacksonrnewhouse: let me check, I don't think so.

PR #90, opened 2023-05-01
Raw Source Preview

alex-astronomer:
Add a Source Detail view

image

Clicking the link takes you to the page, similar to jobs.

image

Tail button starts a preview with a `select * from <source>` query.  Once compiled, the output starts to stream to the main pane.

image


alex-astronomer:
Future plans:

* use the `DetailFieldTable` component for `JobDetail.tsx` pipeline details for coherence
* use `PipelinePreview` component in `CreatePipeline.tsx` to consolidate
* make a more robust distinction between preview and non-preview queries besides just a "preview" prefix

arroyo-console/src/routes/pipelines/JobOutputs.tsx
```diff
@@ -36,15 +36,17 @@ export function PipelineOutputs({ outputs }: { outputs: Array<{id: number, data:
     .reverse();
 
   return (
-    <Table maxWidth="500px">
+    <TableContainer overflow="scroll">
```
mwylde, line 39: We don't want scroll, as that creates a scroll bar even when not needed. I'd look at how the existing job preview page solves this

arroyo-console/src/routes/sources/SourceDetail.tsx
```diff
@@ -0,0 +1,141 @@
...
+                {/* flex with pipeline outputs and a column of text */}
+                <Box>
+                    <Flex w="100%">
+                        <Box w="calc(100vw - 600px)">
+                            <PipelinePreview client={client} jobId={previewJobId} subscribed setSubscribed={setSubscribed} setJobState={setJobState} />
+                        </Box>
+                        <Spacer />
+                        <DetailFieldTable data={[
+                            { field: "Name", value: source?.name ?? "" },
+                            { field: "# of Consumers", value: source?.consumers?.toString() ?? "" },
```
mwylde, line 132: I think this can just be "Consumers"

arroyo-console/src/routes/sources/SourceDetail.tsx
```diff
@@ -0,0 +1,141 @@
...
+                    query: `SELECT * FROM ${source!.name}; `,
+                    sink: { case: "builtin", value: BuiltinSink.Web }
+                }),
+            },
+        }));
+        setPreviewJobId(jobResp.jobId);
+    }
+
+    return (
+        <Box>
```
mwylde, line 114: The spacing of this is a bit inconsistent with the rest of the UI, and having the badge and the button next to each other makes it somewhat unclear which one is the button.

 I'd recommend something like this for the header:

```tsx
      <VStack>
        <Flex width="100%" p={5}>
          <Heading size="sm">Source {source?.name} </Heading>
          <Center>{jobState != null ? <Badge ml={3}>{jobState}</Badge> : null}</Center>
          <Spacer />

          <Button size="sm" onClick={tail} isLoading={subscribed} loadingText="In Progress">
            Tail
          </Button>
        </Flex>
```

mwylde (commented):
This is great! A few design things that can be improved but overall looks really good.

A couple of other bits of feedback:
* I think there needs to be a way to pause or stop the pipeline. Otherwise, for high volume sources the rows move past you too fast to really see.
* Once the pipeline is stopped (after 30 seconds) the "In Progress" spinner keeps going and there's no way to restart the tail. I'd take a look at how this is handled in the job preview.

PR #91, opened 2023-05-01
K8s deployment and helm chart

mwylde:
This PR adds a Helm chart to allow easy deployment of Arroyo into Kubernetes. The helm chart is fairly customizable at this point, although is not likely to be sufficiently so for all use cases. It should however be sufficient for quickly getting users up and running on a local K8s (like minikube or microk8s) and EKS. Once we have support for GCS it should also be easy to run on GKE.

Docs will be forthcoming, but to get a fully self-contained cluster (with prometheus and postgres) running on a local k8s:

local-config.yaml
```yaml
outputDir: "/tmp/arroyo-test"

volumes:
  - name: checkpoints
    hostPath:
      path: /tmp/arroyo-test
      type: DirectoryOrCreate

volumeMounts:
  - name: checkpoints
    mountPath: /tmp/arroyo-test
```

```sh
~/arroyo/k8s $ helm install -f local-config.yaml arroyo-test ./arroyo
```

or within EKS

aws-config.yaml
```yaml
s3:
  bucket: "arroyo-testing"
  region: "us-east-1"
```

```sh
~/arroyo/k8s $ helm install -f aws-config.yaml arroyo-test ./arroyo
```

Once this is merged, we will be able to use our repo as a helm repo to allow installing the chart directly.


k8s/arroyo/Chart.yaml
```diff
@@ -0,0 +1,22 @@
...
+  - data
+  - real-time
+
+dependencies:
+  - name: prometheus
+    version: "^21"
+    repository: https://prometheus-community.github.io/helm-charts
+    condition: prometheus.deploy
+  - name: postgresql
+    version: "^12"
```
jacksonrnewhouse, line 20: Why not be on 15?

jacksonrnewhouse (approved):

k8s/arroyo/Chart.yaml
mwylde: This is version 12 of the helm chart, which is Postgres 15 (https://artifacthub.io/packages/helm/bitnami/postgresql)

PR #101, opened 2023-05-04
Feature/inline create table

jacksonrnewhouse:
This lets you write a CREATE TABLE statement (or multiple), before your primary query this table will be added as a table before DataFusion parses your query, so can be queried. For example, I wrote
```
SELECT person.name as name, person.id as id FROM nexmark where person is not null
```
to my local kafka connection at topic `output`.

I was able to query it with a table like
```
CREATE TABLE people (
  id INT not null,
  name STRING not null,
) with (connection='local', topic='output');

SELECT * FROM people;
```


arroyo-sql/src/lib.rs
```diff
@@ -164,6 +175,83 @@ impl ArroyoSchemaProvider {
     pub fn add_defs(&mut self, source: impl Into<String>, defs: impl Into<String>) {
         self.source_defs.insert(source.into(), defs.into());
     }
+
+    fn get_inlined_source(
+        &self,
+        query: &Statement,
+    ) -> Result<(String, Vec<StructField>, Operator)> {
```
mwylde, line 182: nit: indentation is off here

arroyo-sql/src/lib.rs
```diff
@@ -164,6 +175,83 @@ impl ArroyoSchemaProvider {
...
+                        kafka_input_format: SerializationMode::Json,
+                        // TODO: use setting for account
+                        messages_per_second: 10000000,
+                        client_configs: auth_config_to_hashmap(
+                            kafka_connection.auth_config.clone(),
+                        ),
+                    },
+                ))
+            }
+            _ => bail!("unsupported connection type"),
```
mwylde, line 252: I think it's worth being explicit here so that we know where we need to update when add new connection types. Additionally, this case is actually `None`, not an unsupported connection type, so this actually means that the protobuf is malformed somehow

arroyo-sql/src/lib.rs
```diff
@@ -164,6 +175,83 @@ impl ArroyoSchemaProvider {
...
+                        topic,
+                        // split to a vec
+                        bootstrap_servers: kafka_connection
+                            .bootstrap_servers
+                            .split(",")
+                            .map(|s| s.to_string())
+                            .collect(),
+                        offset_mode: OffsetMode::Latest,
+                        kafka_input_format: SerializationMode::Json,
+                        // TODO: use setting for account
```
mwylde, line 244: this points to a bit of ugliness here, that we're having to solve this class of problems (configuring a kafka source) both here and in the API -- I don't have a good solution to it right now though

mwylde (approved):

arroyo-sql/src/lib.rs
jacksonrnewhouse: This is what `cargo fmt` wants, apparently.

arroyo-sql/src/lib.rs
jacksonrnewhouse: Yeah, agree we need to figure out what to do. We could shove more relevant data into the ArroyoSchemaProvider, apply the constraints after coming out of SQL, or something else.

arroyo-sql/src/lib.rs
jacksonrnewhouse: Could also be Kinesis, cleaned up the errors.

arroyo-sql/src/lib.rs
mwylde: for some reason it won't fix this, but if you manually fix it it won't revert it either

PR #104, opened 2023-05-05
MVP of Rust UDFs

mwylde:
This PR adds SQL support for UDFs written in Rust.

For example, it's now possible to register a UDF, for example:

```rust
fn my_square(x: i64) -> i64 {
  x * x
}
```

and use it from SQL like this:

```sql
SELECT my_square(auction.bid) FROM nexmark
```

UDF names, types, and nullability behavior are all automatically determined by the function definition. For the above example, the argument is `i64`, which will compile into a SQL UDF that on nullable inputs (like `auction.bid`) will be called only if the value is not null. Alternatively, we could define `fn my_square(x: Option<i64>)`, which will instead be called on all inputs.

This PR also includes a basic UI for defining UDFs:
image

This is the start our UDF support. Next steps:
* Better error messages when UDF compilation fails
* Ability to include cargo dependencies in UDF
* Python and JS UDFs
* UDAFs
* Stateful UDFs

### DIFF
--- a/arroyo-sql/src/expressions.rs
+++ b/arroyo-sql/src/expressions.rs
@@ -12,7 +12,7 @@ use arrow_schema::Field;
 use datafusion_common::ScalarValue;
 use datafusion_expr::{
     aggregate_function,
-    expr::Sort,
+    expr::{AggregateFunction, Sort},
     type_coercion::aggregates::{avg_return_type, sum_return_type},
     BinaryExpr, BuiltinScalarFunction, Expr, TryCast,
 };
@@ -482,6 +482,31 @@ impl<'a> ExpressionContext<'a> {
             expression => {
                 bail!("expression {:?} not yet implemented", expression)
             }
+        }
+    }
+
+    pub fn as_two_phase_aggregation(&mut self, expr: &Expr) -> Result<TwoPhaseAggregation> {
+        if !self.input_struct.fields.is_empty() {
+            bail!("expected single field input");
+        }
+        match expr {
+            Expr::AggregateFunction(AggregateFunction {
+                fun,
+                args,
+                distinct: false,
+                filter: None,
+            }) => {
+                if args.len() != 1 {
+                    bail!("unexpected arg length");
+                }
+                let incoming_expression = self.compile_expr(&args[0])?;
+                let aggregator = Aggregator::from_datafusion(fun.clone(), false)?;
+                Ok(TwoPhaseAggregation {
+                    incoming_expression,
+                    aggregator,
+                })
+            }
+            _ => bail!("expected aggregate expression"),
         }
     }
 }
@@ -2234,7 +2259,7 @@ impl RustUdfExpression {
                 || self
                     .args
                     .iter()
-                    .any(|(t, e)| t.is_optional() ^ e.nullable()),
+                    .any(|(t, e)| t.is_optional() && !e.nullable()),
         )
     }
 }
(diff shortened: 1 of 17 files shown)


arroyo-sql/src/operators.rs
```diff
@@ -558,8 +559,11 @@ pub struct TwoPhaseAggregation {
 }
 
 impl TwoPhaseAggregation {
-    pub fn from_expression(expr: &Expr, input_struct: &StructDef) -> Result<TwoPhaseAggregation> {
-        if !input_struct.fields.is_empty() {
+    pub fn from_expression(
+        ctx: &mut ExpressionContext,
```
jacksonrnewhouse, line 563: Should we lift these methods up into ExpressionContext?

arroyo-sql/src/expressions.rs
```diff
@@ -166,287 +170,318 @@ impl Expression {
...
+                    BuiltinScalarFunction::ToHex => bail!("hex not implemented"),
+                    BuiltinScalarFunction::Uuid => bail!("UUID unimplemented"),
+                    BuiltinScalarFunction::Cbrt => bail!("cube root unimplemented"),
+                    BuiltinScalarFunction::Degrees => bail!("degrees not implemented yet"),
+                    BuiltinScalarFunction::Pi => bail!("pi not implemented yet"),
+                    BuiltinScalarFunction::Radians => bail!("radians not implemented yet"),
+                }
+            }
+            Expr::ScalarUDF { fun, args } => match fun.name.as_str() {
+                "get_first_json_object" => {
```
jacksonrnewhouse, line 427: Right now I can register a function that overwrites builtins, any easy way to return a better error? More specifically, right now my UDF gets added to the ArroyoSchemaProvider, planned against, and then returns an error because it hits this block.

arroyo-sql/src/expressions.rs
```diff
@@ -2151,3 +2185,56 @@ impl JsonExpression {
...
+            (|| {
+                #(#defs; )*
+                #ret
+            })()
+        })
+    }
+
+    fn return_type(&self) -> TypeDef {
+        self.ret_type.with_nullity(
+            self.ret_type.is_optional()
```
jacksonrnewhouse, line 2233: This seems off, I think the return type should be null if either
* The UDF returns null
* There are non-null UDF arguments with nullable arguments.

If I have 
```
fn my_square(x: Option<i64>) -> i64 {
  x.map(|x| x * x).unwrap_or(0)
}
```
And I run 
```
select my_square(bid.price) FROM nexmark;
```
Then the result is correctly not nullable. However, if you run
```
select my_square(5) FROM nexmark;
```
The result is wrapped in Some().

arroyo-sql/src/types.rs
```diff
@@ -471,3 +530,13 @@ pub(crate) fn make_decimal_type(precision: Option<u64>, scale: Option<u64>) -> R
         Ok(DataType::Decimal128(precision, scale))
     }
 }
+
+pub fn rust_to_datafusion(typ: &Type) -> Option<DataType> {
```
jacksonrnewhouse, line 534: This can be deleted.

jacksonrnewhouse (commented):
Looks good overall, just a few comments.

arroyo-sql/src/types.rs
mwylde: left over from a rebase

jacksonrnewhouse (approved):

PR #105, opened 2023-05-08
Fix query compilation error when sorting floats

mwylde:
Attempting to sort by a float field currently causes query compilation errors, due to the unmet requirement that the float implement Ord. This PR fixes that by ensuring that floats are wrapped in [OrderedFloat](https://docs.rs/ordered-float/latest/ordered_float/struct.OrderedFloat.html) before it's used in a context where Ord is required.

### DIFF
--- a/arroyo-sql/src/expressions.rs
+++ b/arroyo-sql/src/expressions.rs
@@ -1302,7 +1302,7 @@ pub struct SortExpression {
 
 impl SortExpression {
     pub fn from_expression(ctx: &mut ExpressionContext, sort: &Sort) -> Result<Self> {
-        let mut value = ctx.compile_expr(&sort.expr)?;
+        let value = ctx.compile_expr(&sort.expr)?;
 
         let direction = if sort.asc {
             SortDirection::Asc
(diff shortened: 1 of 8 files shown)


arroyo-sql/src/expressions.rs
```diff
@@ -1330,7 +1340,16 @@ impl SortExpression {
     }
 
     pub fn to_syn_expr(&self) -> syn::Expr {
-        let value_expr = self.value.to_syn_expression();
+        let value = if self.value.return_type().is_float() {
+            Expression::WrapType(WrapTypeExpression::new(
```
jacksonrnewhouse, line 1344: We don't need to dip into Expression, could just inline the syn stuff here, but not a big deal.

jacksonrnewhouse (approved):

arroyo-sql/src/expressions.rs
mwylde: Yeah, I'm imagining we'll have other cases like this and it's worth having the slightly more general solution

PR #107, opened 2023-05-09
sql: Rework sources and sinks to allow for creating tables/views in SQL queries.

jacksonrnewhouse:
This enables some DDL behavior in our SQL API. In particular, it allows for dynamic creation of sources and sinks, creation of in-memory tables and views, and inserting into the created objects.

### Create Memory Table
You can create a table using standard SQL syntax, e.g.

```
CREATE TABLE memory_table (
first_field int,
second_field text not null)
```
This alone is not terribly useful, as the table will not persist outside of the query. You can insert into it and read from it, but those capabilities don't give you much that subqueries (`SELECT * FROM (SELECT first_field, second_field FROM source where third_field is null))` or a WITH clause `with subquery as (SELECT first_field, second_field FROM source where third_field is null) SELECT * FROM subquery`.

#### Create Memory Table From Query
This lets you define a table to be referred to later. It looks like
```
CREATE TABLE memory_table AS (
SELECT first_field, second_field FROM source where third_field is null);
```
#### Create Table with Connection
To create a source or sink within SQL using an existing connector you can write a CREATE TABLE statement with WITH options (different than the WITH clause above). The WITH options must include a `connection` key with the name of the connection. By default the serialization mode will be `Json`, but setting `serialization_mode` to `raw_json` or `json_schema_registry` will enable those modes. Currently only Kafka connections are supported and they require a `topic` key as well.

```
CREATE TABLE connection_table (
first_field int,
second_field text not null)
WITH (
connection = 'my_connection_name',
topic = 'my_topic',
serialization_mode = 'json_schema_registry`)
```

### Insert Into
For the above created tables you can INSERT INTO them with a query, e.g.
```
INSERT INTO connection_table 
SELECT first_field, second_field FROM source where third_field is null)
```
If it is a connection table, this will create it as an output sink. Otherwise it will just be treated as a memory table that can be read from.
### Create VIEW
This behaves the same as creating a memory table.


arroyo-sql/src/lib.rs
```diff
@@ -439,55 +359,249 @@ pub async fn parse_and_get_program(
...
-        bail!("Query is empty");
+struct SqlProgramBuilder<'a> {
+    schema_provider: &'a mut ArroyoSchemaProvider,
+}
+
+impl<'a> SqlProgramBuilder<'a> {
+    fn plan_query(&mut self, query: &str) -> Result<Vec<Table>> {
+        let dialect = PostgreSqlDialect {};
+        let mut outputs = Vec::new();
+        for statement in Parser::parse_sql(&dialect, query).unwrap() {
```
mwylde, line 418: This unwrap seems to remove error handling for parse errors -- is that what we want?

For example, if I accidentally use the wrong quote terminator, previously I would have gotten this error in the frontend:

```
Unterminated string literal at Line: 11, Column 22
```

but now I just get a generic error.

arroyo-sql/src/pipeline.rs
```diff
@@ -888,6 +1022,104 @@ impl<'a> SqlPipelineBuilder<'a> {
...
+            Table::TableFromQuery {
+                name: _,
+                logical_plan: _,
+            } => todo!(),
+            Table::InsertQuery {
+                sink_name,
+                logical_plan,
+            } => {
+                let input = self.insert_sql_plan(&logical_plan)?;
+                println!("tables: {:?}", self.schema_provider.tables);
```
mwylde, line 1057: remove?

arroyo-sql/src/pipeline.rs
```diff
@@ -684,19 +749,88 @@ impl<'a> SqlPipelineBuilder<'a> {
...
+            }
+            crate::Table::TableFromQuery {
+                name: _,
+                logical_plan,
+            } => self.insert_sql_plan(&logical_plan.clone())?,
+            crate::Table::InsertQuery {
+                sink_name: _,
+                logical_plan: _,
+            } => {
+                /* let input = self.insert_sql_plan(logical_plan)?;
```
mwylde, line 810: remove?

arroyo-api/src/pipelines.rs
```diff
@@ -251,10 +247,14 @@ pub(crate) async fn create_pipeline<'a>(
...
     }
 
-    // TODO: this is very hacky
-    let is_preview = req.name.starts_with("preview");
     if is_preview {
         set_parallelism(&mut program, 1);
+        for node in program.graph.node_weights_mut() {
+            // if it is a kafka sink or file sink, switch to null
+            if let Operator::KafkaSink { .. } | Operator::FileSink { .. } = node.operator {
+                node.operator = Operator::NullSink;
```
mwylde, line 255: Should this be a web sink? When I try to preview a pipeline with an insert into kafka:

```sql
CREATE TABLE sink (
  count int
)
WITH (
connection = 'local',
topic = 'test',
serialization_mode = 'json');


INSERT INTO 'sink'
select count(*) as count
from copy
group by hop(interval '2 seconds', interval '10 seconds');
```
I get this error:

> Job cotwkgz9 does not have a web sink

mwylde (commented):
Some general notes:

* The web pipeline creation flow is a bit confusing now, since you still get prompted to add a sink even if there's an insert sink.
* What's the intended behavior when defining a sink in SQL with fields that do not match the query? I.e., I define a sink with field `x` and have a query with a field `y` -- currently it seems like we just line them up by index. What if there are too few or too many?

arroyo-api/src/pipelines.rs
jacksonrnewhouse: switched to a web sink.

mwylde (approved):

PR #108, opened 2023-05-09
Add basic end-to-end integration test

mwylde:
This PR adds a simple integration test that:

* Starts up the services
* Creates a nexmark source
* Starts a pipeline with this SQL:
```sql
select count(*) from nexmark where auction is not null group
                by hop(interval '2 seconds', interval '10 seconds')
```
* waits for the pipeline to get to the RUNNING state
* waits for 10 successful checkpoints
* stops the job
* waits for it to successfully finish

Now that we have the ability to write integration tests, we can extend this:
* More cases (stop and re-starts, rescaling, recovery, etc.)
* More complex SQL
* Sources and sinks
* Correctness testing on output 

This PR also changes the CI to use debug mode to speed it up a bit. Since we're not using the artifacts produced by it at all, doing everything in release mode isn't necessary.

### DIFF
--- a/integ/src/main.rs
+++ b/integ/src/main.rs
@@ -19,8 +19,8 @@ mod embedded {
 
 const CONNECT_TIMEOUT: Duration = Duration::from_secs(30);
 
-pub fn run_service(name: &'static str, args: &[&str], env: Vec<(String, String)>) -> Result<()> {
-    let child = tokio::process::Command::new(name)
+pub fn run_service(name: String, args: &[&str], env: Vec<(String, String)>) -> Result<()> {
+    let child = tokio::process::Command::new(&name)
         .args(args)
         .envs(env)
         .kill_on_drop(true)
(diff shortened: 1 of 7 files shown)


integ/src/main.rs
```diff
@@ -0,0 +1,249 @@
...
+    // create a pipeline
+    let pipeline_name = format!("pipeline_{}", run_id);
+    info!("Creating pipeline {}", pipeline_name);
+    let pipeline_id = client
+        .create_pipeline(CreatePipelineReq {
+            name: pipeline_name.clone(),
+            config: Some(create_pipeline_req::Config::Sql(
+                arroyo_rpc::grpc::api::CreateSqlJob {
+                    query: format!(
+                        "select count(*) from {} where auction is not null group \
```
jacksonrnewhouse, line 176: Might be nice to factor this out into a test-case.

jacksonrnewhouse (approved):

Issue #75, opened 2023-04-24
Show backpressure in the UI


mwylde:
When a node in the job graph is not keeping up with its input data, its input queue fills up and its upstreams cannot continue sending it data until it pulls items from the queue. This _backpressures_ the upstream, which is blocked from doing work until it is able to write its messages downstream. In this way, we prevent faster upstreams from overloading slower downstreams. See [this blogpost](https://www.ververica.com/blog/how-flink-handles-backpressure) for more on the general theory of backpressure in streaming systems (although note the details are bit different in arroyo).

Figuring out whether and where backpressure is occurring is important for users to understand the behavior and performance of their pipelines.

In Arroyo, we have a metric `arroyo_worker_tx_queue_rem` that reports how much space remains in a task's transmit queue. When this is 0, that means that the *downstream* node is causing backpressure on us.

This data should be visible in the UI. We already have infrastructure to pass metrics back to the UI (which currently powers the data rate graphs) so this would involve extending that API to add the `arroyo_worker_tx_queue_rem` metric. For visualization, the simplest approach would be to color the nodes in the pipeline graph according to how backpressured they are (for example, as a fraction of the remaining queue size and total queue size).

The nodes in the graph represent logical operators, but in the physical execution each operator is subdivided into `N` parallel subtasks. Similarly, each operator may have `M` downstream nodes if the edge between them is a shuffle. The backpressure for an _operator_ will be some combination of the backpressure of its parallel _subtasks_ (median or min?).

So it will also be helpful to see the per-subtask backpressure, for example in the operator detail view that currently shows the data rate graphs.

PR #109, opened 2023-05-10
Add backpressure metric

jbeisen:
Add a new per-subtask metric that is the percentage the subtask is being backpressured. It's calculated from the remaining capacity of the queue.

This is a prerequisite for showing the backpressure in the UI (#75).

### DIFF
--- a/arroyo-worker/src/engine.rs
+++ b/arroyo-worker/src/engine.rs
@@ -181,7 +181,8 @@ pub struct Collector<K: Key, T: Data> {
     _ts: PhantomData<(K, T)>,
     sent_bytes: Option<IntCounter>,
     sent_messages: Option<IntCounter>,
-    tx_queue_size: Vec<Vec<Option<IntGauge>>>,
+    tx_queue_rem_gauges: Vec<Vec<Option<IntGauge>>>,
+    tx_queue_size_gauges: Vec<Vec<Option<IntGauge>>>,
 }
 
 impl<K: Key, T: Data> Collector<K, T> {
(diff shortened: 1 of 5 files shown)


arroyo-worker/src/engine.rs
```diff
@@ -215,10 +220,16 @@ impl<K: Key, T: Data> Collector<K, T> {
...
                 let idx = out_idx(&key, out_node_qs.len());
-                self.tx_queue_size[i][idx]
+                self.tx_queue_rem_gauges[i][idx]
                     .iter()
                     .for_each(|c| c.set(self.out_qs[i][idx].tx.capacity() as i64));
 
+                self.backpressure_gauges[i][idx]
+                    .iter()
+                    .for_each(|c| c.set(
+                        ((1.0 - self.out_qs[i][idx].tx.capacity() as f64 / QUEUE_SIZE as f64) * 100.0) as i64)
```
jbeisen, line 230: I'm calculating the _percentage full_ of the queue and multiplying it by 100 so we can use the `IntegerGauge`.  Seems like a lot of type converting, @mwylde what do you think of this?

mwylde: What about just reporting queue size, and then doing the math on the query side?

jbeisen: Currently in `metrics.rs` the `get_query` function hardcodes a simple `rate` query of a single metric. Are you suggesting modifying that to be able to do a more complex promql query that combines two metrics?

Or are you suggesting doing the calculation in the frontend?

jbeisen: I did the former.

mwylde (approved):

PR #110, opened 2023-05-11
Add backpressure graph and color pipeline nodes

jbeisen:
Add a graph in the OperatorDetail component that displays the backpressure metric.

Color the operator nodes in the pipeline graph according to the maximum backpressure of all its subtasks.

Resolves: #75

---

image

### DIFF
--- a/arroyo-console/src/routes/pipelines/JobDetail.tsx
+++ b/arroyo-console/src/routes/pipelines/JobDetail.tsx
@@ -59,6 +59,7 @@ import { ApiClient } from '../../main';
 import { PipelineOutputs } from './JobOutputs';
 import { CodeEditor } from './SqlEditor';
 import PipelineConfigModal from './PipelineConfigModal';
+import { getOperatorBackpressure } from '../../lib/util';
 
 interface JobDetailState {
   pipeline?: JobDetailsResp;
@@ -221,6 +222,7 @@ export function JobDetail({ client }: { client: ApiClient }) {
               <Box flex="1">
                 <PipelineGraph
                   graph={state.pipeline?.jobGraph}
+                  metrics={metrics.metrics}
                   setActiveOperator={setActiveOperator}
                   activeOperator={activeOperator}
                 />
@@ -357,6 +359,17 @@ function OperatorDetail({
     const node = graph.nodes.find(n => n.nodeId == operator_id);
     const node_metrics = metrics.metrics[operator_id];
 
+    const backpressure = getOperatorBackpressure(metrics, operator_id);
+
+    let backpressureBadge;
+    if (backpressure < 0.33) {
+      backpressureBadge = <Badge colorScheme={'green'}>LOW</Badge>;
+    } else if (backpressure < 0.66) {
+      backpressureBadge = <Badge colorScheme={'yellow'}>MEDIUM</Badge>;
+    } else {
+      backpressureBadge = <Badge colorScheme={'red'}>HIGH</Badge>;
+    }
+
     let msgRecv = 0;
     let msgSent = 0;
     let msgSentData;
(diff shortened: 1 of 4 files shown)


arroyo-console/src/routes/pipelines/JobDetail.tsx
```diff
@@ -399,14 +419,26 @@ function OperatorDetail({
           };
         });
       });
+
+      backpressureData = Object.entries(node_metrics.subtasks).map(kv => {
+        return kv[1].backpressure.map(m => {
+          return {
+            label: kv[0],
+            x: new Date(Number(m.time) / 1000),
+            y: m.value,
```
jbeisen, line 428: Note that I didn't add the randomness that the other metrics have. @mwylde what's up with that?

mwylde: It adds a bit of jitter that helps separate out the substasks on the graph

arroyo-console/src/routes/pipelines/JobDetail.tsx
```diff
@@ -422,6 +454,13 @@ function OperatorDetail({
           Events TX
           <TimeSeriesGraph data={msgSentData} timeWindowMs={5 * 60 * 1000} />
         </Box>
+
+        {backpressureData?.some(m => m.length) && (
```
mwylde, line 458: I'm not sure it makes sense to have this as a graph -- generally this doesn't vary over time much (at not on the 5 minute timescale of these graphs).

arroyo-console/src/routes/pipelines/JobGraph.tsx
```diff
@@ -40,16 +58,32 @@ function PipelineGraphNode({
...
   activeOperator?: string;
 }) {
   const nodeTypes = useMemo(() => ({ pipelineNode: PipelineGraphNode }), []);
 
   const nodes = graph.nodes.map(node => {
+    let operatorBackpressure = 0;
+    if (metrics) {
+      // reduce the subtasks backpressure to an operator-level backpressure
+      const nodeMetrics = metrics.metrics[node.nodeId];
+      const backPressureMetrics = Object.entries(nodeMetrics.subtasks).map(kv => {
```
mwylde, line 77: I think this is causing errors before metrics are available as the pipeline is starting up:
> TypeError: Cannot read properties of undefined (reading 'subtasks')

arroyo-console/src/routes/pipelines/JobGraph.tsx
```diff
@@ -9,13 +9,31 @@ import {
   JobMetricsResp,
 } from "../../gen/api_pb";
 
+
+function getNodeColor(backpressure: number): string {
+  // backPressure should be a value between 0 and 100
+
+  const thresholds = [33, 66];
```
mwylde, line 16: Something isn't working quite right here -- even when the pipeline is fully backpressured (i.e., the tx queues are at 0 capacity) I don't see any colors

mwylde: BTW, if you want to test backpressure, the easiest way to simulate it is to add a sleep in an operator:

```
tokio::time::sleep(Duration::from_secs(1).await
```

arroyo-console/src/routes/pipelines/JobDetail.tsx
jbeisen: Fair, maybe we should just make it a badge to indicate the current backpressure status.

arroyo-console/src/routes/pipelines/JobGraph.tsx
jbeisen: This is fixed now. I changed the logic for getting the current backpressure.

arroyo-console/src/routes/pipelines/JobDetail.tsx
jbeisen: I've removed the graph.

arroyo-console/src/routes/pipelines/JobGraph.tsx
jbeisen: This is fixed now.

arroyo-console/src/routes/pipelines/JobDetail.tsx
jbeisen: I removed the graph and added a badge instead.

mwylde (approved):

PR #111, opened 2023-05-11
Use mock service worker for mocking api responses

jbeisen:
This change adds the Mock Service Worker library to intercept and mock requests at the network level. This lets us build UI components without needing real responses from the api.

This includes some basic mocks for the GetJobs, GetJobDetails, and GetJobMetrics grpc calls.

### DIFF
--- a/arroyo-console/src/main.tsx
+++ b/arroyo-console/src/main.tsx
@@ -18,9 +18,10 @@ import { Connections } from './routes/connections/Connections';
 import { ConnectionEditor } from './routes/connections/CreateConnection';
 import { CreatePipeline } from './routes/pipelines/CreatePipeline';
 import { SinkEditor } from './routes/sinks/CreateSink';
-
 import { addCloudRoutes, createRoot, getClient, needsOrgSetup } from './lib/CloudComponents';
 import PageNotFound from './routes/not_found/PageNotFound';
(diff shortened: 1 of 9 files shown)


arroyo-console/src/main.tsx
```diff
@@ -101,5 +103,10 @@ const theme = extendTheme(proTheme, {
   config: config,
 });
 
+if (process.env.NODE_ENV === 'development') {
```
jbeisen, line 106: @mwylde this means that the mocks will apply with the service is run with `pnpm dev`. Is that ok?

mwylde: Is there a way to make that configurable? At least when I'm developing the frontend, I think I'll generally want to talk to a real arroyo api.

jbeisen: I think when developing the frontend there are some real benefits to _not_ connecting it to a real arroyo api until the very end before merging. The main benefit with mocking the API is that you can deliberately develop loading and error states without needing to hack the api to somehow produce those states (which is especially hard if those states IRL are rare or short-lived, like loading).

But since `NODE_ENV` is just an environment variable you can always override it like this if you really want to:
```
NODE_ENV=production pnpm dev
```

and we also have `pnpm preview`.

mwylde (approved):

Issue #25, opened 2023-04-11
Implement RegexpMatch and RegexpReplace


jacksonrnewhouse:
These functions come from DataFusion, and are equivalent to the postgresql regexp_match() and regexp_replace() functions. These are defined for Arrow arrays at https://github.com/apache/arrow-rs/blob/master/arrow-string/src/regexp.rs. Partial progress, such as a solution that does not support any of the optional flags would also be acceptable.

PR #122, opened 2023-05-15
Issue 25 : Adding support for regex_replace and regex_match

edmondop:
Solves #25 

This pull requests add regexp_match and regexp_replace as defined in DataFusion signature. 

Please note that the DataFusion signature is different from the PostgresSQL for what concerns the regexp_replace. In particular, the start and N parameters defined below are not supported
```

regexp_match('foobarbequebaz', '(bar)(beque)') → {bar,beque}

regexp_replace ( string text, pattern text, replacement text, start integer, N integer [, flags text ] ) → text

Replaces the substring that is the N'th match to the POSIX regular expression pattern, or all such matches if N is zero; see [Section 9.7.3](https://www.postgresql.org/docs/current/functions-matching.html#FUNCTIONS-POSIX-REGEXP).

regexp_replace('Thomas', '.', 'X', 3, 2) → ThoXas
```


arroyo-sql/src/expressions.rs
```diff
@@ -2041,8 +2085,7 @@ impl StringFunction {
                     })
                 }
             }
-            StringFunction::RegexpMatch(_, _) => todo!(),
-            StringFunction::RegexpReplace(_, _, _, _, _) => todo!(),
+            StringFunction::RegexpReplace(_, _, _, _) => unreachable!(),
```
jacksonrnewhouse, line 2088: I'd clarify this is the `Some(_)` in the fourth argument case here.

arroyo-worker/src/operators/functions/regexp.rs
```diff
@@ -0,0 +1,51 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+
+    use super::*;
+    use std::error::Error;
+
+    #[test]
+    pub fn test_regexp_match_is_correct() -> Result<(), Box<dyn Error>> {
```
jacksonrnewhouse, line 34: In general you don't need to have return types on your tests. Either it panics or it passes.

jacksonrnewhouse (commented):
Looks good, just a couple small notes and rebasing on master.

arroyo-sql/src/expressions.rs
edmondop: Fixed with https://github.com/ArroyoSystems/arroyo/pull/122/commits/9889c37a41d5d3ae6f678c500f5cbcacf754ad9e

arroyo-worker/src/operators/functions/regexp.rs
edmondop: Fixed with https://github.com/ArroyoSystems/arroyo/pull/122/commits/9889c37a41d5d3ae6f678c500f5cbcacf754ad9e 

I am normally more used to work with functional errors rather than panics, that's why I naturally leaned on using results

jacksonrnewhouse (approved):

PR #127, opened 2023-05-16
sql: Support virtual fields and overriding timestamp via event_time_field

jacksonrnewhouse:
This adds two complementary features: 
Virtual columns and overriding the timestamp with a SQL column. These are currently restricted to tables created in SQL with connections, but should be expanded upon in the future.

### Virtual Columns
We use the postgres syntax of `field_name field_type GENERATED ALWAYS AS ($EXPRESSION)`. The `$EXPRESSION` is will come through the SQL AST as a `sqlparser::ast::Expr`. This then needs to be converted to dataFusion's Expr, and then Arroyo's Expression. I don't love how disjoint this is from the normal dataflow, but we're restricted by DataFusion's capabilities right now. 

### Event Time Field
We also now support overriding the timestamp from a source configured within SQL. This is done by the `event_time_field` parameter in the `WITH` clause of the `CREATE TABLE` statement. If present it will validate that the column exists and is a timestamp. Currently the pipeline will panic if the timestamp column is null, something that should be configurable in the future.

### DIFF
--- a/arroyo-sql/src/expressions.rs
+++ b/arroyo-sql/src/expressions.rs
@@ -179,7 +179,7 @@ pub struct ExpressionContext<'a> {
 }
 
 impl<'a> ExpressionContext<'a> {
-    pub fn compile_expr(&mut self, expression: &Expr) -> Result<Expression> {
+    pub fn compile_expr(&self, expression: &Expr) -> Result<Expression> {
         match expression {
             Expr::Alias(expr, _alias) => self.compile_expr(expr),
             Expr::Column(column) => Ok(Expression::Column(ColumnExpression::from_column(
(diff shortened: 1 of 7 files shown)


jacksonrnewhouse:
Here's an example query that behaves as desired
```
CREATE TABLE top_one (
    row_number int,
    auction int,
    count int,
    time string,
    parsed_timestamp datetime GENERATED ALWAYS AS (CAST(time as TIMESTAMP)),
    )
    WITH (
        connection = 'local',
        topic = 'top_one',
        event_time_field = 'parsed_timestamp'
    );

SELECT row_number, auction, count FROM top_one
```

mwylde (approved):
Looks great! Can you also add this to the doc?

jacksonrnewhouse:
Will do

PR #140, opened 2023-05-28
Write eventsource and kafka source errors to db

jbeisen:
Create a job_log_messages table in the database. When an eventsource or
kafka source encounters an error, call the WorkerError rpc method on the
controller, which logs the error to the database.

### DIFF
--- a/arroyo-controller/src/lib.rs
+++ b/arroyo-controller/src/lib.rs
@@ -12,7 +12,8 @@ use arroyo_rpc::grpc::{
     WorkerFinishedResp,
 };
 use arroyo_rpc::grpc::{
-    SinkDataReq, SinkDataResp, TaskCheckpointEventReq, TaskCheckpointEventResp,
+    SinkDataReq, SinkDataResp, TaskCheckpointEventReq, TaskCheckpointEventResp, WorkerErrorReq,
+    WorkerErrorRes,
 };
 use arroyo_server_common::log_event;
 use arroyo_types::{from_micros, ports, DatabaseConfig, NodeId, WorkerId};
@@ -47,6 +48,7 @@ mod states;
 include!(concat!(env!("OUT_DIR"), "/controller-sql.rs"));
 
 use crate::schedulers::{nomad::NomadScheduler, NodeScheduler, ProcessScheduler, Scheduler};
+use types::public::LogLevel;
 use types::public::StopMode;
 
 pub const CHECKPOINTS_TO_KEEP: u32 = 5;
@@ -416,6 +418,31 @@ impl ControllerGrpc for ControllerServer {
             .push(tx);
 
         Ok(Response::new(ReceiverStream::new(rx)))
+    }
+
+    async fn worker_error(
+        &self,
+        request: Request<WorkerErrorReq>,
+    ) -> Result<Response<WorkerErrorRes>, Status> {
+        info!("Got worker error.");
+        let req = request.into_inner();
+        let client = self.db.get().await.unwrap();
+        match queries::controller_queries::create_job_log_message()
+            .bind(
+                &client,
+                &req.job_id,
+                &req.operator_id,
+                &(req.task_index as i64),
+                &LogLevel::error,
+                &req.message,
+                &req.details,
+            )
+            .one()
+            .await
+        {
+            Ok(_) => Ok(Response::new(WorkerErrorRes {})),
+            Err(err) => Err(Status::from_error(Box::new(err))),
+        }
     }
 }
 
--- a/arroyo-rpc/src/lib.rs
+++ b/arroyo-rpc/src/lib.rs
@@ -58,6 +58,12 @@ pub enum ControlResp {
         task_index: usize,
         error: String,
     },
+    Error {
+        operator_id: String,
+        task_index: usize,
+        message: String,
+        details: String,
+    },
 }
 
 pub struct FileAuthInterceptor {
--- a/arroyo-worker/src/engine.rs
+++ b/arroyo-worker/src/engine.rs
@@ -21,7 +21,7 @@ use arroyo_rpc::grpc::controller_grpc_client::ControllerGrpcClient;
 use arroyo_rpc::grpc::{
     CheckpointMetadata, HeartbeatReq, TableDeleteBehavior, TableDescriptor, TableType,
     TableWriteBehavior, TaskAssignment, TaskCheckpointCompletedReq, TaskCheckpointEventReq,
-    TaskFailedReq, TaskFinishedReq, TaskStartedReq,
+    TaskFailedReq, TaskFinishedReq, TaskStartedReq, WorkerErrorReq,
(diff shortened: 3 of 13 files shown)


arroyo-api/migrations/V6__add_job_log_messages.sql
```diff
@@ -0,0 +1,10 @@
+create type log_level as ENUM ('info', 'warn', 'error');
```
mwylde, line 1: Heads up that #144 added a V6 migration, so this will need to be bumped to V7

arroyo-rpc/src/lib.rs
```diff
@@ -58,6 +58,10 @@ pub enum ControlResp {
         task_index: usize,
         error: String,
     },
+    Error {
+        message: String,
```
mwylde, line 62: Should this also include the operation_id and task index?

arroyo-worker/src/engine.rs
```diff
@@ -1138,6 +1138,20 @@ impl Engine {
                                     None
                                 }
                             }
+                            Some(ControlResp::Error {message, details}) => {
+                                info!("Handling worker error");
```
mwylde, line 1142: this log line isn't that useful as is -- I'd either remove it, or have it log the details if they're not otherwise being logged

arroyo-worker/src/operators/sources/kafka/mod.rs
```diff
@@ -184,7 +184,12 @@ where
                             }
                         },
                         Err(err) => {
-                            error!("encountered error {}", err)
+                            ctx.control_tx.send(
```
mwylde, line 187: This should probably also capture some of the fatal errors, like failing to deserialize a message (line 180) or failing to create the consumer (167)

arroyo-controller/src/lib.rs
```diff
@@ -414,6 +416,28 @@ impl ControllerGrpc for ControllerServer {
...
+            .bind(
+                &client,
+                &req.job_id,
+                &LogLevel::error,
+                &req.message,
+                &req.details,
+            )
+            .one()
+            .await
+            .unwrap();
```
mwylde, line 437: This should probably return an error back to the client rather than panicking

mwylde (approved):

arroyo-rpc/src/lib.rs
jbeisen: Do you think those should be saved to the db and shown in the UI too?

mwylde: I think at least knowing the operator will be useful (in some cases it may be obvious, but in others it will not be -- imagine that the error is that the message failed to deserialize, but you have multiple sources)

jbeisen: Ok, I added both to the database

arroyo-controller/src/lib.rs
jbeisen: But if we're preserving the original behavior of the worker panicking, then what's the point of handling the error of failing to write the error message?

mwylde: This is within the controller, not the worker. The controller is responsible for all jobs, so shouldn't panic.

jbeisen: Gotcha. Ok I made it return an error using `Status::from_error`. Is that good or should we force it to be a specific error like `Status::Internal`?

arroyo-worker/src/operators/sources/kafka/mod.rs
jbeisen: 👍 done

arroyo-worker/src/engine.rs
jbeisen: Removed it

arroyo-worker/src/operators/sources/kafka/mod.rs
```diff
@@ -164,74 +164,111 @@ where
...
-                            if let Some(v) = msg.payload() {
-                                ctx.collector.collect(Record {
-                                    timestamp: from_millis(msg.timestamp().to_millis().unwrap() as u64),
-                                    key: None,
-                                    value: self.serialization_mode.deserialize_slice(&v).unwrap(),
-                                }).await;
-                                offsets.insert(msg.partition(), msg.offset());
-                                rate_limiter.until_ready().await;
+        let consumer = self.get_consumer(ctx).await;
+        match consumer {
```
jbeisen, line 168: @mwylde is this the way you had in mind to handle the error? It seems not so nice to have such a huge match statement but I'm not sure what the alternative is.

mwylde: Yeah, I wouldn't put everything in a giant match statement, since the error handling is something that's just relevant at the beginning. Instead I would put just the assignment in the match:

```rust
let consumer = match self.get_consumer(ctx).await {
  Ok(consumer) => consumer,
  Err(err) => {
        ctx.control_tx
            .send(ControlResp::Error {
                operator_id: ctx.task_info.operator_id.clone(),
                task_index: ctx.task_info.task_index.clone(),
                message: "Error while getting kafka consumer".to_string(),
                details: format!("{:?}", err),
            })
            .await
            .unwrap();
        panic!("Error while getting kafka consumer")
  }
```

arroyo-worker/src/operators/sources/kafka/mod.rs
```diff
@@ -164,74 +164,111 @@ where
...
+                ctx.control_tx
+                    .send(ControlResp::Error {
+                        operator_id: ctx.task_info.operator_id.clone(),
+                        task_index: ctx.task_info.task_index.clone(),
+                        message: "Error while getting kafka consumer".to_string(),
+                        details: format!("{:?}", err),
+                    })
+                    .await
+                    .unwrap();
+                panic!("Error while getting kafka consumer")
```
mwylde, line 270: this should also report the actually error (so add `err` to the panic)

arroyo-worker/src/operators/sources/kafka/mod.rs
```diff
@@ -164,74 +164,111 @@ where
...
+                                                    timestamp: from_millis(msg.timestamp().to_millis().unwrap() as u64),
+                                                    key: None,
+                                                    value: deserialized,
+                                                }).await;
+                                            },
+                                            Err(err) => {
+                                                 ctx.control_tx.send(
+                                                    ControlResp::Error {
+                                                        operator_id: ctx.task_info.operator_id.clone(),
+                                                        task_index: ctx.task_info.task_index.clone(),
```
mwylde, line 191: clone is not necessary here, since task_index is a u16 which implements copy

arroyo-controller/src/lib.rs
mwylde: I would create a Status::Internal error with an appropriate message. It doesn't matter that much here, since the consumers are internal, but it's a good practice.

PR #143, opened 2023-05-29
Return errors if CREATE TABLE statements are misused.

jacksonrnewhouse:
Adds some validation so that we error if

1. A non-supported value in the WITH portion of CREATE TABLE is specified. These are likely to be typos and better to error than swallow.
2. Fail if the query tries to insert into a MemoryTableWithConnectionConfig with virtual fields.


arroyo-sql/src/lib.rs
```diff
@@ -446,6 +446,15 @@ impl<'a> SqlProgramBuilder<'a> {
             let name = name.to_string();
             let mut with_map = HashMap::new();
             for option in with_options {
+                if !matches!(
+                    option.name.value.to_lowercase().as_str(),
+                    "connection" | "topic" | "format" | "event_time_field" | "watermark_field"
```
mwylde, line 451: This will break HTTP sources. Generally I don't think it makes sense to have a fixed list here, since every connector will have its own options.

jacksonrnewhouse: I adjusted this to be source-specific. Still feels like a good idea to have validation on this.

jacksonrnewhouse:
This is getting implemented in #184.

PR #147, opened 2023-05-30
chore: simplify the code

chenquan:


arroyo-api/src/optimizations.rs
```diff
@@ -44,39 +44,43 @@ fn remove_in_place(graph: &mut DiGraph<StreamNode, StreamEdge>, node: NodeIndex)
...
-                        }
+                let mut ins = graph.edges_directed(idx, Incoming);
+                if !(no_shuffles && in_degree == 1) {
+                    continue;
+                }
+
+                let source_idx = ins.next().unwrap().source();
+                let in_node = graph.node_weight_mut(source_idx).unwrap();
+                if let Operator::Window { agg, .. } = &mut in_node.operator {
+                    if !agg.is_none() {
```
jacksonrnewhouse, line 67: It might be cleaner as 
```suggestion
                    if agg.is_some() {
```

jacksonrnewhouse (commented):
Thanks for opening this! Can you run `cargo fmt` on it? We have a check that requires the project to be formatted with it.

arroyo-api/src/optimizations.rs
chenquan: That's a good idea.

chenquan:
> Thanks for opening this! Can you run `cargo fmt` on it? We have a check that requires the project to be formatted with it.

done.

jacksonrnewhouse (approved):

PR #149, opened 2023-05-31
Add Errors tab to job details page

jbeisen:
Add a tab that displays the latest error for each operator/subtask pair,
if one exists. The tab uses the OperatorErrors component, which calls
the GetOperatorErrors rpc method, which queries the job_log_messages
table to get the errors.

---

image

### DIFF
--- a/arroyo-console/src/routes/pipelines/JobsIndex.tsx
+++ b/arroyo-console/src/routes/pipelines/JobsIndex.tsx
@@ -34,6 +34,7 @@ import { DeleteJobReq, GetJobsReq, JobStatus } from '../../gen/api_pb';
 import React, { useEffect, useRef, useState } from 'react';
 import { FiCopy, FiXCircle } from 'react-icons/fi';
 import { ApiClient } from '../../main';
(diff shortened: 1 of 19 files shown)


mwylde (approved):

arroyo-console/src/routes/pipelines/JobsIndex.tsx
```diff
@@ -84,10 +85,7 @@ const columns: Array<ColumnDef> = [
       if (s.startTime == null) {
         return '-';
       } else {
-        return new Intl.DateTimeFormat('en', {
-          dateStyle: 'short',
-          timeStyle: 'short',
-        }).format(new Date(Number(s.startTime) / 1000));
+        return formatDate(s.startTime);
```
mwylde, line 88: This doesn't seem correct—I"m now seeing dates like `4/28/50, 11:08 PM`

mwylde: ah, because formatDate is missing the /1000

PR #150, opened 2023-05-31
add docker.yaml to build and push docker images.

jacksonrnewhouse:
This builds all 8 images. Currently it is restricted to semver tags and branches with "docker" in the name. I'm using BuildJet. There's likely work to be done to improve build times. Here's an example build run off of this branch: https://github.com/ArroyoSystems/arroyo/actions/runs/5136599976/jobs/9243549066


mwylde (approved):
awesome!

PR #151, opened 2023-05-31
Remove AggregatingStrategy, as that is handled differently.

jacksonrnewhouse:
We had mainly removed this from the SqlOperator, but there were corner cases that could fail.


mwylde:
needs `cargo fmt`

mwylde (approved):

PR #152, opened 2023-06-01
point k8s at latest

jacksonrnewhouse:
Also sets the K8S_IMAGE variable for the Helm chart.

### DIFF
--- a/k8s/arroyo/values.yaml
+++ b/k8s/arroyo/values.yaml
@@ -15,7 +15,7 @@ api:
   image:
     repository: ghcr.io/arroyosystems/arroyo-services
     pullPolicy: IfNotPresent
-    tag: "amd64"
+    tag: "latest"
 
 controller:
   resources:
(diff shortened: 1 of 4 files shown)


k8s/arroyo/values.yaml
```diff
@@ -15,7 +15,7 @@ api:
   image:
     repository: ghcr.io/arroyosystems/arroyo-services
     pullPolicy: IfNotPresent
-    tag: "amd64"
+    tag: "latest"
```
mwylde, line 18: is this tag the one that will pull the correct image for your arch?

jacksonrnewhouse: We're not currently automatically pushing to latest. I think we can either do it manually when we cut a release or setup a separate task. My inclination is to not build on every merge. Maybe every tag like `[0-9]+.[0-9]+.[0-9]`, but then publishing an old tag would roll back.

mwylde: I actually was asking what's the difference between `multi-arch` and `latest`. Is `latest` pointing to a particular architecture, or is it the multi-arch manifest?

jacksonrnewhouse: I think we can publish a latest manifest. Alternately, we could keep it pinned to an explicit version.

mwylde (approved):

PR #160, opened 2023-06-06
Improve state machine robustness

mwylde:
This PR improves the robustness of the controller state machine in a few ways:

* Removes a potential panic in the process scheduler, when the pipeline binaries can't be found (now, we will attempt to recompile)
* Adds support for stopping the pipeline in more states (compiling, scheduling, and rescaling)


jacksonrnewhouse (approved):
Looks good, one typo in comments.

PR #164, opened 2023-06-08
POC: Use grpc-gateway to make a reverse-proxy RESTful api

jbeisen:
This is a proof of concept to show how we can use [grpc-gateway](https://github.com/grpc-ecosystem/grpc-gateway) to expose a RESTful api that acts as reverse-proxy to the existing grpc api.

One issue that we'll need to solve if we go this route is that we're using buf to compile the protobufs and generate the reverse-proxy server (by running `buf generate`), but cargo is set up to use protoc. We'll need to settle on one.


arroyo-rpc/proto/api.proto
```diff
@@ -849,8 +868,15 @@ message OutputData {
...
 
+
+
 service ApiGrpc {
-  rpc CreateConnection(CreateConnectionReq) returns (CreateConnectionResp);
+  rpc CreateConnection(CreateConnectionReq) returns (CreateConnectionResp) {
+    option (google.api.http) = {
+      post: "/v1/connections"
+      body: "*"
+    };
```
jbeisen, line 878: This is how we define routes and their parameters.

Issue #28, opened 2023-04-12
Implement date_trunc(), date_part(), and extract()


jacksonrnewhouse:
The input will be a possibly optional SystemTime, as that is the representation for timestamps currently. Follow the documentation from data fusion for [date_trunc](https://arrow.apache.org/datafusion/user-guide/sql/scalar_functions.html#date-trunc), [date_part](https://arrow.apache.org/datafusion/user-guide/sql/scalar_functions.html#date-part), and [extract](https://arrow.apache.org/datafusion/user-guide/sql/scalar_functions.html#extract).

The most natural place is a new variant on Expression for DateFunction(), with a corresponding DateExpression. Testing should be done by adding additional assertions to arroyo-sql-testing.

AK-ayush:
Hi @jacksonrnewhouse, I'd like to work on this and #25 if that's okay with you?

jacksonrnewhouse:
@AK-ayush, that'd be great. If you have any trouble getting your dev environment setup or figuring out some of the moving parts of arroyo-sql, feel free to reach out in the Discord.

kamalesh0406:
Hello, I can also help out with this issue. @AK-ayush would you be willing to split the implementation of each function between us? I have the bandwidth to implement `date_part`. Thanks!!

AK-ayush:
Sure @kamalesh0406, please go ahead.

edmondop:
Can I move forward with this? It sees there is no pr open

kamalesh0406:
@edmondop Sure you can look into it. I was only able to partially complete it and I don't think so I have the time to work on it right now :) .

jacksonrnewhouse:
Handled by #181.

PR #181, opened 2023-06-21
Implement date_part, date_trunc and extract

edmondop:
Solves #28 . As confirmed on the DataFusion channel, the EXTRACT is rewritten by the query planner into a date_trunc, so a SQL test is present but no function in the worker and no enum in the arroyo-sql for the `DateTimeFunction`

### DIFF
--- a/arroyo-sql/src/expressions.rs
+++ b/arroyo-sql/src/expressions.rs
@@ -2670,17 +2670,26 @@ impl DateTimeFunction {
 
     fn to_syn_expression(&self) -> syn::Expr {
         let function = self.non_null_function_invocation();
-        let (expr1, expr2): (syn::Expr, syn::Expr) = match self {
+        let (null_arg, expr1, expr2): (bool, syn::Expr, syn::Expr) = match self {
             DateTimeFunction::DatePart(arg1, arg2) => {
                 let arg2 = format!("arroyo_types::DatePart::{:?}", arg2);
-                (arg1.to_syn_expression(), parse_str(&arg2).unwrap())
+                (
+                    arg1.nullable(),
+                    arg1.to_syn_expression(),
+                    parse_str(&arg2).unwrap(),
+                )
             }
             DateTimeFunction::DateTrunc(arg1, arg2) => {
                 let arg2 = format!("arroyo_types::DateTruncPrecision::{:?}", arg2);
-                (arg1.to_syn_expression(), parse_str(&arg2).unwrap())
+                (
+                    arg1.nullable(),
+                    arg1.to_syn_expression(),
+                    parse_str(&arg2).unwrap(),
+                )
             }
         };
-        parse_quote!({
+        match null_arg {
+            true => parse_quote!({
                 use arroyo_types::{DatePart, DateTruncPrecision};
                 if let Some(arg1) = #expr1 {
                     let arg2 = #expr2;
@@ -2688,8 +2697,14 @@ impl DateTimeFunction {
                 } else {
                     None
                 }
-            }
-        )
+            }),
+            false => parse_quote!({
+                use arroyo_types::{DatePart, DateTruncPrecision};
+                let arg1 = #expr1;
+                let arg2 = #expr2;
+                #function
+            }),
+        }
     }
 
     fn return_type(&self) -> TypeDef {
--- a/arroyo-sql/src/expressions.rs
+++ b/arroyo-sql/src/expressions.rs
@@ -2654,12 +2654,6 @@ impl TryFrom<(BuiltinScalarFunction, Vec<Expression>)> for DateTimeFunction {
     }
 }
 
-fn enum_to_syn<T: serde::Serialize + std::fmt::Debug>(t: &T, prefix: &str) -> syn::Expr {
-    let s = serde_json::to_string(t).unwrap();
-    let s = format!("{}::{}", prefix, &s[1..s.len() - 1]); // Remove the quotation marks
-    syn::parse_str(&s).unwrap()
-}
-
 impl DateTimeFunction {
     fn non_null_function_invocation(&self) -> syn::Expr {
         match self {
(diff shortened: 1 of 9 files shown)


arroyo-sql/src/expressions.rs
```diff
@@ -2593,3 +2604,104 @@ impl CaseExpression {
...
+                ))
+            }
+        }
+    }
+
+    fn to_syn_expression(&self) -> syn::Expr {
+        let function = self.non_null_function_invocation();
+        let (expr1, expr2) = match self {
+            DateTimeFunction::DatePart(arg1, arg2) => {
+                (arg1.to_syn_expression(), enum_to_syn(arg2, "DatePart"))
```
jacksonrnewhouse, line 2683: Rather than enum_to_syn, which is kinda hacky, you could do something like
```suggestion
                let arg2 = format!("{:?}", arg2);
                (arg1.to_syn_expression(), parse_quote!(#arg2))
```
Because the enums implement debug, they should print out correctly.

arroyo-sql/src/expressions.rs
```diff
@@ -2593,3 +2604,104 @@ impl CaseExpression {
...
+                let arg2 = #expr2;
+                #function
+            }
+        )
+    }
+
+    fn return_type(&self) -> TypeDef {
+        match self {
+            DateTimeFunction::DateTrunc(_, _) => {
+                TypeDef::DataType(DataType::Timestamp(TimeUnit::Microsecond, None), false)
```
jacksonrnewhouse, line 2702: We've been kinda cavalier on which TimeUnit we use, but I think it'd be good to standardize on Nanosecond.

arroyo-sql/src/expressions.rs
```diff
@@ -2593,3 +2604,104 @@ impl CaseExpression {
...
+        let function = self.non_null_function_invocation();
+        let (expr1, expr2) = match self {
+            DateTimeFunction::DatePart(arg1, arg2) => {
+                (arg1.to_syn_expression(), enum_to_syn(arg2, "DatePart"))
+            }
+            DateTimeFunction::DateTrunc(arg1, arg2) => (
+                arg1.to_syn_expression(),
+                enum_to_syn(arg2, "DateTruncPrecision"),
+            ),
+        };
```
jacksonrnewhouse, line 2689: I'd recommend rather than the intricacies of `enum_to_syn()`, just using the debug formatting. Something like
```suggestion
                let (expr1, expr2) : (syn::Expr, syn::Expr) = match self {
            DateTimeFunction::DatePart(arg1, arg2) => {
                let arg2 = format!("arroyo_types::DatePart::{:?}", arg2);
                (arg1.to_syn_expression(), parse_str(&arg2).unwrap())
            }
            DateTimeFunction::DateTrunc(arg1, arg2) => {
                let arg2 = format!("arroyo_types::DateTruncPrecision::{:?}", arg2);
                (arg1.to_syn_expression(), parse_str(&arg2).unwrap())
            },
        };    
```

arroyo-sql/src/expressions.rs
```diff
@@ -2593,3 +2604,104 @@ impl CaseExpression {
...
+        parse_quote!({
+                use arroyo_types::{DatePart, DateTruncPrecision};
+                let arg1 = #expr1;
+                let arg2 = #expr2;
+                #function
+            }
+        )
+    }
+
+    fn return_type(&self) -> TypeDef {
```
jacksonrnewhouse, line 2699: This doesn't currently support null timestamps. The nullity of the result should depend on the nullity of the second argument.

arroyo-sql/src/expressions.rs
```diff
@@ -2593,3 +2604,104 @@ impl CaseExpression {
...
     }
 }
+
+#[derive(Debug, Clone)]
+pub enum DateTimeFunction {
+    DateTrunc(Box<Expression>, Box<DateTruncPrecision>),
+    DatePart(Box<Expression>, Box<DatePart>),
+}
+
+fn convert_expression_to<T>(expr: Expression, expr_name: &str) -> Result<T, anyhow::Error>
```
jacksonrnewhouse, line 2614: I think this function is doing too much. I'd split it into two methods. One that extracts a literal string from an Expression, and then call the specific try_into() on the String.

jacksonrnewhouse (changes requested):
Looking pretty good. Just a few pieces of feedback.

arroyo-sql/src/expressions.rs
edmondop: Addressed with https://github.com/ArroyoSystems/arroyo/pull/181/commits/764bd3d36a3117a579589c4a7d73cae634f3db81

arroyo-sql/src/expressions.rs
edmondop: Addressed with https://github.com/ArroyoSystems/arroyo/pull/181/commits/764bd3d36a3117a579589c4a7d73cae634f3db81

arroyo-sql/src/expressions.rs
edmondop: Addressed with https://github.com/ArroyoSystems/arroyo/pull/181/commits/764bd3d36a3117a579589c4a7d73cae634f3db81

arroyo-sql/src/expressions.rs
edmondop: Addressed with https://github.com/ArroyoSystems/arroyo/pull/181/commits/46171ecfe07f9eaf77e9682828e3dc92e3bd77f6

arroyo-sql/src/expressions.rs
edmondop: Actually, sorry, addressed with https://github.com/ArroyoSystems/arroyo/pull/181/commits/2ea64ea338ba3fdac714a78457a594c595e14fb0

jacksonrnewhouse (approved):
Looks great!

PR #184, opened 2023-06-27
Connectors redesign

mwylde:
This PR is a complete redesign of how sources and sinks work in Arroyo. Currently, details about each source or sink (which we'll generically call _connectors_, since they typically connect Arroyo to an external system) are hard-coded in many parts of our system, including:
* The web console
* API
* Database
* SQL layer
* Compiler
* Pipeline representation
* Worker code

This makes the process of adding a new connector very difficult and tedious, and makes it very had to change common details about how connectors interact with the rest of the system. For example, we previously added DDL support for creating connectors in SQL but it only worked for Kafka.

This PR attempts to solve that problem by genericizing all of the interactions between connectors and the rest of the system—in other words, none of those layers will know about kafka or server-sent events, they just need to know about "connectors" as a generic concept.

All of the connector-specific logic is now centralized into a single struct in the new arroyo-connectors crate, which implements the `Connector` trait.

The metadata for that connector is sent to the web ui to build the configuration interfaces and determines how the connector is constructed via SQL DDL statements:

image

I've also reworked the console to reduce the emphasis on the connection/source/sink distinction, which was confusing for users. Now there is a single flow to create a source or sink that starts with the type of connection the user would like to create (like Kafka) without requiring the user to juggle multiple types of data.

--- 

This is still a WIP. Some known issues:
* ~Nexmark sources can't be created via SQL yet, as we need to figure out how to handle SQL schemas for connectors with fixed schemas~ this is going to stay out of scope for this PR, as it requires better datafusion struct support (or possibly using `CREATE TABLE EXTERNAL`)
* ~Impulse source hasn't been migrated yet~
* ~RawJson schemas aren't properly inferred~
* There's no way to delete connections or ~view their configuration~
* Update tables are not well tested and likely have issues
* ~Lots of tests need to be updated~
* ~Connector icons aren't available yet in the UI~


arroyo-sql/src/tables.rs
```diff
@@ -0,0 +1,616 @@
...
+            let connector = with_map.remove("connector");
+
+            match connector.as_ref().map(|c| c.as_str()) {
+                Some("memory") | None => {
+                    if fields.iter().any(|f| f.expression.is_some()) {
+                        bail!("Virtual fields are not supported in memory tables; instead write a query");
+                    }
+
+                    if !with_map.is_empty() {
+                        bail!("Memory tables do not allow any with options; to create a connector table set the 'connector' option");
```
jacksonrnewhouse, line 494: This triggers for a query like
```
CREATE TABLE bid_count (
  bids bigint
) WITH (
    connector = 'memory',
    format = 'json'
);
INSERT INTO bid_count
SELECT count(*) as bids FROM small_nexmark where bid is not null;
SELECT * FROM bid_count
```
And the error message is confusing
> Memory tables do not allow any with options; to create a connector table set the 'connector' option

arroyo-connectors/src/impulse.rs
```diff
@@ -0,0 +1,178 @@
...
+        let message_count: Option<i64> = options
+            .remove("message_count")
+            .map(|t| i64::from_str(&t))
+            .transpose()
+            .map_err(|_| anyhow!("invalid value for event_time_interval; expected float"))?;
+
+        // validate the schema
+        if let Some(s) = s {
+            if s.fields != impulse_schema().fields {
+                bail!("invalid schema for impulse source");
```
jacksonrnewhouse, line 118: Might be good to return the expected schema. It's also currently very strict. The field order needs to be (counter, subtask_index), and the type has to be`bigint unsigned`.

arroyo-connectors/src/kafka.rs
```diff
@@ -0,0 +1,399 @@
...
+                    Some(Err(e)) => {
+                        warn!("Error while reading from kafka in test: {:?}", e);
+                        return Err(format!("Error while reading messages from Kafka: {}", e));
+                    }
+                    None => {
+                        // wait
+                    }
+                }
+
+                tokio::time::sleep(Duration::from_millis(500)).await;
```
jacksonrnewhouse, line 331: Could just move this into the None block.

jacksonrnewhouse (commented):
Just a couple small notes, should be good to merge later today.

arroyo-console/src/routes/connections/ChooseConnector.tsx
```diff
@@ -0,0 +1,93 @@
...
+} from '@chakra-ui/react';
+import { ApiClient } from '../../main';
+import { useNavigate } from 'react-router-dom';
+import { useConnectors } from '../../lib/data_fetching';
+
+export function ChooseConnector({ client }: { client: ApiClient }) {
+  const navigate = useNavigate();
+  const { connectors } = useConnectors(client);
+
+  return (
```
jbeisen, line 27: I think a good pattern to follow is to define several smaller pieces separately and use the return statement just to put them together at a high level. It makes it much more readable and easier to deal with. [CreatePipeline.tsx](https://github.com/ArroyoSystems/arroyo/blob/master/arroyo-console/src/routes/pipelines/CreatePipeline.tsx) is a good example of that.

jbeisen:
I tried creating an eventsource connection with Mastodon and am getting an error
>Received error from server: UnexpectedResponse(404)

jbeisen:
What do you think of adding a short README when we add a new crate? Just one or two sentences would be helpful for new developers.

arroyo-console/src/routes/connections/CreateConnection.tsx
```diff
@@ -1,462 +1,176 @@
...
-      icon: SiApachekafka,
-      description: 'Confluent Cloud, Amazon MSK, or self-hosted',
-      initialState: new KafkaConnection({
-        authConfig: new KafkaAuthConfig({
-          authType: { case: 'noAuth', value: new NoAuth({}) },
-        }),
-      }),
-      editor: <ConfigureKafka state={state} setState={setState} setReady={setReady} />,
-      disabled: false,
+      title: 'Configure connection',
```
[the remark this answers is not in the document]
mwylde: That's called a "placeholder" -- it shows an example, not a default value

arroyo-sql/src/tables.rs
```diff
@@ -0,0 +1,616 @@
...
+            SerializationMode::DebeziumJson => ProcessingMode::Update,
+            _ => ProcessingMode::Append,
+        }
+    }
+
+    pub fn as_sql_source(&self) -> Result<SqlOperator> {
+        match self.connection_type {
+            ConnectionType::Source => {}
+            ConnectionType::Sink => {
+                bail!("cannot read from sink")
```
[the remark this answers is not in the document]
mwylde: I've added map_errs to the calls of as_sql_sink and as_sql_source to include the table name, so this is handled holistically

arroyo-sql/src/tables.rs
mwylde: I think you suggested supporting `connector = 'memory'` — do you have a suggestion for how to handle this?

arroyo-connectors/src/impulse.rs
mwylde: Yeah, it would be good, but with our current error reporting capabilities in the UI we only support one or two lines of text

I think the real answer here is CREATE EXTERNAL so users don't have to specify a schema for connectors that have a specific schema

arroyo-connectors/src/kafka.rs
mwylde: sure

mwylde:
> I tried creating an eventsource connection with Mastodon and am getting an error
> 
> > Received error from server: UnexpectedResponse(404)

@jbeisen can you try with the configuration in the updated tutorial:
https://github.com/ArroyoSystems/arroyo-docs/blob/04f3aa6595809fa6d0357b17921aaabcbac3cae8/tutorial/mastodon.mdx

arroyo-console/src/routes/connections/CreateConnection.tsx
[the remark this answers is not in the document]
mwylde: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#placeholder

jacksonrnewhouse (approved):

arroyo-sql/src/tables.rs
mwylde: I updated it to show different error messages for the `connector = 'memory'` case and the no-connector case

PR #188, opened 2023-07-01
Websocket source

mwylde:
This PR introduces a POC websocket connector.

This allows interacting with many websocket APIs, for example coinbase:

```sql
CREATE TABLE coinbase (
    type TEXT,
    price TEXT,
) WITH (
    connector = 'websocket',
    endpoint = 'wss://ws-feed.exchange.coinbase.com',
    subscription_message = '{
    "type": "subscribe",
    "product_ids": [
        "ETH-USD",
        "BTC-USD"
    ],
    "channels": ["ticker"]
}',
    format = 'json'
);

SELECT CAST(price as FLOAT) from coinbase
WHERE type = 'ticker';
```

Websocket is a very open-ended protocol, which makes it challenging to target from a configuration approach. This initial POC allows a single subscription message to be configured, but doesn't support any complex behavior around tracking consumed messages / reporting consumption on reconnection / etc.

It is currently functional enough to support easy exploration of many real-world websocket APIs.

### DIFF
--- a/arroyo-connectors/src/websocket.rs
+++ b/arroyo-connectors/src/websocket.rs
@@ -1,0 +1,117 @@
+use anyhow::{anyhow};
+use arroyo_rpc::grpc::{
+    self,
+    api::{ConnectionSchema, TestSourceMessage},
+};
+use tokio::sync::mpsc::Sender;
+use tonic::Status;
+use typify::import_types;
+
+use serde::{Deserialize, Serialize};
+
+use crate::{
+    pull_opt, serialization_mode, Connection, ConnectionType, EmptyConfig, OperatorConfig,
+};
+
+use super::Connector;
+
+const TABLE_SCHEMA: &str = include_str!("../../connector-schemas/websocket/table.json");
+
+import_types!(schema = "../connector-schemas/websocket/table.json");
+const ICON: &str = include_str!("../resources/sse.svg");
+
+pub struct WebsocketConnector {}
+
+impl Connector for WebsocketConnector {
+    type ConfigT = EmptyConfig;
+
+    type TableT = WebsocketTable;
+
+    fn name(&self) -> &'static str {
+        "websocket"
+    }
+
+    fn metadata(&self) -> grpc::api::Connector {
+        grpc::api::Connector {
+            id: "websocket".to_string(),
+            name: "Websocket".to_string(),
+            icon: ICON.to_string(),
+            description: "Connect to a Websocket server".to_string(),
+            enabled: true,
+            source: true,
+            sink: false,
+            testing: true,
+            custom_schemas: true,
+            connection_config: None,
+            table_config: TABLE_SCHEMA.to_owned(),
+        }
+    }
+
+    fn test(
+        &self,
+        _: &str,
+        _: Self::ConfigT,
+        _: Self::TableT,
+        _: Option<&ConnectionSchema>,
+        tx: Sender<Result<TestSourceMessage, Status>>,
+    ) {
+        tokio::task::spawn(async move {
+            tx.send(Ok(TestSourceMessage { error: false, done: true, message: "Yay".to_string() })).await.unwrap();
+        });
+    }
+
+    fn table_type(&self, _: Self::ConfigT, _: Self::TableT) -> grpc::api::TableType {
+        return grpc::api::TableType::Source;
+    }
+
+    fn from_config(
+        &self,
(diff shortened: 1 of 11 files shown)


arroyo-connectors/src/websocket.rs
```diff
@@ -0,0 +1,203 @@
...
+    fn test(
+        &self,
+        _: &str,
+        _: Self::ConfigT,
+        table: Self::TableT,
+        _: Option<&ConnectionSchema>,
+        tx: Sender<Result<TestSourceMessage, Status>>,
+    ) {
+        tokio::task::spawn(async move {
+            let send = |error: bool, done: bool, message: String| {
```
jacksonrnewhouse, line 63: These send calls were a bit hard to figure out until I checked it out. Not a big deal.

arroyo-worker/src/connectors/websocket.rs
```diff
@@ -0,0 +1,237 @@
...
+    _t: PhantomData<(K, T)>,
+}
+
+#[source_fn(out_k = (), out_t = T)]
+impl<K, T> WebsocketSourceFunc<K, T>
+where
+    K: DeserializeOwned + Data,
+    T: DeserializeOwned + Data,
+{
+    pub fn from_config(config: &str) -> Self {
```
jacksonrnewhouse, line 54: Should this return a result? I realize that'd be a general change to the API, likely requiring make_graph() to also return a result.

arroyo-worker/src/connectors/websocket.rs
```diff
@@ -0,0 +1,237 @@
...
+                                        tx.send(tungstenite::Message::Pong(d)).await
+                                            .map(|_| None)
+                                            .map_err(|e| UserError::new("Failed to sned pong to websocket server", e.to_string()))
+                                    },
+                                    tungstenite::Message::Pong(_) => {
+                                        // ignore
+                                        Ok(None)
+                                    },
+                                    tungstenite::Message::Close(_) => {
+                                        ctx.report_error("Received close frame from server".to_string(), "".to_string()).await;
```
jacksonrnewhouse, line 181: It's not obvious to me this is an error.

arroyo-worker/src/connectors/websocket.rs
```diff
@@ -0,0 +1,237 @@
...
+                                    tungstenite::Message::Text(t) => {
+                                        self.serialization_mode.deserialize_str(&t).map(|t| Some(t))
+                                    },
+                                    tungstenite::Message::Binary(bs) => {
+                                        self.serialization_mode.deserialize_slice(&bs).map(|t| Some(t))
+                                    },
+                                    tungstenite::Message::Ping(d) => {
+                                        tx.send(tungstenite::Message::Pong(d)).await
+                                            .map(|_| None)
+                                            .map_err(|e| UserError::new("Failed to sned pong to websocket server", e.to_string()))
```
jacksonrnewhouse, line 174: typo:
```suggestion
                                            .map_err(|e| UserError::new("Failed to send pong to websocket server", e.to_string()))
```

jacksonrnewhouse (approved):

arroyo-worker/src/connectors/websocket.rs
mwylde: Right now we don't have any way to report errors when constructing the graph -- we assume that things are setup correctly by the compiler and that any errors are caught before that happens. Invalid config at this point could only be caused by a version mismatch between the api/controller and the worker code or a bug in the system, not due to user error.

arroyo-worker/src/connectors/websocket.rs
```diff
@@ -0,0 +1,237 @@
...
+        let (mut tx, mut rx) = ws_stream.split();
+
+        if let Some(msg) = &self.subscription_message {
+            if let Err(e) = tx.send(tungstenite::Message::Text(msg.clone())).await {
+                ctx.report_error(
+                    "Failed to send subscription message to websocket server".to_string(),
+                    e.to_string(),
+                )
+                .await;
+                panic!(
```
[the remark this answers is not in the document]
mwylde: yes, the async/await is synchronous from the perspective of the code.

arroyo-worker/src/connectors/websocket.rs
mwylde: it's an error from the perspective of a long-running query; if the server tells us to go away we don't have a reliable way to continue from there

PR #189, opened 2023-07-02
Create REST api

jbeisen:
Add a REST api with a single endpoint - GET /v1/ping. Genereate an openapi spec and expose a swagger UI.

This also adds a new crate arroyo-openapi that writes the openapi spec to a file, which is used by arroyo-console to generate typescript interfaces for the api components.


mwylde (approved):
Looks great!

PR #191, opened 2023-07-06
Add pipeline endpoints to rest api

jbeisen:
Add endpoints for managing pipelines to the REST api. Creating a
pipeline automatically creates a single job for the pipeline.

Note that the REST API depends on the gRPC API so that it can use the
'start_or_preview' function in the POST /v1/pipelines endpoint. When we
are ready to deprecate the gRPC API we can move that logic into
pipelines.rs.


arroyo-types/src/api.rs
```diff
@@ -0,0 +1,74 @@
+use serde::{Deserialize, Serialize};
+use utoipa::ToSchema;
+
+#[derive(Serialize, Deserialize, Clone, ToSchema, Debug)]
+#[serde(rename_all = "camelCase")]
+pub struct PipelinePost {
+    pub name: String,
+    pub definition: Option<String>,
+    pub udfs: Option<Vec<Udf>>,
```
mwylde, line 9: I think this can just be `Vec<Udf>` and the client can pass an empty array if they're not defining any; generally you'd only want to wrap Vec in Option if there's an semantic difference between the empty list and not having it be set.

arroyo-types/src/api.rs
```diff
@@ -0,0 +1,74 @@
+use serde::{Deserialize, Serialize};
+use utoipa::ToSchema;
+
+#[derive(Serialize, Deserialize, Clone, ToSchema, Debug)]
+#[serde(rename_all = "camelCase")]
+pub struct PipelinePost {
+    pub name: String,
+    pub definition: Option<String>,
```
mwylde, line 8: I'd would make this `pub query: String`.  Eventually we will support other ways of defining jobs but right now all we will support in the API is SQL queries.

arroyo-types/src/api.rs
```diff
@@ -0,0 +1,74 @@
...
+    pub start_time: Option<u64>,
+    pub finish_time: Option<u64>,
+    pub tasks: Option<u64>,
+    pub failure_message: Option<String>,
+}
+
+#[derive(Serialize, Deserialize, Clone, ToSchema, Debug, FromPrimitive)]
+#[serde(rename_all = "camelCase")]
+pub enum UdfLanguage {
+    Rust = 0,
```
mwylde, line 46: Do these need to be numeric? I believe that serde will support specifying enums via names: https://serde.rs/json.html.

In protobuf these are numeric because that's how protobuf enums work, but we don't need to do that for JSON.

arroyo-types/src/api.rs
```diff
@@ -0,0 +1,74 @@
...
+#[derive(Serialize, Deserialize, Clone, ToSchema, Debug)]
+#[serde(rename_all = "camelCase")]
+pub struct Udf {
+    pub language: UdfLanguage,
+    pub definition: String,
+}
+
+#[derive(Serialize, Deserialize, Clone, ToSchema, Debug)]
+#[serde(rename_all = "camelCase")]
+pub struct PipelineCollection {
```
mwylde, line 58: Is it possible to make this generic? We will have a lot of collections and we'd want them to all have the same pagination structure.

arroyo-types/src/api.rs
```diff
@@ -0,0 +1,74 @@
...
+    pub definition: Option<String>,
+    pub udfs: Vec<Udf>,
+}
+
+#[derive(Serialize, Deserialize, Clone, ToSchema, Debug)]
+#[serde(rename_all = "camelCase")]
+pub struct JobPost {
+    pub pipeline_id: String,
+    pub running_desired: Option<bool>,
+    pub tasks: Option<u64>,
```
mwylde, line 26: I think this is just returned by the API (it's the number of tasks in the pipeline graph), not a thing the user sets.

arroyo-types/src/api.rs
```diff
@@ -0,0 +1,74 @@
...
+    pub language: UdfLanguage,
+    pub definition: String,
+}
+
+#[derive(Serialize, Deserialize, Clone, ToSchema, Debug)]
+#[serde(rename_all = "camelCase")]
+pub struct PipelineCollection {
+    pub total: i32,
+    pub page: i32,
+    pub per_page: i32,
```
mwylde, line 61: Is per_page needed, or can the user infer that from the number of items?

arroyo-types/src/api.rs
```diff
@@ -0,0 +1,74 @@
...
+pub struct Udf {
+    pub language: UdfLanguage,
+    pub definition: String,
+}
+
+#[derive(Serialize, Deserialize, Clone, ToSchema, Debug)]
+#[serde(rename_all = "camelCase")]
+pub struct PipelineCollection {
+    pub total: i32,
+    pub page: i32,
```
mwylde, line 60: What do you think about using cursors instead of pages (e.g., in https://stripe.com/docs/api/pagination)?

Page-based pagination has the issue that the data may change in between requests. So I request `/v1/jobs` and get back:


```
{
  "page": 0,
  "pages": 2,
  "items": [
    1, 2, 3, 4, 5
  ]
}
```

then I want the next page, so I request `/v1/jobs?page=1`. But in the meantime, record 2 has been deleted so I end up with 

```
{
  "page": 0,
  "pages": 2,
  "items": [
    7, 8, 9, 10, 11
  ]
}
```

(never seeing 6).

Cursor-based APIs address this by giving you a consistent position the database to query from so that you can miss records due to additions/deletions. I think it's also a bit more user friendly, since it means the user doesn't have to calculate the next page, they can just request the cursor.

jbeisen: Since we're using these endpoints for a UI we probably want to have a pagination component that lets users jump to a specific page, and unfortunately that's not possible with cursor-based pagination. Practically I don't think that that race condition is a big issue because the UI will know to refresh the data when certain api calls are made (e.g. if a deletion occurs then SWR will automatically and immediately refresh the collection).

mwylde: That's true, though I'm not sure how useful jumping to a particular page is in this context. The APIs will also be used by developers directly to build their own integrations and issues like this can make it hard to build those things reliably.

arroyo-api/src/rest.rs
```diff
@@ -1,32 +1,144 @@
...
-        (status = 200, description = "Ping endpoint",),
+        (status = 200, description = "Ping endpoint"),
     ),
 )]
 pub async fn ping() -> impl IntoResponse {
     Json("Pong")
 }
 
-pub fn create_rest_app() -> Router {
+#[utoipa::path(
```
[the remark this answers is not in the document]
jbeisen: Sure, done.

arroyo-types/src/api.rs
jbeisen: I've actually removed this for the first release of the api. Let's just only accept Rust udfs for now.

arroyo-types/src/api.rs
jbeisen: Unfortunately it can't be a rust generic because utoipa isn't able to interpret generics for generating the openapi spec. But I did write a declarative macro that generates the types.

arroyo-types/src/api.rs
jbeisen: I changed this to cursor-based

arroyo-types/src/api.rs
jbeisen: Ok I've changed it to cursor based.

arroyo-api/queries/api_queries.sql
```diff
@@ -84,11 +84,21 @@ WHERE organization_id = :organization_id AND id = :id;
 
 ----------- pipelines -------------------
 
+--: DbPipelineRest (textual_repr?, udfs?)
+
+--! get_pipelines : DbPipelineRest
+SELECT pipelines.pub_id as id, name, type, textual_repr, udfs, program FROM pipelines
+WHERE pipelines.organization_id = :organization_id AND pipelines.pub_id IS NOT NULL;
```
jbeisen, line 91: @mwylde since the REST API is based on the new public ids, I've excluded older pipelines that don't have one. Is that ok?

mwylde: I think that's fine

arroyo-types/src/api.rs
```diff
@@ -0,0 +1,56 @@
+use serde::{Deserialize, Serialize};
```
mwylde, line 1: Can these be moved out of arroyo-types? Arroyo-types is the base crate for everything in our system, so only things that are needed by basically every part of the system (including UDFs compiled to WASM) should be here. It looks like this is just use by the api, so maybe can be moved there? 

We also try to keep dependencies minimal, so ideally we wouldn't add utoipa here.

arroyo-types/src/api.rs
```diff
@@ -0,0 +1,56 @@
...
+}
+
+// Collections need to be created with this macro rather than a generic type
+// because utoipa::ToSchema (and the OpenAPI spec) don't support generics natively
+macro_rules! collection_type {
+    ($struct_name:ident, $item_type:ty) => {
+        #[derive(Serialize, Deserialize, Clone, Debug, ToSchema)]
+        #[serde(rename_all = "camelCase")]
+        pub struct $struct_name {
+            pub data: Vec<$item_type>,
```
mwylde, line 49: should this also have the cursor field? we want users to be able to integrate in ways that are future-proof once we turn on pagination.

arroyo-api/src/main.rs
```diff
@@ -237,13 +239,14 @@ async fn server(pool: Pool) {
...
         .unwrap_or_else(|_| format!("http://localhost:{}", ports::CONTROLLER_GRPC));
 
+    let api_server_pool = pool.clone();
     let server = ApiServer {
-        pool,
+        pool: api_server_pool,
         controller_addr,
     };
 
     let rest_addr = format!("0.0.0.0:{}", 8003).parse().unwrap();
```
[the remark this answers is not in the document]
jbeisen: Done. I moved the static assets into rest.rs so we're only using the Router there.

arroyo-api/src/pipelines.rs
```diff
@@ -287,3 +326,198 @@ pub(crate) async fn sql_graph(
...
+                    language: 0,
+                    definition: u.definition.to_string(),
+                })
+                .collect(),
+            preview: false,
+        })),
+    };
+
+    let pipeline_pub_id = generate_id(IdTypes::Pipeline);
+    let transaction = client.transaction().await.map_err(log_and_map)?;
```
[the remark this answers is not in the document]
jbeisen: I removed it.

arroyo-types/src/api.rs
mwylde: Yeah, I don't think we'll need them in the controller. The controller only interacts with the API via the database (i.e., the API writes stuff to the DB, the controller reads it and writes back some other stuff, which the API reads). And if we do, we can use a higher common dependency, like arroyo-rpc.

arroyo-api/Cargo.toml
```diff
@@ -86,4 +86,4 @@ utoipa = "3"
 
 [lib]
 crate-type = ["lib"]
-path = "./src/rest.rs"
+path = "./src/main.rs"
```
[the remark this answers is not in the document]
jbeisen: This is fixed now

arroyo-types/src/api.rs
jbeisen: The way stripe does it is it just accepts a `starting_after` field, so the client essentially gets the cursor from the last result rather than explicitly.

arroyo-openapi/api-spec.json
```diff
@@ -8,6 +8,11 @@
     },
     "version": "1.0.0"
   },
+  "servers": [
```
mwylde, line 11: it looks like this is still out of sync -- I think we discussed gitignoring this?

mwylde: possibly related, even after rebuilding I get this error on the swagger UI:

image

arroyo-api/src/rest.rs
```diff
@@ -1,32 +1,96 @@
...
+        .route("/pipelines/:id/jobs", get(get_jobs));
+
     Router::new()
-        .merge(SwaggerUi::new("/swagger-ui").url("/api-docs/openapi.json", ApiDoc::openapi()))
-        .route("/v1/ping", get(ping))
+        .merge(
+            SwaggerUi::new("/api/v1/swagger-ui")
+                .url("/api/v1/api-docs/openapi.json", ApiDoc::openapi()),
+        )
+        .nest("/api/v1", api_routes)
```
mwylde, line 88: As discussed, `/api` 404s should fallback to a JSON error message instead of the HTML

mwylde (commented):
Overall great, a few more items then we can merge.

One thing that would be good to fix if it's not too hard are the error messages for invalid bodys. Currently, you get a plaintext string like:

> Failed to deserialize the JSON body into the target type: missing field `udfs` at line 1 column 97

which should be a JSON response

mwylde (approved):

arroyo-openapi/api-spec.json
jbeisen: Fixed

arroyo-api/src/rest.rs
jbeisen: Done.

arroyo-api/Cargo.toml
```diff
@@ -2,6 +2,7 @@
 name = "arroyo-api"
 version = "0.3.0"
 edition = "2021"
+license = "MIT"
```
jbeisen, line 5: @mwylde utoipa reads this to include it in the openapi spec. I see there's both Apache and MIT licenses in the repo. Do you know which one applies to use of the API?

mwylde: Arroyo is dual-licensed under Apache and MIT, so the API is both.

mwylde (approved):

PR #193, opened 2023-07-07
Fix docker builds by including arroyo-openapi

mwylde:


jacksonrnewhouse (approved):
Any way to have a quick test for this?

mwylde:
we could do something a bit hacky (parse the Cargo.toml, grep the dockerfiles for each of the crates)

mwylde:
Added a check for this in CI

PR #194, opened 2023-07-07
initial parquet sink

jacksonrnewhouse:
This is mostly working, but still a lot of polishing left to do. At this point an overview of the approach would probably be the most helpful thing.


jacksonrnewhouse:
#197 includes this

PR #197, opened 2023-07-11
Commits and parquet file sink

jacksonrnewhouse:
This adds a parquet sink and introduces a committing phase during checkpointing. This allows for the support of exactly once sinks, the first of which is implemented to support writing Parquet to S3. Parquet on S3 allows for the production of typed parquet files to an S3 directory. The files are created using S3's multipart upload API, and we support the creation of parquet files across multiple checkpoints.

When restoring from checkpoints the subtask with id 0 will be responsible for finishing any files that were previously in-flight, as well as committing any files that were ready to be committed.

Right now we assume idempotency for the completion of multipart uploads, which seems to work. According to the S3 _Glacier_ documentation, ["Complete Multipart Upload is an idempotent operation. After your first successful complete multipart upload, if you call the operation again within a short period, the operation will succeed and return the same archive ID."](https://awscli.amazonaws.com/v2/documentation/api/latest/reference/glacier/complete-multipart-upload.html#:~:text=Complete%20Multipart%20Upload%20is%20an,return%20the%20same%20archive%20ID.) However, the normal S3 docs say that trying to complete an already finished upload might return a 404: "[Description: The specified multipart upload does not exist. The upload ID might be invalid, or the multipart upload might have been aborted or completed](https://docs.aws.amazon.com/AmazonS3/latest/API/API_CompleteMultipartUpload.html)."

I've manually tested the behavior under a variety of cases, including stopping it through the UI, uncontrolled stopping, and failing to log commits. In all cases it was able to resume writes without losing or duplicating data.


arroyo-api/migrations/V12__add_committing.sql
```diff
@@ -0,0 +1 @@
+ALTER TYPE checkpoint_state ADD VALUE 'committing'
```
mwylde, line 1: does this need a semicolon?

arroyo-worker/src/connectors/two_phase_committer.rs
```diff
@@ -0,0 +1,202 @@
...
+    grpc::{TableDeleteBehavior, TableDescriptor, TableType, TableWriteBehavior},
+    CheckpointEvent, ControlMessage,
+};
+use arroyo_state::tables::GlobalKeyedState;
+use async_trait::async_trait;
+use serde::{Deserialize, Serialize};
+use tracing::{info, warn};
+use typify::import_types;
+
+import_types!(schema = "../connector-schemas/parquet/table.json");
```
mwylde, line 15: why is this importing the parquet table config?

arroyo-worker/src/connectors/two_phase_committer.rs
```diff
@@ -0,0 +1,202 @@
...
+
+use crate::engine::Context;
+use arroyo_types::*;
```
mwylde, line 18: generally would not suggest using `*` inputs from a whole crate, as this can cause compilation errors (and potentially changes to behavior) due to new exports in that crate

arroyo-worker/src/connectors/two_phase_committer.rs
```diff
@@ -0,0 +1,202 @@
...
+    type PreCommit: Data;
+
+    fn name(&self) -> String;
+    async fn init(
+        &mut self,
+        task_info: &TaskInfo,
+        data_recovery: Vec<Self::DataRecovery>,
+    ) -> Result<()>;
+    async fn insert_record(&mut self, record: &Record<K, T>) -> Result<()>;
+    // TODO: figure out how to have the relevant vectors be of pointers across async boundaries.
```
mwylde, line 55: they can be pointers across async bounderies so long as they are `&mut` (because `&T` is send only if `T` is sync, but `&mut T` is send if `T` is send).

arroyo-worker/src/connectors/two_phase_committer.rs
```diff
@@ -0,0 +1,202 @@
...
+    async fn process_element(&mut self, record: &Record<K, T>, _ctx: &mut Context<(), ()>) {
+        self.committer
+            .insert_record(record)
+            .await
+            .expect("record inserted");
+    }
+
+    async fn on_close(&mut self, ctx: &mut crate::engine::Context<(), ()>) {
+        info!("waiting for commit message");
+        if let Some(ControlMessage::Commit { epoch }) = ctx.control_rx.recv().await {
```
mwylde, line 137: This logic assumes that the next control message on the queue is a commit message -- is that guaranteed?

arroyo-worker/src/connectors/two_phase_committer.rs
```diff
@@ -0,0 +1,202 @@
...
+            warn!("no commit message received, not committing")
+        }
+    }
+
+    async fn handle_checkpoint(
+        &mut self,
+        checkpoint_barrier: &arroyo_types::CheckpointBarrier,
+        ctx: &mut crate::engine::Context<(), ()>,
+    ) {
+        info!("received checkpoint barrier {:?}", checkpoint_barrier);
```
mwylde, line 149: should this be an info log? I think the underlying infrastructure already logs this.

arroyo-connectors/src/kafka.rs
```diff
@@ -74,7 +74,7 @@ impl Connector for KafkaConnector {
             ),
             TableType::Sink { .. } => (
                 ConnectionType::Sink,
-                "connectors::kafka::sink::KafkaSinkFunc",
+                "connectors::kafka::sink::KafkaSinkFunc::<#in_k, #in_t>",
```
mwylde, line 77: I think you'll need to update all of the other sinks as well

arroyo-controller/src/job_controller/checkpointer.rs
```diff
@@ -161,9 +219,16 @@ impl CheckpointState {
...
                         api::TaskCheckpointEventType::CheckpointSyncFinished
                     }
+                    grpc::TaskCheckpointEventType::FinishedCommit => {
+                        api::TaskCheckpointEventType::CheckpointPreCommit
+                    }
                 } as i32,
             });
 
+        if grpc::TaskCheckpointEventType::FinishedCommit == c.event_type() {
+            panic!("shouldn't receive finished commits while checkpointing");
```
mwylde, line 229: panics in the job controller should only be used for cases that indicate an unrecoverable programming error

arroyo-worker/src/connectors/two_phase_committer.rs
jacksonrnewhouse: There's a number of logging statements that I intend to remove before merging but were helpful for debugging.

arroyo-console/src/routes/pipelines/CreatePipeline.tsx
```diff
@@ -99,7 +99,7 @@ export function CreatePipeline({ client }: { client: ApiClient }) {
       setUdfsInput(def.udfs[0].definition || '');
       setOptions({
         ...options,
-        name: def.name + '-copy',
+        name: def.name,
```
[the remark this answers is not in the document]
jacksonrnewhouse: Yeah, in development the repeated -copy-copy-copy becomes a pain. Since the name isn't necessarily unique I figured it was fine.

arroyo-controller/src/compiler.rs
```diff
@@ -147,6 +147,13 @@ members = [
 exclude = [
   "wasm-fns",
 ]
+[patch.crates-io]
+parquet = {git = 'https://github.com/ArroyoSystems/arrow-rs', branch = '39_0_0/write_trailing_bytes'}
```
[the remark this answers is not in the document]
jacksonrnewhouse: Yeah, the three places are 
1. The top-level Cargo.toml.
2. The computed Cargo.toml in compiler.rs when we build out the entire crate.
3. The fixed Cargo.toml in build_dir for when we're using the compiler service.

arroyo-controller/src/job_controller/checkpointer.rs
jacksonrnewhouse: I'll have it bail, as this likely means that it won't be able to checkpoint.

arroyo-controller/src/schedulers/mod.rs
```diff
@@ -182,6 +182,7 @@ impl Scheduler for ProcessScheduler {
                     .env(JOB_ID_ENV, &job_id)
                     .env(NODE_ID_ENV, format!("{}", 1))
                     .env(RUN_ID_ENV, format!("{}", start_pipeline_req.run_id))
+                    .env("RUST_BACKTRACE", "1")
```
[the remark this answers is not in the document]
jacksonrnewhouse: No

mwylde:
Got a compilation error with this query:

```sql
CREATE TABLE mastodon (
    value TEXT
) WITH (
    connector = 'sse',
    format = 'raw_string',
    endpoint = 'https://mstdn.social/api/v1/streaming/public',
    events = 'update'
);


CREATE TABLE filesystem (
    tag TEXT,
    window TIMESTAMP,
    count INT,
    rownum INT
) WITH (
    connector = 'filesystem',
    fileformat = 'json',
    s3_bucket = 'arroyo-testing',
    s3_directory = 'outputs',
    aws_region = 'us-east-1'
);

CREATE VIEW tags AS (
    SELECT tag FROM (
        SELECT extract_json_string(value, '$.tags[*].name') AS tag
     FROM mastodon)
    WHERE tag is not null
);

INSERT INTO filesystem
SELECT tag,window.start_time,count,row_num FROM (
    SELECT *, ROW_NUMBER() OVER (
        PARTITION BY window
        ORDER BY count DESC) as row_num
    FROM (SELECT count(*) as count,
        tag,
        hop(interval '5 seconds', interval '15 minutes') as window
            FROM tags
            group by tag, window)) WHERE row_num <= 5;
```

```2023-07-19T23:22:11.154930Z ERROR arroyo_server_common: panicked at 'connectors::filesystem::JsonFileSystemSink::<#in_k, #in_t>: Error("expected one of: `for`, parentheses, `fn`, `unsafe`, `extern`, identifier, `::`, `<`, `dyn`, square brackets, `*`, `&`, `!`, `impl`, `_`, lifetime")', arroyo-controller/src/compiler.rs:44:19 panic.file="arroyo-controller/src/compiler.rs" panic.line=44 panic.column=19```

File: arroyo-connectors/resources/parquet.svg
Comment (mwylde) [inline arroyo-connectors/resources/parquet.svg:1]: Icons should be white and roughly square -- also should be a FS icon and not parquet

File: arroyo-connectors/src/filesystem.rs [hunk 12]
```diff
@@ -0,0 +1,191 @@
...
+        })
+    }
+
+    fn from_options(
+        &self,
+        name: &str,
+        opts: &mut std::collections::HashMap<String, String>,
+        schema: Option<&ConnectionSchema>,
+    ) -> anyhow::Result<crate::Connection> {
+        let write_target = if let (Some(s3_bucket), Some(s3_directory), Some(aws_region)) = (
```
mwylde, line 113: The error handling here isn't very user friendly, since if the user misses one of these options or mistypes them they get an unhelpful "Only s3 target are supported". See how error handling is done in the other connectors.

arroyo-connectors/src/filesystem.rs
```diff
@@ -0,0 +1,191 @@
...
+                s3_directory,
+                aws_region,
+            }
+        } else {
+            bail!("Only s3 targets are supported currently")
+        };
+
+        let inactivity_rollover_seconds = opts
+            .remove("inactivity_rollover_seconds")
+            .map(|value| value.parse::<i64>().map_err(|e| anyhow::anyhow!(e)))
```
mwylde, line 129: The error handling here isn't great, since it includes a weird rust error but doesn't tell the user which field has a problem

arroyo-connectors/src/filesystem.rs
```diff
@@ -0,0 +1,191 @@
...
+            .transpose()?;
+
+        let file_settings = Some(FileSettings {
+            inactivity_rollover_seconds,
+            max_parts,
+            rollover_seconds,
+            target_file_size,
+            target_part_size,
+        });
+        let format_settings = match pull_opt("fileformat", opts)?.as_str() {
```
mwylde, line 155: I think it's confusing that all other connectors are configured using `format`, but here it's `fileformat`

arroyo-controller/src/compiler.rs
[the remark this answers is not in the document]
mwylde: There's also the ones in various Docker containers (single and compiler).

arroyo-connectors/resources/parquet.svg
jacksonrnewhouse: Removed it as we aren't putting it in the UI yet.

arroyo-connectors/src/filesystem.rs
jacksonrnewhouse: Improved it, I think.

arroyo-worker/src/connectors/two_phase_committer.rs
jacksonrnewhouse: It is except for if the source data ends, which right now is only for test sources.

arroyo-controller/src/compiler.rs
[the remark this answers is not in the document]
jacksonrnewhouse: I think single and compiler share docker/build_base.

mwylde (approved):

PR #206, opened 2023-07-15
Adding global error handling

edmondop:
Since many components already use the SWR hook, we can handle in the outer layer of the API errors that are raised by nested components. The Home is a special use case since it uses fetch directly, without using SWR, and so needed to be changed. Now any page, if the backend is down, shows an error like so:
Screenshot 2023-07-15 at 4 03 55 PM


arroyo-console/src/routes/home/Home.tsx
```diff
@@ -48,27 +48,24 @@ export const Stat = (props: Props) => {
...
-    };
+const fetchJobs = async (client: ApiClient) => {
+  const clientObj = await client();
+  return (await clientObj.getJobs(new GetJobsReq({}))).jobs;
+};
 
-    fetchData();
-  }, []);
+export function Home({ client }: { client: ApiClient }) {
+  const { data: jobs } = useSWR('jobs', () => fetchJobs(client));
```
jbeisen, line 57: I suggest we keep all `useSWR` calls in `data_fetching.ts`. Doing so keeps all the api client interaction in one place and makes it easier to trigger refreshes of dependent api calls.

jbeisen (changes requested):

arroyo-console/src/routes/home/Home.tsx
edmondop: Fixed with https://github.com/ArroyoSystems/arroyo/pull/206/commits/1f76c8bd2bfb017c97c38ae2d182964a169776e8

jbeisen (approved):
lgtm

mwylde:
Thanks @edmondop!

PR #207, opened 2023-07-16
Validating table name

edmondop:


arroyo-console/src/routes/connections/ConnectionTester.tsx
```diff
@@ -53,6 +53,12 @@ export function ConnectionTester({
   let config = JSON.parse(JSON.stringify(state.table));
   delete config.__meta;
 
+  const isValidSQLTableName = (name: string | undefined) => {
+    // This is a very basic check and may not cover all cases.
+    // Update this according to your actual validation rules.
+    return name && /^[a-zA-Z][a-zA-Z0-9_]*$/.test(name);
```
mwylde, line 59: This regex looks good, except that tables can also begin with an underscore: https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS

mwylde (commented):
One comment but looks great!

arroyo-console/src/routes/connections/ConnectionTester.tsx
edmondop: Fixed with https://github.com/ArroyoSystems/arroyo/pull/207/commits/3d32775dc0defd8a0c672afad5a1a1192748b0ea

mwylde (approved):

PR #218, opened 2023-07-26
implement exactly-once commits to Kafka sinks and read_committed reads to Kafka sources.

jacksonrnewhouse:
Using the transactional kafka APIs we now support putting writing to Kafka via transactions. For each subtask a new kafka producer is created on each checkpoint, which writes all of the data within the checkpoint for that subtask in one transaction. When a checkpoint barrier is received it will switch initialize a new producer, while keeping the old one around for a commit message.

Currently we don't support recovering if the pipeline is interrupted after the checkpoints have been created but before the checkpoint is committed. The semantics of this are pretty tricky and there are a number of approaches with various safety and reliability tradeoffs. Flink, for instance, interrogates the internal transaction state from __transaction_state and then uses reflection to rebuild internal state. I don't think this is an option given the libraries we're using. Another approach that should work is to reliably determine if the previous transaction was successfully committed and, if not, replay the data by reading the uncommitted records off of kafka into a new commit. This will require careful management of active transactions, as Kafka doesn't directly expose the commit state of prior transaction.

I've also added the ability to only have our Kafka sources read committed data, via `read_mode` values of `read_uncommitted` and `read_committed`.

I'd appreciate feedback on the naming of the parameters? For things like kafka configs, should we be using the same names as the underlying service, maybe prefixed by the service? In our case, this'd mean that instead of `read_mode` we use `kafka.isolation.level`.

### DIFF
--- a/arroyo-connectors/src/kafka.rs
+++ b/arroyo-connectors/src/kafka.rs
@@ -157,9 +157,23 @@ impl Connector for KafkaConnector {
                         None | Some("latest") => SourceOffset::Latest,
                         Some(other) => bail!("invalid value for source.offset '{}'", other),
                     },
+                    read_mode: match opts.remove("read_mode").as_ref().map(|f| f.as_str()) {
+                        Some("read_committed") => SourceReadMode::ReadCommitted,
+                        Some("read_uncommitted") | None => SourceReadMode::ReadUncommitted,
+                        Some(other) => bail!("invalid value for source.read_mode '{}'", other),
+                    },
                 }
             }
-            "sink" => TableType::Sink {},
+            "sink" => {
+                let commit_mode = opts.remove("commit_mode");
+                TableType::Sink {
+                    commit_mode: match commit_mode.as_ref().map(|f| f.as_str()) {
+                        Some("at_least_once") | None => SinkCommitMode::AtLeastOnce,
+                        Some("exactly_once") => SinkCommitMode::ExactlyOnce,
+                        Some(other) => bail!("invalid value for commit_mode '{}'", other),
+                    },
+                }
+            }
             _ => {
                 bail!("type must be one of 'source' or 'sink")
             }
--- a/arroyo-connectors/src/kafka.rs
+++ b/arroyo-connectors/src/kafka.rs
@@ -157,19 +157,19 @@ impl Connector for KafkaConnector {
                         None | Some("latest") => SourceOffset::Latest,
                         Some(other) => bail!("invalid value for source.offset '{}'", other),
                     },
-                    read_mode: match opts.remove("read_mode").as_ref().map(|f| f.as_str()) {
-                        Some("read_committed") => SourceReadMode::ReadCommitted,
-                        Some("read_uncommitted") | None => SourceReadMode::ReadUncommitted,
+                    read_mode: match opts.remove("source.read_mode").as_ref().map(|f| f.as_str()) {
+                        Some("read_committed") => Some(SourceReadMode::ReadCommitted),
+                        Some("read_uncommitted") | None => Some(SourceReadMode::ReadUncommitted),
                         Some(other) => bail!("invalid value for source.read_mode '{}'", other),
                     },
                 }
             }
             "sink" => {
-                let commit_mode = opts.remove("commit_mode");
+                let commit_mode = opts.remove("sink.commit_mode");
                 TableType::Sink {
                     commit_mode: match commit_mode.as_ref().map(|f| f.as_str()) {
-                        Some("at_least_once") | None => SinkCommitMode::AtLeastOnce,
-                        Some("exactly_once") => SinkCommitMode::ExactlyOnce,
+                        Some("at_least_once") | None => Some(SinkCommitMode::AtLeastOnce),
+                        Some("exactly_once") => Some(SinkCommitMode::ExactlyOnce),
                         Some(other) => bail!("invalid value for commit_mode '{}'", other),
                     },
                 }
--- a/arroyo-worker/src/connectors/kafka/sink/mod.rs
+++ b/arroyo-worker/src/connectors/kafka/sink/mod.rs
@@ -8,7 +8,7 @@ use arroyo_types::*;
 use std::collections::HashMap;
 use std::marker::PhantomData;
 
-use tracing::warn;
+use tracing::{error, warn};
 
 use rdkafka::producer::{DeliveryFuture, FutureProducer, FutureRecord, Producer};
 use rdkafka::util::Timeout;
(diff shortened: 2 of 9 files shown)


connector-schemas/kafka/table.json
```diff
@@ -22,18 +22,38 @@
...
+                            "enum": [
+                                "read_committed",
+                                "read_uncommitted"
+                            ]
                         }
                     },
                     "required": [
-                        "offset"
+                        "offset",
+                        "read_mode"
```
mwylde, line 37: I believe this is a breaking change that will make existing kafka sources unusable. Can we make it not required and use read_uncommitted as the default value when we use it?

If that's not possible, perhaps it would make sense to include a DB migration that adds this field to existing kafka connections.

connector-schemas/kafka/table.json
```diff
@@ -22,18 +22,38 @@
...
                     ],
                     "additionalProperties": false
                 },
                 {
                     "type": "object",
                     "title": "Sink",
                     "properties": {
+                        "commit_mode": {
+                            "type": "string",
+                            "description": "Type of committing behavior for Kafka Sink",
```
mwylde, line 47: I think this description could be expanded a bit to say that this enables transactional semantics and requires that transactional support be enabled in Kafka

arroyo-connectors/src/kafka.rs
```diff
@@ -157,9 +157,23 @@ impl Connector for KafkaConnector {
                         None | Some("latest") => SourceOffset::Latest,
                         Some(other) => bail!("invalid value for source.offset '{}'", other),
                     },
+                    read_mode: match opts.remove("read_mode").as_ref().map(|f| f.as_str()) {
```
mwylde, line 160: I think this should be `source.read_mode` and `sink.commit_mode` to match the json and the existing `source.offset` config.

As far as the actual names I'm a bit torn—I do think "read_mode" is a better name than "isolation.level" but there's also benefit in using the existing terminology

connector-schemas/kafka/table.json
```diff
@@ -22,18 +22,38 @@
...
+                            "enum": [
+                                "at_least_once",
+                                "exactly_once"
+                            ]
+                        }
                     },
+                    "required": [
+                        "commit_mode"
```
mwylde, line 55: same here — is it possible to make it not required and use a default?

arroyo-worker/src/connectors/kafka/sink/mod.rs
```diff
@@ -76,21 +102,80 @@ impl<K: Key + Serialize, T: Data + Serialize> KafkaSinkFunc<K, T> {
         format!("kafka-producer-{}", self.topic)
     }
 
-    async fn on_start(&mut self, _ctx: &mut Context<(), ()>) {
-        info!("Creating kafka producer for {}", self.bootstrap_servers);
-        let mut client_config = ClientConfig::new();
+    async fn on_start(&mut self, ctx: &mut Context<(), ()>) {
+        self.init_producer(&ctx.task_info)
```
mwylde, line 106: I believe you mentioned that transactional support needs to be enabled in the broker for this to work. If it's not, how does the error manifest? Ideally we'd get that back to the user using the error reporting system: https://github.com/ArroyoSystems/arroyo/blob/master/arroyo-worker/src/engine.rs#L471

arroyo-worker/src/connectors/kafka/sink/mod.rs
```diff
@@ -76,21 +102,80 @@ impl<K: Key + Serialize, T: Data + Serialize> KafkaSinkFunc<K, T> {
...
+        match &mut self.consistency_mode {
+            ConsistencyMode::AtLeastOnce => {
+                self.producer = Some(client_config.create()?);
+            }
+            ConsistencyMode::ExactlyOnce {
+                next_transaction_index,
+                ..
+            } => {
+                client_config.set("enable.idempotence", "true");
+                let transactional_id = format!(
```
mwylde, line 145: should this include the topic as well?

arroyo-worker/src/connectors/kafka/sink/mod.rs
```diff
@@ -150,4 +235,40 @@ impl<K: Key + Serialize, T: Data + Serialize> KafkaSinkFunc<K, T> {
...
+        let ConsistencyMode::ExactlyOnce { next_transaction_index: _, producer_to_complete } = &mut self.consistency_mode else {
+            warn!("received commit but consistency mode is not exactly once");
+            return
+        };
+
+        let Some(committing_producer)= producer_to_complete.take() else {
+            unimplemented!("received a commit message without a producer ready to commit. Restoring from commit phase not yet implemented");
+        };
+        committing_producer
+            .commit_transaction(Timeout::Never)
```
mwylde, line 249: What's the implication of settings this to Timeout::Never?

arroyo-worker/src/connectors/kafka/sink/mod.rs
```diff
@@ -150,4 +235,40 @@ impl<K: Key + Serialize, T: Data + Serialize> KafkaSinkFunc<K, T> {
...
+            .expect("committing producer should succeed");
```
mwylde, line 250: does it make sense to have any retry logic here, given that a single failure is unrecoverable?

arroyo-connectors/src/kafka.rs
jacksonrnewhouse: Going with source.read_mode and sink.commit_mode for now.

arroyo-worker/src/connectors/kafka/sink/mod.rs
jacksonrnewhouse: No, that conversation we had was just that if you're running kafka locally you need some additional settings. In docker-compose.yaml it looks like

```
      KAFKA_OFFSETS_TOPIC_REPLICATION_FACTOR: 1
      KAFKA_TRANSACTION_STATE_LOG_REPLICATION_FACTOR: 1
      KAFKA_MIN_INSYNC_REPLICAS: 1
      KAFKA_TRANSACTION_STATE_LOG_MIN_ISR: 1
```

This converts to settings in /etc/kafka/kafka.properties like
```
offsets.topic.replication.factor=1
transaction.state.log.replication.factor=1
min.insync.replicas=1
transaction.state.log.min.isr=1
```

Transactions have been in kafka since 0.11 from 2017, so I don't think it is worth having a lot of checks for that.

arroyo-worker/src/connectors/kafka/sink/mod.rs
jacksonrnewhouse: Sure, seems reasonable.

arroyo-worker/src/connectors/kafka/sink/mod.rs
jacksonrnewhouse: This means it will block without a timeout. Happy to switch to a different timeout, or have it be configurable.

arroyo-worker/src/connectors/kafka/sink/mod.rs
jacksonrnewhouse: I'm not sure I can think of something we could do to recover from an error here, without a bunch of the other retry management work we've discussed.

arroyo-worker/src/connectors/kafka/sink/mod.rs
```diff
@@ -150,4 +236,49 @@ impl<K: Key + Serialize, T: Data + Serialize> KafkaSinkFunc<K, T> {
...
+            if committing_producer
+                .commit_transaction(Timeout::After(Duration::from_secs(10)))
+                .is_ok()
+            {
+                break;
+            } else {
+                error!("failed to commit {} times, retrying", commits_attempted);
+                commits_attempted += 1;
+            }
+        }
```
mwylde, line 260: Should this panic if we fail all 5 attempts?

mwylde (approved):

arroyo-worker/src/connectors/kafka/sink/mod.rs
jacksonrnewhouse: doh

PR #221, opened 2023-07-27
Use REST API for majority of console

jbeisen:
Update the console to use the REST API for most data fetching, excluding
connection tables, metrics, checkpoint details, and subscribing to job
output. Mildly redesigns the pipelines page to make the distinction
between pipelines and jobs more clear.

Note that this also changes the remaining GRPC endpoints to operate on
the public ids that the REST API exposes. The GRPC endpoints internally
look up the internal id and then operate on that.

Also remove global error handling and explicitly handle the API being
unavailable by polling the /v1/ping endpoint. The page displays a
message if this request fails.

---

New pipelines page:

image

API unavailable:

image


arroyo-api/src/rest.rs
```diff
@@ -77,17 +79,25 @@ pub fn create_rest_app(server: ApiServer, pool: Pool) -> Router {
...
         .allow_origin(cors::Any);
 
+    let jobs_routes = Router::new()
+        .route("/", get(get_pipeline_jobs))
+        .route("/:job_id/errors", get(get_job_errors))
+        .route("/:job_id/checkpoints", get(get_job_checkpoints));
+
     let api_routes = Router::new()
         .route("/ping", get(ping))
+        .route("/validate", post(validate_pipeline))
```
mwylde, line 93: what's your though on having this top-level vs. nested under pipelines?

arroyo-api/src/rest_types.rs
```diff
@@ -31,6 +41,68 @@ pub struct Pipeline {
     pub checkpoint_interval_micros: u64,
     pub stop: StopType,
     pub created_at: u64,
+    pub action: Option<StopType>,
```
mwylde, line 44: The `action` fields are pretty specific to powering the UI of the console and exist to shift some complex logic from the console to the server. I think it would be good to avoid documenting them as part of the public API as they are basically derivative of other information and could easily change in the future if the UI changes.

I know it's a bit impure but is there a way we could hide these away somewhere?

mwylde (commented):
This looks great!

arroyo-api/src/rest_types.rs
jbeisen: The way we're generating the api spec I don't think allows excluding a field from the spec, which wouldn't really work anyway with openapi-fetch in the frontend. 

Right now the only way I can think of simplifying it is by moving some of the logic to the frontend.

arroyo-api/src/rest.rs
jbeisen: Done

integ/src/main.rs
```diff
@@ -16,6 +20,8 @@ mod embedded {
     embed_migrations!("../arroyo-api/migrations");
 }
 
+include!("../../arroyo-openapi/client/src/lib.rs");
```
jbeisen, line 23: The client is included like this since it's generated by arroyo-openapi's build.rs. I'm not sure if there's a better way.

mwylde (approved):
This looks great! 

I noticed one small regression when stopping a job. The job state (as shown in the console) will move to stopped, but the action button will still look like this:

image

integ/src/main.rs
mwylde: I think you can depend on the arroyo-openapi crate in the integ Cargo.toml instead of including the file here

.github/workflows/ci.yml
```diff
@@ -56,6 +56,10 @@ jobs:
           override: true
       - name: Setup npm
         uses: actions/setup-node@v3
+      - name: Install OpenAPI Generator
+        run: |
+          npm install @openapitools/openapi-generator-cli -g
```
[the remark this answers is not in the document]
jbeisen: Yup - https://github.com/ArroyoSystems/arroyo-docs/pull/22

integ/src/main.rs
jbeisen: I don't think this is possible because the client is not technically a part of the `arroyo-openapi` crate. Since it's generated by `arroyo-openapi/build.rs` it doesn't actually exist until build time.

jbeisen: This is similar to how we have to use the `include!` macro for the generated api sql queries, I think it's just a consequence of the mono repo setup.

jbeisen: Good idea, done.

PR #222, opened 2023-07-28
Add support for idle watermarks

mwylde:
This PR adds support for _idle watermarks_ throughout the system and updates the PeriodicWatermarkGenerator to be able to produce idle watermarks.

Idle watermarks help solve a problem in partitioned watermark systems: progress can be held back across the entire application because one partition is idle. For example, if we're reading from two kafka partitions with a very un-even load distribution, we may end up in a situation where one partition is producing data and advancing its watermark, but the other one is not receiving any events and is not able to advance its watermark.

If we shuffle those two partitions, the one that's not receiving watermarks will prevent us from moving forward and processing data on the other one.

The situation is particularly bad if we have more subtasks than kafka partitions: then the subtasks without partitions will never emit a watermark, and the application will never make progress.

With idle watermarks, a source (or watermark generator) can make its partition as _idle_, which means that it gets ignored for watermark calculations.

As of this PR, this can happen in two cases:
1. When a partitioned source (or source that can only have a single partition, like websocket) has more subtasks than partitions, subtasks without partitions will immediately mark themselves idle
2. By default, the PeriodicWatermarkGenerator created for a source will emit an idle watermark after 5 minutes of idleness, which is configurable by setting `idle_micros` when constructing the source table.


arroyo-macro/src/lib.rs
```diff
@@ -472,6 +490,23 @@ fn impl_stream_node_type(
...
+        let tick_setup = tick_ms.as_ref().map(|t| {
+            quote! {
+                let mut ticks = 0u64;
+                let mut next_tick = std::time::SystemTime::now() + std::time::Duration::from_millis(#t);
+            }
+        });
+
+        let tick_case = tick_ms.as_ref().map(|t| {
+            quote! {
+                _ = tokio::time::sleep(next_tick.duration_since(SystemTime::now()).unwrap_or(Duration::ZERO)) => {
```
jacksonrnewhouse, line 502: What about using `tokio::time::interval`? Something like

```
let mut interval = tokio::time::interval([Duration](std::time::Duration::from_millis(#t)));
interval.set_missed_tick_behavior(MissedTickBehavior::Skip);
```
and then
```
_ = interval.tick() {}
```

in the select block.

arroyo-worker/src/engine.rs
```diff
@@ -130,11 +92,66 @@ pub trait StreamNode: Send {
     ) -> JoinHandle<()>;
 }
 
+pub struct WatermarkHolder {
+    // This is the last watermark with an actual value; this helps us keep track of the watermark we're at even
+    // if we're currently idle
+    last_present_watermark: Option<SystemTime>,
+    cur_watermark: Option<Watermark>,
```
jacksonrnewhouse, line 99: What about having Watermark cover the `None` case?

arroyo-worker/src/operators/aggregating_window.rs
```diff
@@ -215,15 +222,16 @@ impl<K: Key, T: Data, BinA: Data, MemA: Data, OutT: Data>
...
-        while self.should_advance(watermark) {
-            self.advance(ctx).await;
+    async fn handle_watermark(&mut self, watermark: Watermark, ctx: &mut Context<K, OutT>) {
+        match watermark {
+            Watermark::EventTime(t) => {
+                while self.should_advance(t) {
+                    self.advance(ctx).await;
+                }
+            }
+            Watermark::Idle => (),
```
jacksonrnewhouse, line 232: Could do this with an 
```
if let Watermark::EventTime(t) = watermark {
    while self.should_advance(t) {
        self.advance(ctx).await;
    }
}

jacksonrnewhouse (approved):
Looks great!

arroyo-worker/src/operators/aggregating_window.rs
mwylde: It's intentionally a match instead of an if let to make it more future-proof. I was worried that if we at some point in the future added another variant (Watermark::ProcessingTime for example) that would introduce bugs in all of the call sites, so I wanted to ensure that would cause a compile error.

arroyo-worker/src/engine.rs
mwylde: In practice most of the handling is disjoint between the None and Some cases, and it makes using it much more convenient because we can use `?` for None.

arroyo-macro/src/lib.rs
mwylde: nice, didn't know about tokio::interval

PR #228, opened 2023-08-01
Stream job output using server-sent events

jbeisen:
Add an endpoint to the REST API to stream the job output using Server-Sent Events. Consume the output in the console.

### DIFF
--- a/arroyo-api/src/jobs.rs
+++ b/arroyo-api/src/jobs.rs
@@ -1,25 +1,34 @@
 use crate::queries::api_queries::{DbCheckpoint, DbLogMessage, DbPipelineJob};
 use arroyo_datastream::Program;
+use arroyo_rpc::grpc;
 use arroyo_rpc::grpc::api::{
     CheckpointDetailsResp, CheckpointOverview, CreateJobReq, JobDetailsResp, JobStatus,
     PipelineProgram, StopType,
 };
+use arroyo_rpc::grpc::controller_grpc_client::ControllerGrpcClient;
 use arroyo_rpc::public_ids::{generate_id, IdTypes};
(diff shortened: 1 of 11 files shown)


arroyo-api/src/jobs.rs
```diff
@@ -368,6 +377,89 @@ pub async fn get_job_checkpoints(
     }))
 }
 
+/// Subscribe to a job's output
+#[utoipa::path(
+    get,
+    path = "/v1/pipelines/{pipeline_id}/jobs/{job_id}/stream",
```
mwylde, line 383: what do you think about a less generic name than stream? maybe output?

mwylde (commented):
This looks great. When I run it locally I'm not seeing results when I hit "Start Preview" in the pipeline editor. Can you verify that that's working for you?

jbeisen:
@mwylde I just confirmed I'm getting results. Can you check your console to see if there are any errors?

arroyo-api/src/jobs.rs
jbeisen: Sure, done.

mwylde:
I'm getting this error:

```
Uncaught (in promise) ConnectError: [unimplemented] This functionality has been moved to the REST API.
    at extractHeadersError (grpc-web-transport.js:235:12)
    at grpc-web-transport.js:121:266
    at async Object.serverStream (grpc-web-transport.js:104:24)
    at async Object.next (promise-client.js:50:46)
    at async subscribeToOutput (CreatePipeline.tsx:116:24)
```

mwylde (approved):

mwylde:
Cleaned everything and it's now working. Apologies for the noise.

PR #230, opened 2023-08-04
Move rest types to arroyo-rpc

jbeisen:
In order to remove some protobuf types we need their replacements to be usable by the same dependencies. Keeping these types in in arroyo-rpc allows that.

### DIFF
--- a/arroyo-api/src/jobs.rs
+++ b/arroyo-api/src/jobs.rs
@@ -23,12 +23,13 @@ const PREVIEW_TTL: Duration = Duration::from_secs(60);
 
 use crate::pipelines::{query_job_by_pub_id, query_pipeline_by_pub_id};
 use crate::rest::AppState;
-use crate::rest_types::{
-    Checkpoint, CheckpointCollection, JobCollection, JobLogMessage, JobLogMessageCollection,
-    OutputData,
-};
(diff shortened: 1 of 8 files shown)


arroyo-api/src/metrics.rs
```diff
@@ -70,75 +101,42 @@ pub async fn get_operator_metric_groups(
     let end = (to_millis(SystemTime::now()) / 1000) as i64;
     let start = end - 5 * 60;
 
-    impl MetricNames {
```
jbeisen, line 73: With `MetricNames` in arroyo-rpc we can no longer have an implementation here.

arroyo-api/src/jobs.rs
```diff
@@ -269,11 +270,17 @@ pub async fn get_job_errors(
 
 impl Into<JobLogMessage> for DbLogMessage {
     fn into(self) -> JobLogMessage {
+        let level: JobLogLevel = match self.log_level {
+            LogLevel::info => JobLogLevel::Info,
+            LogLevel::warn => JobLogLevel::Warn,
+            LogLevel::error => JobLogLevel::Error,
+        };
```
jbeisen, line 277: I had to do this casting inline since it's a database type, which arroyo-rpc doesn't have access to. Same with `StopMode` below.

mwylde (approved):

PR #232, opened 2023-08-07
Fix action button behavior

jbeisen:
Enable the button even if an action is in progress.


jbeisen:
Yeah, I didn't change that logic in the API, just moved it into a function so it could be used by both the rest api and grpc api. But in https://github.com/ArroyoSystems/arroyo/pull/221/commits/9c5de6eaab8c54705bea3dca622ed01c9bfbf40b I added a condition to the console that disabled the button if an action is in progress, which I shouldn't have done.

mwylde (approved):
Makes sense!

PR #234, opened 2023-08-08
connectors: Add Kinesis Source and Sink

jacksonrnewhouse:
This adds support for Kinesis via a source and sink. Sources and Sinks are configured via the name of the stream.
Source also takes a "source.offset" of either "Earliest" or "Latest", similar to kafka.

#### Execution Flow
The source operates by having an active future for every open shard the operator owns. These futures are all advanced via a FuturesUnordered, and the futures are all just labeled bits of compute, without any data that should be owned by the overall control loop. This lets us take accurate snapshots without waiting for the futures to complete. Unfortunately, the futures produced by calling something like `async fn my_async_function(&self)`  are always owned by `self`. The `BoxedFuture` gets around this, at the cost of some complexity.

In order to find the affiliated shard state, the futures are returned with a `name`, which corresponds to the shard_id.

This approach is similar to what I implemented for the Filesystem Sink. Might be worth figuring out how to better structure it or if there's a replacement we prefer.

##### Semantics
The operator consumes each shard of data in order, similar to the Kafka source. The set of shards that a subtask is responsible for is determined solely by the hash of the shard_id. Flink defaults to this behavior, but also has an option to evenly divide the hash space.

There are no order guarantees for how data is read off of different shards. While we eagerly fetch from all live shards, it is possible that one might fall behind. Since watermarks are at the subtask level, it is possible that a shard that falls behind will have its data dropped. Finally, Kinesis has the notion of "parent" and "child" shards, with the child shard having data after the parents. Ensuring this while letting child shards be on different subtasks is not currently possible, and no effort was made to ensure this.

#### Execution Flow
The Kinesis sink operates in batches. Right now it waits until one of the following conditions: 500 messages, 4.5MB of data, or 1s has passed. The first two are dictated by the constraints on the PutRecords method, while the timeout is there to ensure we don't have messages linger in the sink. 

The PutRecords method allows for partial success of writes, often because of capacity limits on some but not all of the shards in the stream. Once we decide to flush we currently try repeatedly to finish the batch. While this is happening no new messages are consumed, leading to back-pressure.

#### Semantics
Because the batch puts can succeed only some of the records, the output order is not guaranteed. Additionally, because Kinesis requires a key, if there isn't one in the Pipeline DAG we add a UUID4. 

Right now if there's a key over 256 bytes we won't be able to write. I think our SQL solution will, in general, use the UUID4, but when we add partitioning support, we should have a plan for this.


arroyo-worker/src/connectors/kinesis/source/mod.rs
```diff
@@ -0,0 +1,508 @@
...
+                    let shard_id = result.name;
+                    match self.handle_async_result_split(shard_id,
+                        result.result.map_err(|e| UserError::new("Fatal Kinesis error", e.to_string()))?, ctx).await? {
+                        Some(future) => {
+                            futures.push(future);
+                        },
+                        None => {}
+                    }
+                },
+                _ = sleep(Duration::from_secs(60)) => {
```
mwylde, line 385: The logic here is that if we don't do anything else for 60 seconds, we'll call sync_shards. Is it intended that this never runs in steady-state, or is this meant to run every 60 seconds?

(Also generally some more comments would be great to explain how our code is interacting with the semantics of kinesis)

arroyo-worker/src/connectors/kinesis/source/mod.rs
```diff
@@ -0,0 +1,508 @@
...
+                    if self.sync_shards(ctx).await.is_err() {
+                        warn!("failed to sync shards");
```
mwylde, line 387: should this report the error?

arroyo-worker/src/connectors/kinesis/source/mod.rs
```diff
@@ -0,0 +1,508 @@
...
+    }
+
+    async fn process_records(
+        &mut self,
+        get_records_output: GetRecordsOutput,
+        ctx: &mut Context<(), T>,
+    ) -> Result<Option<String>, UserError> {
+        let records = get_records_output.records.unwrap_or_default();
+        for record in records {
+            let data = record.data.unwrap().into_inner();
```
mwylde, line 433: Do you know in what cases `record.data` and `record.approximate_arrival_timestamp` would be None?

arroyo-worker/src/connectors/kinesis/source/mod.rs
```diff
@@ -0,0 +1,508 @@
...
+            let value = formats::deserialize_slice(&self.format, &data)?;
+            let timestamp = record.approximate_arrival_timestamp.unwrap();
+            let output_record = Record {
+                timestamp: from_nanos(timestamp.as_nanos() as u128),
+                key: Some(()),
```
mwylde, line 438: Why `Some(())` and not `None`?

mwylde (commented):
Initial review of just the source code. I'll spent some more time later today trying to fully understand the control flow there, and digging into the sink.

arroyo-worker/src/connectors/kinesis/source/mod.rs
jacksonrnewhouse: No, it's meant to be something that should trigger about every 60 seconds. Should probably be using a tokio Interval.

arroyo-worker/src/connectors/kinesis/source/mod.rs
jacksonrnewhouse: I was going to say they were both required and this is just a deficiency of the Kinesis client, but that's not entirely accurate. Data is required, while ApproximateArrivalTimestamp is not. Looking through the Wayback machine, it looks like the field was added at the end of 2023. It seems safe to assume it is always present. https://aws.amazon.com/about-aws/whats-new/2015/09/amazon-kinesis-server-side-timestamp/

arroyo-worker/src/connectors/kinesis/source/mod.rs
```diff
@@ -0,0 +1,508 @@
...
+        let config = load_from_env().await;
+        self.kinesis_client = Some(KinesisClient::new(&config));
+        let mut futures = FuturesUnordered::new();
+        let get_shard_iterator_call = self.get_shard_iterator_call();
+        for shard in self.shards.values_mut() {
+            if shard.closed {
+                continue;
+            }
+            let get_shard_iterator_call = shard.set_shard_iterator(get_shard_iterator_call.clone());
+            futures.push(shard.update_shard_iterator(get_shard_iterator_call));
```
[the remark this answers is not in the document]
jacksonrnewhouse: Yeah, this whole thing became a mess because I was trying to use a "convenience" method that returned the unwrapped &KinesisClient. This meant I was running into issues anytime I called get_mut() on self.shard, as that is a mutable borrow on self.shard so you can't also use the result of any other Self function. Inlining the shard unwrap fixes all of this, which I just did.

arroyo-worker/src/connectors/kinesis/source/mod.rs
jacksonrnewhouse: fixed.

arroyo-worker/src/connectors/kinesis/source/mod.rs
jacksonrnewhouse: Done

arroyo-worker/src/connectors/kinesis/source/mod.rs
jacksonrnewhouse: fixed.

arroyo-worker/src/connectors/kinesis/source/mod.rs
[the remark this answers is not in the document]
jacksonrnewhouse: Yeah, this whole thing became a mess because I was trying to use a "convenience" method that returned the unwrapped &KinesisClient. This meant I was running into issues anytime I called get_mut() on self.shard, as that is a mutable borrow on self.shard so you can't also use the result of any other Self function. Inlining the shard unwrap fixes all of this, which I just did.

arroyo-worker/src/connectors/kinesis/source/mod.rs
```diff
@@ -0,0 +1,552 @@
...
+    /// records and control messages. There are three prongs to the tokio select loop:
+    /// * A `FuturesUnordered` tha contains futures for reading off of shards.
+    /// * An interval that periodically polls for new shards, initializing their futures.
+    /// * Polling off of the control queue, to perform checkpointing and stop the operator.
+    async fn run_int(&mut self, ctx: &mut Context<(), T>) -> Result<SourceFinishType, UserError> {
+        let config = load_from_env().await;
+        self.kinesis_client = Some(KinesisClient::new(&config));
+        let starting_futures = self.init_shards(ctx).await.map_err(|e| {
+            UserError::new(
+                "failed to initialize shards. Underlying error:",
```
mwylde, line 393: The first parameter of the user error should be an error name; it's not concatenated with the error details. So this could just be "failed to initialize shards"

arroyo-worker/src/connectors/kinesis/source/mod.rs
```diff
@@ -0,0 +1,552 @@
...
+        &mut self,
+        shard_id: String,
+    ) -> Result<Option<BoxedFuture<AsyncNamedResult<AsyncResult>>>, UserError> {
+        let shard_state = self.shards.get_mut(&shard_id).unwrap();
+        Ok(Some(shard_state.get_update_shard_iterator_future(
+            self.kinesis_client.as_ref().unwrap(),
+        )))
+    }
+
+    fn get_shard_iterator_call(&self) -> GetShardIterator {
```
mwylde, line 377: It looks like this is now unused and can be deleted (not sure why it's not generating a warning for unused code...)

arroyo-worker/src/connectors/kinesis/sink/mod.rs
```diff
@@ -0,0 +1,238 @@
...
+            _phantom: PhantomData,
+        }
+    }
+
+    fn name(&self) -> String {
+        format!("kinesis-producer-{}", self.name)
+    }
+
+    async fn on_start(&mut self, _ctx: &mut Context<(), ()>) {
+        let config = load_from_env().await;
```
mwylde, line 54: what do you think about allowing users to configure the kinesis endpoint? This would allow local development (with something like kinesalite) and would allow talking to kinesis in different regions.

arroyo-worker/src/connectors/kinesis/sink/mod.rs
```diff
@@ -0,0 +1,238 @@
...
+        }
+    }
+
+    async fn process_element(&mut self, record: &Record<K, T>, _ctx: &mut Context<(), ()>) {
+        let k = record
+            .key
+            .as_ref()
+            .map(|k| serde_json::to_string(k).unwrap())
+            .unwrap_or_else(|| Uuid::new_v4().to_string());
+        let v = serde_json::to_string(&record.value).unwrap();
```
mwylde, line 74: This can be updated to use the new format stuff introduced in #235 (see the kafka sink for an example)

arroyo-worker/src/connectors/kinesis/sink/mod.rs
```diff
@@ -0,0 +1,238 @@
...
+    }
+
+    async fn on_start(&mut self, _ctx: &mut Context<(), ()>) {
+        let config = load_from_env().await;
+        let client = KinesisClient::new(&config);
+        self.client = Some(client);
+    }
+
+    async fn handle_checkpoint(&mut self, _: &CheckpointBarrier, _: &mut Context<(), ()>) {
+        if let Some(batch_preparer) = self.in_progress_batch.take() {
```
mwylde, line 60: In the future we may want to write the batch to state instead of flushing during checkpointing but this is probably ok for now

arroyo-worker/src/connectors/kinesis/sink/mod.rs
```diff
@@ -0,0 +1,238 @@
...
+            self.flush_with_retries(batch_preparer)
+                .await
+                .expect("failed to flush batch during processing");
+        } else {
+            self.in_progress_batch = Some(batch_preparer);
+        }
+    }
+
+    async fn handle_tick(&mut self, _: u64, _ctx: &mut Context<(), ()>) {
+        if self.in_progress_batch.is_none() {
```
mwylde, line 97: I don't think this is needed, given the let-else that follows it

arroyo-worker/src/connectors/kinesis/sink/mod.rs
```diff
@@ -0,0 +1,238 @@
...
+struct FlushConfig {
+    max_record_count: usize,
+    max_data_size: usize,
+    max_age: Duration,
+}
+
+impl FlushConfig {
+    fn new_from_table(table: &KinesisTable) -> Self {
+        let TableType::Sink { batch_flush_interval_millis, batch_max_buffer_size, records_per_batch } = &table.type_ else {
+            panic!("found non-sink kafka config in sink operator");
```
mwylde, line 168: nit: kafka => kinesis

arroyo-worker/src/connectors/kinesis/sink/mod.rs
jacksonrnewhouse: I think we should handle that when we build out more complete AWS configuration support. Right now we mainly assume the default environment will tell you what you need.

mwylde (approved):

PR #242, opened 2023-08-10
Switch all connection endpoints to REST API

jbeisen:
Deprecate all the GRPC methods for managing connections (now called
'connection profiles') and connection tables and move them to the REST
API. Switch the console to use the new endpoints.

Also fix some styling issues in the console, including limiting the
error message on the create pipeline page to 2 lines.


arroyo-api/src/connection_tables.rs
```diff
@@ -86,45 +98,235 @@ async fn get_and_validate_connector<E: GenericClient>(
...
+        loop {
+            let m = match rx.recv().await {
+                Some(d) => d,
+                None => {
+                    warn!("Stream closed");
+                    break;
+                }
+            };
+
+            let message = match m {
```
mwylde, line 180: It might be a bit cleaner to combine this match block with the previous one

arroyo-api/src/connection_tables.rs
```diff
@@ -86,45 +98,235 @@ async fn get_and_validate_connector<E: GenericClient>(
...
+
+    let (tx2, rx2) = channel(8);
+
+    connector
+        .test(&req.name, &config, &req.config, schema.as_ref(), tx)
+        .map_err(|e| {
+            Status::invalid_argument(format!("Failed to parse config or schema: {:?}", e))
+        })?;
+
+    tokio::spawn(async move {
```
mwylde, line 170: instead of having this intermediate task converting from the gRPC-style messages being produced now into SSE messages, it might make sense to update the interface for the connectors — but also fine to leave this for now

arroyo-api/src/connection_tables.rs
```diff
@@ -86,45 +98,235 @@ async fn get_and_validate_connector<E: GenericClient>(
...
+    Some(ConnectionProfile {
+        id: c.pub_id.clone(),
+        name: c.connection_name.as_ref()?.clone(),
+        connector: c.connection_type.as_ref()?.clone(),
+        description: connector.config_description(&config).ok()?,
+        config,
+    })
+}
+
+pub(crate) async fn get<C: GenericClient>(
```
mwylde, line 217: Do we still need this method? It looks like it's used by pipelines but could that rely on the new get_connection_tables instead?

arroyo-api/src/connection_tables.rs
```diff
@@ -133,81 +335,118 @@ pub(crate) async fn create(
...
-    req: CreateConnectionTableReq,
-    auth: AuthData,
-    client: &impl GenericClient,
-) -> Result<Receiver<Result<TestSourceMessage, Status>>, Status> {
-    let (connector, _, config, schema) = get_and_validate_connector(&req, &auth, client).await?;
+impl TryInto<ConnectionTable> for DbConnectionTable {
+    type Error = ();
+    fn try_into(self) -> Result<ConnectionTable, Self::Error> {
+        let Some(connector) = connector_for_type(&self.connector) else {
+                warn!("invalid connector {} in saved ConnectionTable {}", self.connector, self.id);
```
mwylde, line 355: What do you think about moving all of the warnings into errors that are returned to the caller? That way the caller can handle them appropriately (possibly warning, or failing in some other way)

arroyo-api/src/connection_tables.rs
```diff
@@ -133,81 +335,118 @@ pub(crate) async fn create(
...
-    let (tx, rx) = channel(8);
+        let connection = get_connection_profile(&self, &*connector);
+
+        let table = serde_json::to_string(&self.config).unwrap();
+        let schema: Option<ConnectionSchema> = self.schema.and_then(|s| {
+            serde_json::from_value(s)
+                .map_err(|e| {
+                    warn!("Error during connection schema deserialization: {:?}", e);
+                })
+                .ok()
```
mwylde, line 367: I don't think this is necessarily the correct behavior — if we fail to deserialize the schema that will cause problems later and we should error out here instead of telling our caller that there was no schema

arroyo-api/src/connection_tables.rs
```diff
@@ -133,81 +335,118 @@ pub(crate) async fn create(
...
+                return Err(());
+            }
+        };
 
-    connector
-        .test(&req.name, &config, &req.config, schema.as_ref(), tx)
-        .map_err(|e| {
-            Status::invalid_argument(format!("Failed to parse config or schema: {:?}", e))
-        })?;
+        let schema = match schema {
```
mwylde, line 384: Do you think it makes sense to move this into the previous match block? For example

``` 
let schema = match ... {
  Ok(Some(schema)) => schema,
  Ok(None) => return Err("Schema required but not available");
  Err(e) => return Err(format!("Invalid schema for {}: {:?}",  self.id, e));
}

File: arroyo-api/src/connection_tables.rs [hunk 7]
```diff
@@ -133,81 +335,118 @@ pub(crate) async fn create(
...
             let table = serde_json::to_string(&t.config).unwrap();
-            let schema = t.schema.map(|s| serde_json::from_value(s).unwrap());
+
+            let schema: Option<ConnectionSchema> =
+                t.schema
+                    .and_then(|value| match serde_json::from_value(value) {
+                        Ok(connection_schema) => Some(connection_schema),
+                        Err(err) => {
+                            warn!("Connection schema deserialization failed: {:?}", err);
+                            None
```
mwylde, line 446: Similarly, we should not return partial connection tables — if we fail to deserialize the schema we should warn and then omit the table

arroyo-api/src/connection_tables.rs
```diff
@@ -86,45 +98,235 @@ async fn get_and_validate_connector<E: GenericClient>(
...
+        .filter_map(|t| {
+            // return None;
+            let Some(connector) = connector_for_type(&t.connector) else {
+                warn!("invalid connector {} in saved ConnectionTable {}", t.connector, t.id);
+                return None;
+            };
+
+            let connection_profile = get_connection_profile(&t, &*connector);
+
+            let table = serde_json::to_string(&t.config).unwrap();
```
mwylde, line 238: this probably should return an error instead of unwrapping (and below as well)—otherwise a single bad config in your database prevents you from using the system

mwylde (changes requested):
Not quite done yet but have some initial comments. Overall looks great!

One issue I noticed on the console is that deleting a connection table that still has pipelines correctly produces an error from the API, but the console still shows a success message

arroyo-rpc/src/types.rs
```diff
@@ -296,3 +299,284 @@ pub struct Connector {
...
     pub connection_config: Option<String>,
 }
+
+#[derive(Serialize, Deserialize, Clone, Debug, ToSchema)]
+#[serde(rename_all = "camelCase")]
+pub struct ConnectionProfile {
+    pub id: String,
+    pub name: String,
+    pub connector: String,
+    pub config: String,
```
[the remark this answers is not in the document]
jbeisen: Good idea. Are you suggesting that we get rid of the `Connector` trait's associated types `ConfigT` and `TableT`?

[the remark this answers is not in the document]
mwylde: I think that's still useful because it turns them into concrete types that the connectors can work with. But instead of using strings as the inputs to the ErasedConnector functions, we can use Values (and throughout, avoiding a bunch of serialization/deserialization).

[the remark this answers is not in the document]
mwylde: Although we can rename ConfigT to ProfileT or something

arroyo-rpc/src/types.rs
```diff
@@ -296,3 +299,284 @@ pub struct Connector {
...
+        }
+    }
+}
+
+impl From<String> for ConnectionType {
+    fn from(value: String) -> Self {
+        match value.to_lowercase().as_str() {
+            "source" => ConnectionType::Source,
+            "sink" => ConnectionType::Sink,
+            _ => panic!("invalid table type {}", value.to_lowercase().as_str()),
```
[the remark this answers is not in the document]
jbeisen: I've made it a TryFrom but the `table_type` column of `connection_tables` is a string, not JSON, so I don't think Deserialize helps here. Am I missing something?

arroyo-rpc/src/types.rs
```diff
@@ -296,3 +299,284 @@ pub struct Connector {
...
+    pub field_name: String,
+    pub field_type: SourceFieldType,
+    pub sql_name: Option<String>,
+    pub nullable: bool,
+}
+
+#[derive(Serialize, Deserialize, Clone, Debug, ToSchema)]
+#[serde(rename_all = "snake_case")]
+pub enum SchemaDefinition {
+    Json(String),
```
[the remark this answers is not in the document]
jbeisen: I've renamed these enum variants resemble their protobuf names, but doing so still requires serde aliases to be backwards compatible.

arroyo-api/src/connection_tables.rs
```diff
@@ -41,29 +52,30 @@ async fn get_and_validate_connector<E: GenericClient>(
...
+        &req.connection_profile_id
+    {
+        let connection_profile = api_queries::get_connection_profile_by_pub_id()
+            .bind(c, &auth.organization_id, &connection_profile_id)
             .opt()
             .await
             .map_err(log_and_map)?
             .ok_or_else(|| {
-                Status::failed_precondition(format!("No connection with id '{}'", connection_id))
+                Status::failed_precondition(format!(
```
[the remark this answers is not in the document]
jbeisen: Yup, changed it

arroyo-api/src/connection_tables.rs
jbeisen: That would require making connectors depend on axum so that they can construct `axum::response::sse::Event` objects, right?

arroyo-api/src/connection_tables.rs
jbeisen: I renamed this and made `get_connection_tables` use it.

arroyo-api/src/connection_tables.rs
mwylde: I think that's fine — currently it has a dependency on tonic that won't be needed after these changes

arroyo-rpc/src/types.rs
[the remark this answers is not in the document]
mwylde: No, you're right — I forgot that we store this separately in the DB

arroyo-rpc/src/types.rs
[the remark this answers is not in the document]
mwylde: I think we could also do a migration (with postgres json functions) to rename them in the DB

arroyo-api/src/connection_tables.rs
jbeisen: Sure, done. Currently the only caller is `create_connection_table`, where it makes sense to log and return an `ErrorResp`.

arroyo-rpc/src/types.rs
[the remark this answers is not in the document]
jbeisen: Cool, done. One thing I wasn't sure about was `EmptyConfig`. Since `serde_json::Value::Null` doesn't parse into that, I had to change `EmptyConfig` to be a type alias for `serde_json::Value`, which works, but is a little unclear because it should always be the `Null` variant. Do you have any better ideas?

arroyo-api/src/connection_tables.rs
jbeisen: True, I've made it return an error.

arroyo-api/src/connection_tables.rs
jbeisen: The previous match block is now already matching on an `Option<Result<...>> ` so combining another one into it would be a bit complicated, wouldn't it?

arroyo-api/src/connection_tables.rs
jbeisen: I agree, done.

arroyo-api/src/connection_tables.rs
jbeisen: `t.config` is now passed directly to `get_schema` as a `serde_json::Value` so no need to unwrap.

arroyo-api/src/connection_tables.rs
jbeisen: Cool, done.

arroyo-rpc/src/types.rs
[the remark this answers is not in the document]
mwylde: `EmptyConfig` is actually `{}`, not null so I think it should work correctly as is

arroyo-api/src/connection_tables.rs
mwylde: I do think it's clearer to combine them, but not a big deal

mwylde (approved):

arroyo-rpc/src/types.rs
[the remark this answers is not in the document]
jbeisen: Right, but an empty json is deserialized as `serde_json::Value::Null`, so this causes a panic:

```let _: EmptyConfig = serde_json::from_value(serde_json::Value::Null).unwrap();```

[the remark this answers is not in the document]
jbeisen: ``` Invalid connector config for 110: Error("invalid type: null, expected struct EmptyConfig", line: 0, column: 0)```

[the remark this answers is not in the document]
jbeisen: Never mind, you were right, I've put it back how it was.

mwylde:
🎉

PR #244, opened 2023-08-11
Fix job output rendering

jbeisen:
Modifying the existing array of output doesn't trigger a re-render since the reference to the array in state doesn't change. This creates a new array on each message to force a re-render.


mwylde (approved):
Nice catch!

PR #245, opened 2023-08-11
Report errors instead of panicking when we receive invalid debezium messages

mwylde:


arroyo-types/src/lib.rs
```diff
@@ -274,12 +274,56 @@ pub enum UpdatingData<T: Data> {
...
+    fn try_from(value: DebeziumShadow<T>) -> Result<Self, Self::Error> {
+        match value.op {
+            DebeziumOp::Create => {
+                if value.after.is_none() {
+                    return Err("`after` must be set for Debezium create messages");
+                }
+            }
+            DebeziumOp::Update => {
+                if value.before.is_none() {
+                    return Err("`before` must be set for Debezium update messages; you may need to set REPLICA IDENTIFY FULL on your database");
```
jacksonrnewhouse, line 305: I think this is postgres specific behavior. Worth calling out?

jacksonrnewhouse (approved):

arroyo-types/src/lib.rs
mwylde: yeah, I'll add that -- unfortunately no way to know here what the database is

PR #249, opened 2023-08-14
Show warning message if we can't get metrics

jbeisen:
This situation happens if Prometheus isn't running.

---

image


mwylde:
When I start up a pipeline with an operator box selected, I'm getting this error:

image

I think it might predate this PR but would be good to fix as well

jbeisen:
@mwylde Good catch, fixed.

jacksonrnewhouse (approved):

Issue #260, opened 2023-08-18
Virtual fields are included in source struct


mwylde:
When creating a table with virtual fields (for example ,to generate a computed timestamp or watermark field), the source struct should only contain the physical fields, as this is used to deserialize data off of the source, then the virtual fields should be added in a post-source filter step. 

However (possibly as a consequence of the changes in #184) virtual fields _are_ being included in the source struct. 

For example:

```sql
create table stream (
  timestamp BIGINT NOT NULL,
  event_time TIMESTAMP NOT NULL GENERATED ALWAYS AS (CAST(from_unixtime(timestamp * 1000000000) as TIMESTAMP))
) WITH (
  connector = 'kafka',
  bootstrap_servers = 'localhost:9092',
  topic = 'stream',
  format = 'json',
  type = 'source',
  event_time_field = 'event_time'
);

select * from stream;
```

Creates a pipeline with this source struct:

```rust
pub struct generated_struct_5280044764730053267 {
    pub timestamp: i64,
    #[serde(with = "arroyo_worker::formats::timestamp_as_rfc3339")]
    pub event_time: std::time::SystemTime,
}
```

But because the virtual field is non-nullable, this causes deserialization errors when the events (correctly) not include the compute `event_time` field.

Issue #259, opened 2023-08-18
SQL compilation failure when using non-nullable timestamp as event_time_field or watermark_field


mwylde:
This query fails to compile:

```sql
create table demo_stream (
  timestamp BIGINT NOT NULL,
  event_time TIMESTAMP GENERATED ALWAYS AS (CAST(from_unixtime(timestamp * 1000000000) as TIMESTAMP))
) WITH (
  connector = 'kafka',
  bootstrap_servers = 'localhost:9092',
  topic = 'demo-stream',
  format = 'json',
  type = 'source',
  event_time_field = 'event_time'
);

select * from demo_stream;
```

When setting the even_time_field and watermark_field, we currently assume those timestamps are nullable:

```
error[E0599]: no method named `expect` found for struct `SystemTime` in the current scope
   --> pipeline/src/main.rs:141:46
    |
138 |   ...                   timestamp: arg
    |  __________________________________-
139 | | ...                       .event_time
140 | | ...                       .clone()
141 | | ...                       .expect("require a non-null timestamp"),
    | |                           -^^^^^^ method not found in `SystemTime`
    | |___________________________|
    |

error[E0599]: no method named `unwrap_or_else` found for struct `SystemTime` in the current scope
   --> pipeline/src/main.rs:175:42
    |
173 | / ...                   arg.watermark
174 | | ...                       .clone()
175 | | ...                       .unwrap_or_else(|| std::time::SystemTime::now())
    | |                           -^^^^^^^^^^^^^^ method not found in `SystemTime`
    | |___________________________|
```

jacksonrnewhouse:
This is caused because of a mismatch in the data type of the watermark field according to the CREATE TABLE signature `event_time TIMESTAMP` and the expression to compute it `CAST(from_unixtime(timestamp * 1000000000) as TIMESTAMP)`. In particular, the `event_time TIMESTAMP` doesn't imply a not-null constraint, but the expression is not nullable. This results in a StructField whose type is not-null but the generated expression is nullable, and thus an Option. We could throw an error for it at https://github.com/ArroyoSystems/arroyo/blob/8f7cb3ca757661145f441515a17f2e470e66e3d1/arroyo-sql/src/tables.rs#L487. We could also add a CAST to the declared type in the case of a mismatch.

PR #264, opened 2023-08-21
sql: Refactor virtual expressions, cast on type mismatch virtual expressions.

jacksonrnewhouse:
This addresses #259 and #260.


mwylde (approved):
Does this also work for other implicit casts, like numeric widening (INT -> BIG INT)?

PR #265, opened 2023-08-22
Switch checkpoint details to REST API

jbeisen:
Deprecate the GRPC method and use the new REST endpoint in the console.
This does away with the checkpoint timelines and instead shows the
duration of each checkpointing span.

This also introduce a format for non-paginated collections as API
response objects.

---

image


arroyo-rpc/src/types.rs
```diff
@@ -209,23 +209,66 @@ pub struct Checkpoint {
...
+    pub start_time: u64,
+    pub finish_time: u64,
+    pub span_type: CheckpointSpanType,
+    pub description: String,
+}
+
+#[derive(Serialize, Deserialize, Clone, Debug, ToSchema)]
+#[serde(rename_all = "camelCase")]
+pub struct SubtaskCheckpointGroup {
+    pub idx: u32,
```
mwylde, line 233: maybe "index" — a bit clearer for a public api?

mwylde (approved):

PR #269, opened 2023-08-23
Send connection profile config as json and validate it

jbeisen:


mwylde (approved):

arroyo-connectors/src/lib.rs
```diff
@@ -178,7 +178,8 @@ impl<C: Connector> ErasedConnector for C {
         Ok(self.config_description(self.parse_config(s)?))
     }
 
-    fn validate_config(&self, _: &serde_json::Value) -> Result<(), serde_json::Error> {
+    fn validate_config(&self, config: &serde_json::Value) -> Result<(), serde_json::Error> {
+        self.parse_config(config)?;
```
mwylde, line 182: This change needs to be made for validate_table as well — looks like we lost the validation logic in the refactor

jbeisen: Right, done.

mwylde (approved):

PR #281, opened 2023-08-29
Do not panic when epoch is not found

jbeisen:


arroyo-controller/src/states/scheduling.rs
```diff
@@ -350,9 +350,16 @@ impl State for Scheduling {
...
         {
-            let mut metadata = StateBackend::load_checkpoint_metadata(&ctx.config.id, epoch)
+            let mut metadata = match StateBackend::load_checkpoint_metadata(&ctx.config.id, epoch)
                 .await
-                .unwrap_or_else(|| panic!("epoch {} not found for job {}", epoch, ctx.config.id));
+            {
+                Some(metadata) => metadata,
+                None => {
+                    let message = format!("epoch {} not found for job {}", epoch, ctx.config.id);
+                    return Err(fatal(message.clone(), anyhow!(message)));
```
mwylde, line 359: The message (first argument to fatal) is what's shown to the user in the web ui, so it should be short and easy for end users to understand. It also doesn't need to include the job id, because that's logged separately.

arroyo-controller/src/states/scheduling.rs
```diff
@@ -350,9 +350,16 @@ impl State for Scheduling {
             needs_commits,
         }) = checkpoint_info.clone()
         {
-            let mut metadata = StateBackend::load_checkpoint_metadata(&ctx.config.id, epoch)
+            let mut metadata = match StateBackend::load_checkpoint_metadata(&ctx.config.id, epoch)
```
mwylde, line 353: Nit: using `ok_or_else` and `?` is generally preferable to a match statement where you're just trying to map the error case.

mwylde (approved):

arroyo-controller/src/states/scheduling.rs
jbeisen: True, I've changed it.
image

PR #282, opened 2023-08-29
Support deletes in the parquet backend

jbeisen:
Modify the parquet schema to support writing deletions. Currently only
the KeyedState table performs any deletions. Reading the parquet file
when restoring from a checkpoint accounts for deletes.

This also adds a new mechanism for generational compaction. After
completing a checkpoint, the controller will look at the backend and for
each operator and table, if there are at least MIN_FILES_TO_COMPACT
(currently set to 4) of the same generation, it will read those files,
account for deletes, and write a set of newly partitioned, compacted
files. The controller then calls an rpc on the workers to trigger them
to load that new data.

Compaction defaults to disabled, and can be enable with the environment
variable COMPACTION_ENABLED.

---

Here's a sequence diagram for how compaction works:

image

### DIFF
--- a/arroyo-state/src/lib.rs
+++ b/arroyo-state/src/lib.rs
@@ -1,16 +1,19 @@
+use crate::tables::DataTuple;
 use anyhow::Result;
 use arroyo_rpc::grpc::{
     CheckpointMetadata, OperatorCheckpointMetadata, TableDeleteBehavior, TableDescriptor,
     TableType, TableWriteBehavior,
 };
-use arroyo_rpc::ControlResp;
+use arroyo_rpc::{CompactionResult, ControlResp};
 use arroyo_types::{CheckpointBarrier, Data, Key, TaskInfo};
 use async_trait::async_trait;
 use bincode::config::Configuration;
+use bincode::{Decode, Encode};
 use std::any::Any;
 use std::collections::hash_map::DefaultHasher;
 use std::collections::HashMap;
 use std::hash::{Hash, Hasher};
+use std::ops::RangeInclusive;
 use std::time::{Duration, SystemTime};
 use tables::{
     GlobalKeyedState, GlobalKeyedStateCache, KeyTimeMultiMap, KeyTimeMultiMapCache, KeyedState,
@@ -18,10 +21,12 @@ use tables::{
 };
 use tokio::sync::mpsc::Sender;
 
+mod metrics;
 pub mod parquet;
 pub mod tables;
 
 pub const BINCODE_CONFIG: Configuration = bincode::config::standard();
+pub const FULL_KEY_RANGE: RangeInclusive<u64> = 0..=u64::MAX;
 
 pub type StateBackend = parquet::ParquetBackend;
 
@@ -50,6 +55,25 @@ pub fn timestamp_table(
         delete_behavior: delete_behavior as i32,
         write_behavior: write_behavior as i32,
         retention_micros: retention.as_micros() as u64,
+    }
+}
+
+#[derive(Debug, Encode, Decode)]
+#[repr(u8)]
+pub enum DataOperation {
+    Insert = 0,
+    DeleteKey = 1, // delete single key/value pair of Global/TimeKeyMap
+                   // DeleteValue,  // delete single value of a KeyTimeMultiMap
+                   // DeleteBefore, // delete all values for key before timestamp (only for KeyTimeMultiMap)
+}
+
+impl From<u8> for DataOperation {
+    fn from(op: u8) -> Self {
+        match op {
+            0 => DataOperation::Insert,
+            1 => DataOperation::DeleteKey,
+            _ => panic!("Unknown DataOperation {}", op),
+        }
     }
 }
 
@@ -82,17 +106,19 @@ pub trait BackingStore {
 
     fn name() -> &'static str;
 
-    // parepares a checkpoint to be written
+    fn task_info(&self) -> &TaskInfo;
+
+    // prepares a checkpoint to be written
     #[allow(unused_variables)]
     async fn initialize_checkpoint(job_id: &str, epoch: u32, operators: &[&str]) -> Result<()> {
         Ok(())
     }
 
-    async fn complete_operator_checkpoint(metadata: OperatorCheckpointMetadata);
+    async fn write_operator_checkpoint_metadata(metadata: OperatorCheckpointMetadata);
 
     async fn complete_checkpoint(metadata: CheckpointMetadata);
 
-    async fn compact_checkpoint(
+    async fn cleanup_checkpoint(
         metadata: CheckpointMetadata,
         old_min_epoch: u32,
         new_min_epoch: u32,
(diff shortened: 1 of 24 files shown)


arroyo-state/src/parquet.rs
```diff
@@ -623,7 +694,14 @@ impl Default for ParquetStats {
 }
 
 impl RecordBatchBuilder {
-    fn insert(&mut self, key_hash: u64, timestamp: SystemTime, key: Vec<u8>, data: Vec<u8>) {
+    fn insert(
```
jacksonrnewhouse, line 697: Rather than `insert` taking a DataOperation we could have `insert()` and `remove()` separately.

jbeisen: I thought about that, but what's really happening is an _insert_ into the `RecordBatchBuilder`, so in the context of this implementation of a backing store and this parquet record batch builder, I think these semantics are accurate.

arroyo-state/src/tables.rs
```diff
@@ -316,16 +325,23 @@ impl<K: Key, V: Data> KeyTimeMultiMapCache<K, V> {
...
+                DataOperation::Insert => {
+                    values
+                        .entry(key)
+                        .or_default()
+                        .entry(timestamp)
+                        .or_default()
+                        .push(value);
+                }
+                DataOperation::Delete => {
+                    values.entry(key).or_default().remove(&timestamp); // TODO: delete just the value?
```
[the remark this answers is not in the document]
jbeisen: For option 4 would I think there are two ways to implement it:
a. extend the `DataOperation::Delete` enum variant to include a timestamp, store that in parquet, and expand that into the individual deletes at checkpoint restoration time.
b. expand the operation into individual deletes at cache update time and write the individual deletes to parquet.

I think option a is more efficient because we only need to do a single linear search of the vector, right?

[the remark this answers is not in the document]
jbeisen: The enum could look something like this:
```
pub enum DataOperation {
    Insert,
    DeleteKey, // delete single key/value if Global/TimeKeyMap, delete all values for key if KeyTimeMultiMap 
    DeleteBefore, // delete all values for key before timestamp (only for KeyTimeMultiMap)
}
```

arroyo-state/src/lib.rs
```diff
@@ -115,10 +130,30 @@ pub trait BackingStore {
...
+        table_type: TableType,
+        before: SystemTime,
+        key: &mut K,
+    );
+
     async fn write_key_value<K: Key, V: Data>(&mut self, table: char, key: &mut K, value: &mut V);
+    async fn delete_key_value<K: Key>(&mut self, table: char, key: &mut K);
 
-    async fn get_global_key_values<K: Key, V: Data>(&self, table: char) -> Vec<(K, V)>;
-    async fn get_key_values<K: Key, V: Data>(&self, table: char) -> Vec<(K, V)>;
```
jbeisen, line 121: I combined these functions since the only difference was the key range.

jacksonrnewhouse: I don't love this, as it pushes up a single key assignment strategy (one contiguous hash range per subtask) into the API. Having the two methods results in semantics of "get the data the subtask owns" and "get all of the data for the subtask".

arroyo-state/src/parquet.rs
```diff
@@ -378,48 +438,148 @@ impl BackingStore for ParquetBackend {
...
+        storage_client: StorageProvider,
+        new_min_epoch: u32,
+        table_descriptor: &TableDescriptor,
+    ) -> Option<ParquetStoreData> {
+        // accumulate this partition's tuples from all the table's files
+        // (spread across multiple epochs and files)
+        let mut tuples_in = vec![];
+        for (_epoch, files) in epoch_files {
+            for file in files {
+                // TODO: should we exclude files that are already compact?
```
jacksonrnewhouse, line 478: I think probably for the initial version. We'll want phased compaction in the future, but I worry that not doing so means we'll end up copying all live data at whatever the compaction interval is.

arroyo-state/src/parquet.rs
```diff
@@ -378,48 +438,148 @@ impl BackingStore for ParquetBackend {
...
-                    let Some(BackendData::ParquetStore(parquet_store)) = &backend_data.backend_data
-                    else {
-                        unreachable!("expect parquet backends")
-                    };
-                    parquet_store.file.clone()
-                })
-                .collect();
+                .expect("expect new_min_epoch metadata to still be present");
+
+        // these files will be compressed, so the worker will be able to drop them
```
jacksonrnewhouse, line 553: I think we should try to use "compacted" rather than "compressed" throughout, as compression is happening on a different axis.

arroyo-state/src/parquet.rs
```diff
@@ -428,17 +588,85 @@ impl ParquetBackend {
...
     }
-    fn triples_from_parquet_bytes<K: Key, V: Data>(
+
+    /// Return rows from the given bytes that are in the given key range
+    fn tuples_from_parquet_bytes<K: Key, V: Data>(
         &self,
         bytes: Vec<u8>,
         range: &RangeInclusive<u64>,
-    ) -> Vec<(SystemTime, K, V)> {
+    ) -> Vec<(SystemTime, K, Option<V>, DataOperation)> {
```
jacksonrnewhouse, line 669: 4-tuples are scary, might be worth just defining a struct for this.

arroyo-state/src/parquet.rs
```diff
@@ -479,19 +712,162 @@ impl ParquetBackend {
...
+        record_batch: arrow_array::RecordBatch,
+        storage: &StorageProvider,
+    ) -> Result<usize> {
+        let props = WriterProperties::builder()
+            .set_compression(parquet::basic::Compression::ZSTD(ZstdLevel::default()))
+            .set_statistics_enabled(EnabledStatistics::None)
+            .build();
+        let cursor = Vec::new();
+        let mut writer = ArrowWriter::try_new(cursor, record_batch.schema(), Some(props)).unwrap();
+        writer.write(&record_batch).expect("Writing batch");
```
jacksonrnewhouse, line 838: Given this returns a Result, I'd just add a `?` rather than the expect and unwrap calls.

arroyo-state/src/lib.rs
jbeisen: Ok I've added the two functions back to the trait but made their implementations both use the same function.

yf-sz:
您的信件已收到，谢谢！

arroyo-state/src/parquet.rs
jbeisen: I've changed this to only compact files of the specified generation, so this shouldn't be a problem.

arroyo-state/src/tables.rs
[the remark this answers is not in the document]
jbeisen: I've described the operations in the [doc](https://docs.google.com/document/d/1430Raxk4jmJSZOhEjNbUGdBex9hCC4UZhyshlPaTaa8/edit), but the initial implementation let's just do `DeleteKey`.

arroyo-state/src/tables.rs
```diff
@@ -1,10 +1,85 @@
...
+                            data.entry(tuple.timestamp)
+                                .or_default()
+                                .insert(tuple.key, tuple.value);
+                        }
+                        DataOperation::DeleteKey => {
+                            // for each timestamp, remove the key
+                            for (_, map) in data.iter_mut() {
+                                map.remove(&tuple.key);
+                            }
+                        }
```
jbeisen, line 55: @jacksonrnewhouse This is the main compaction logic for time key maps, please take a look.

jacksonrnewhouse (approved):

PR #293, opened 2023-09-01
Enabling windowed/non-window join

edmondop:


jacksonrnewhouse:
I'm a little surprised that this just compiled, and have some concerns about the correctness of the resulting pipeline. Once we push out some more testing capabilities we should evaluate whether the code here does the correct thing.

edmondop:
Agreed, would love to partecipate to setting up some more testing capabilities

PR #295, opened 2023-09-04
Show that preview pipeline has completed

jbeisen:
Show a success alert when a preview job has completed. Also close the EventSource when stopping a preview so that there are not multiple pending requests.

---
image


arroyo-console/src/routes/pipelines/CreatePipeline.tsx
```diff
@@ -183,6 +185,14 @@ export function CreatePipeline() {
     setTabIndex(1);
   };
 
+  const stopPreview = async () => {
+    await updatePipeline({ stop: 'immediate' });
+
+    if (outputSource) {
```
mwylde, line 191: I wonder if instead of doing this when the user clicks "stop" we should instead be following the state machine of the pipeline so that we can also close it when the job stops on its own

mwylde (approved):

PR #297, opened 2023-09-06
Webhook sink

mwylde:
This PR adds a webhook sink, which makes an HTTP POST request to a user-specified server.

Failure handling for sinks is a complex issue, particularly when the sink is talking to arbitrary web servers that may not always be available.

Ideally we would have the following properties:
1. All events are eventually written successfully (with a 200 status code) to the HTTP server
2. If the server is down (or unable to keep up with our volume) we should backpressure
3. None of this should block checkpointing

This PR achieves 1 and 2 by retrying requests indefinitely; backpressure is achieved by limiting the maximum number of in-flight requests (currently hardcoded to 50).

However, this does block checkpointing until we are able to successfully write the inflight requests.

An alternative approach could be to store in-flight requests in state so that we can continue to consume data and allow checkpoints; this would achieve 1 and 3 but means we lose our ability to backpressure with potentially unbounded state and work growth (this also requires state deletion).

Ultimately I think a good implementation is going to require the ability to checkpoint under backpressure (for example by following Flink's unaligned checkpoint approach).


arroyo-connectors/src/webhook.rs
```diff
@@ -0,0 +1,207 @@
...
+
+        let client = reqwest::ClientBuilder::new()
+            .default_headers(headers?)
+            .timeout(Duration::from_secs(5))
+            .build()
+            .map_err(|e| anyhow!("could not construct HTTP client: {:?}", e))?;
+
+        let req = client
+            .post(&config.endpoint)
+            // TODO: use the schema to construct a correctly-formatted message
```
jacksonrnewhouse, line 54: Do we need to do this before merging?

arroyo-worker/src/connectors/webhook.rs
```diff
@@ -0,0 +1,187 @@
...
+    formats::DataSerializer,
+    SchemaData,
+};
+
+import_types!(schema = "../connector-schemas/webhook/table.json");
+
+const MAX_INFLIGHT: u32 = 50;
+
+#[derive(Clone, Debug, Encode, Decode, PartialEq, PartialOrd, Default)]
+pub struct WebhookSinkState {}
```
jacksonrnewhouse, line 31: What's the point of an empty struct stored on the key `()`? It looks like this could be removed without affecting behavior. Leftover from a previous approach?

jacksonrnewhouse (approved):
Looks good overall, would like to understand if the WebhookSinkState is necessary.

arroyo-connectors/src/webhook.rs
mwylde: I think it's a nice-to-have (if the receiving endpoint is expecting a particular format, it may return an error during testing) but the user can still decide to create the source.

arroyo-worker/src/connectors/webhook.rs
mwylde: Yeah, it can be removed

PR #299, opened 2023-09-07
Integration tests

jacksonrnewhouse:
This is a sizable PR with the overall goal of providing correctness assertions for Arroyo SQL Pipelines. It includes the following components.

### Single File Source and Sink
This is a new connection type that is intended to read and write single files of newline delimited JSON. Because it looks on the local file system and only handles a single file it isn't really intended for production use-cases. The new smoke_tests.rs use this to write the output of a test to an output file and read from a fixed input. Should we consider gating this for "prod" queries in some manner?

### Restructure Program Running
Prior to this there was a general assumption that a Program would be invoked by a controller and sending various control messages back via GRPC APIs. We had hacked around this for the local runner by letting the GRPC client be an Option and only using it if it was present. Instead, starting a pipeline now returns a Receiver<ControlResp>, and then the invoker can do with it what it will. For GRPC this is still using the same APIs, while for the correctness tests we poll it directly until the job finishes.

### Ordering of Stopping
There was an issue with the ordering of `Message::EndOfData` and `Message::Watermark` out of the PeriodicWatermarkGenerator. In particular, the EndOfData was getting sent first, meaning that the max timestamp watermark was never propogating downstream, flushing data out of stateful operators. This is addressed by changing the semantics in arroyo-macro. Now, on_close() is called before a finishing message is sent downstream.

### Correctness Testing
The macros in arroyo-sql-macro were reworked to support making assertions on some output file which should be written after running a query. This is done by the `correctness_run_codegen!{}` macro, which takes in a test name and some SQL. The SQL will have $input_dir replaced with the absolute path of arroyo-sql-testing/inputs and $output_path replaced with the absolute path of arroyo-sql-testing/outputs/$TEST_NAME.json. After running the pipeline that output path will be tested against the same in arroyo-sql-testing/golden_outputs and assert on sorted equality.

### SQL Changes
The behavior around window functions was ambiguous and would compile queries that were invalid, such as ones involving two different windows. This has been reworked and the relevant correctness tests now pass. Similarly, there were some introduced by the CodeGenerator refactor in joining windowed outputs which have been addressed.


arroyo-worker/src/lib.rs
```diff
@@ -187,15 +190,28 @@ impl LocalRunner {
...
         loop {
-            tokio::time::sleep(Duration::from_secs(1)).await;
+            while let Some(control_message) = control_rx.recv().await {
+                println!("received {:?}", control_message);
+                if let ControlResp::TaskFinished {
+                    operator_id: _,
+                    task_index: _,
+                } = control_message
+                {
+                    running_nodes -= 1;
```
mwylde, line 209: I think it's a bit more reliable to insert into a hashset and count that instead of just counting, so that you make sure you get a message from every subtask and not multiple messages from one

arroyo-worker/src/lib.rs
```diff
@@ -316,6 +332,125 @@ impl WorkerServer {
...
+        worker_id: WorkerId,
+        job_id: String,
+    ) -> Result<()> {
+        let mut controller = ControllerGrpcClient::connect(self.controller_addr.clone()).await?;
+        tokio::spawn(async move {
+            let mut tick = tokio::time::interval(Duration::from_secs(5));
+            tick.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Skip);
+            loop {
+                select! {
+                        msg = control_rx.recv() => {
```
mwylde, line 349: nit: indentation within the select is off (rustfmt ignores everything inside macros)

arroyo-sql-testing/src/lib.rs
```diff
@@ -1,4 +1,5 @@
 mod full_query_tests;
+mod smoke_tests;
```
mwylde, line 2: I think it's worth putting this (and probably full_query_tests as well) behind a flag. On my machine, this branch takes 60 to check from clean, compared to 45 for master and 40 seconds without full_query_tests.

mwylde (approved):
This is an amazing contribution.

arroyo-worker/src/connectors/filesystem/single_file/sink.rs
```diff
@@ -0,0 +1,99 @@
...
+                .await
+                .unwrap(),
+        );
+        if offset > 0 {
+            self.file.as_mut().unwrap().set_len(offset).await.unwrap();
+        }
+    }
+
+    async fn process_element(&mut self, record: &Record<K, T>, _ctx: &mut Context<(), ()>) {
+        let row = serde_json::to_string(&record.value).unwrap();
```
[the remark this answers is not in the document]
jacksonrnewhouse: Will do

PR #300, opened 2023-09-07
Add stepper to create pipeline page

jbeisen:
Use a stepper component to show the steps of creating a pipeline and for showing the status of each step (sql validation and preview job status).

Also put the buttons in line with the tabs and make put the components into resizable panels.

image


jbeisen:
Closing since I decided that the stepper had too many confusing states.

PR #301, opened 2023-09-07
Use resizable panels on create pipeline page

jbeisen:
Also puts the buttons in line with the tabs and shows the preview job status in the Results tab.

image


mwylde (approved):
nice, looks great!

PR #302, opened 2023-09-09
Add support for serializing raw_string formats

mwylde:
This PR adds support for `raw_string` as an output format, which writes a single string field as UTF-8 encoded bytes.

It can be used like this:

```sql
CREATE TABLE raw_sink (
  output TEXT
) WITH (
  connector = 'kafka',
  type = 'sink',
  topic = 'outputs',
  format = 'raw_string'
);

INSERT INTO raw_sink
SELECT bid.channel
FROM nexmark;
```

When combined with UDFs utilizing serde_json functions, this allow serialization of arbitrarily complex structures that are not supported via our normal JSON support. For example, this UDF:

```rust
fn my_to_json(f: f64) -> String {
    let v = serde_json::json!({
        "my_complex": {
            "nested": f
        }
    });

    serde_json::to_string(&v).unwrap()
}
```

with this query:

```sql
INSERT INTO sink
SELECT my_to_json(avg(CAST(price as FLOAT))) from coinbase
WHERE type = 'ticker'
GROUP BY hop(interval '5' second, interval '1 minute');
```

Will produce records like 

```
{"my_complex":{"nested":25900.074265625}}
{"my_complex":{"nested":25899.815841674805}}
```

The implementation is not currently type safe, in the sense that `to_raw_string` is implemented for only valid output types (those that have a single TEXT field in their schema). We validate at the query compilation level that this is correct, but I don't see a way to have the type system enforce it.

### DIFF
--- a/arroyo-sql/src/types.rs
+++ b/arroyo-sql/src/types.rs
@@ -241,6 +241,26 @@ impl StructDef {
         let schema_data_impl = if self.name.is_none() {
             // generate a SchemaData impl but only for generated types
             let name = self.struct_name();
+
+            let to_raw_string = if self.fields.len() == 1
+                && matches!(
+                    self.fields[0].data_type,
+                    TypeDef::DataType(DataType::Utf8, _)
+                ) {
+                let field = &self.fields[0].field_ident();
+                if self.fields[0].nullable() {
+                    quote! {
+                        self.#field.as_ref().map(|v| v.as_bytes().to_vec())
+                    }
+                } else {
+                    quote! {
+                        self.#field.as_bytes().to_vec()
+                    }
+                }
+            } else {
+                quote! { unimplemented!("to_raw_string is not implemented for this type") }
+            };
+
             Some(quote! {
                 impl arroyo_worker::SchemaData for #struct_type {
                     fn name() -> &'static str {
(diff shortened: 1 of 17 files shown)


arroyo-sql/src/types.rs
```diff
@@ -241,6 +241,26 @@ impl StructDef {
...
+                    TypeDef::DataType(DataType::Utf8, _)
+                ) {
+                let field = &self.fields[0].field_ident();
+                if self.fields[0].nullable() {
+                    quote! {
+                        self.#field.as_ref().map(|v| v.as_bytes().to_vec())
+                    }
+                } else {
+                    quote! {
+                        self.#field.as_bytes().to_vec()
```
jacksonrnewhouse, line 257: I think to_raw_string() is expecting an `Option<Vec<u8>>`

jacksonrnewhouse (changes requested):
Can you add a compilation test with a field like `raw_value TEXT NOT NULL`?

arroyo-sql/src/types.rs
mwylde: Good catch, fixed and added a test

jacksonrnewhouse (approved):

PR #311, opened 2023-09-12
Polling http source

mwylde:
This PR adds a new source, `polling_http`, which polls an HTTP endpoint at a periodic (configurable) rate, and emits for either every result returned or every time it changes.

It's used like this:

```sql
CREATE TABLE polling_source (
  value TEXT NOT NULL
) WITH (
  connector = 'polling_http',
  endpoint = 'http://localhost:9091',
  headers = 'Authorization: Bearer 1234,Content-Type: application/json',
  method = 'POST',
  body = '{}',
  emit_behavior = 'changed',
  format = 'raw_string'
);
```

This source is useful for turning a non-streaming API into a streaming one.


arroyo-connectors/src/polling_http.rs
```diff
@@ -0,0 +1,221 @@
...
+    fn construct_test_request(
+        client: &Client,
+        config: &PollingHttpTable,
+    ) -> anyhow::Result<Request> {
+        let mut req = client.request(
+            match config.method {
+                None | Some(Method::Get) => reqwest::Method::GET,
+                Some(Method::Post) => reqwest::Method::POST,
+                Some(Method::Put) => reqwest::Method::PUT,
+                Some(Method::Patch) => reqwest::Method::PATCH,
```
jbeisen, line 35: Why would a PATCH request be used? Or really anything other than GET?

arroyo-worker/src/connectors/polling_http.rs
```diff
@@ -0,0 +1,260 @@
...
+                unreachable!("sources shouldn't receive commit messages");
+            }
+        }
+        None
+    }
+
+    async fn request(&mut self) -> Result<T, UserError> {
+        let mut request = self
+            .client
+            .request(self.method.clone(), self.endpoint.clone());
```
jbeisen, line 163: I'm wondering if there should be a timeout and a maximum body size

arroyo-connectors/src/polling_http.rs
mwylde: There are APIs that you need to POST to, and seemed worth not artificially limiting it

arroyo-worker/src/connectors/polling_http.rs
mwylde: There's a 5 second timeout, which could be made configurable. A max response size is probably a good idea, but not straightforward to implement with reqwest.

mwylde: added a 5MB response size limit

jacksonrnewhouse (approved):

jbeisen (approved):

PR #312, opened 2023-09-13
implement UDAF support

jacksonrnewhouse:
Summary
This PR introduces User-Defined Aggregate Function (UDAF) support to the Arroyo stream processing system. Users can now write Rust UDAFs by defining rust functions with Vec arguments. This includes support for multiple-argument UDF, but does not currently support sorting, due to https://github.com/apache/arrow-datafusion/issues/7531. Once that is fixed we can support it.

Changes
Modified arroyo-sql-macro/src/lib.rs to allow for passing in a UDF block, which can be used for testing UDFs and UDAFs.
Added UDAF support in arroyo-sql/src/expressions.rs.
Introduced a new enum AggregateComputation, which can be either Builtin or UDAF.

Testing
Added new test cases for UDAFs and UDFs.

Backward Compatibility
The changes are backward compatible; existing functionalities should not be affected.

Future Work
Support for aggregate filters and sorts which are currently not supported. Support for two-phase aggregations, which store intermediate computations in the state store.


arroyo-sql/src/lib.rs
```diff
@@ -180,13 +203,26 @@ impl ArroyoSchemaProvider {
             };
 
             let mut args: Vec<TypeDef> = vec![];
+            let mut vec_arguments = 0;
             for (i, arg) in function.sig.inputs.iter().enumerate() {
+                info!("arg {}: {:?}", i, quote::quote!(#arg).to_string());
```
mwylde, line 208: this is a bit noisy — maybe downgrade to debug?

arroyo-sql/src/lib.rs
```diff
@@ -180,13 +203,26 @@ impl ArroyoSchemaProvider {
...
                 match arg {
                     FnArg::Receiver(_) => bail!("self types are not allowed in UDFs"),
                     FnArg::Typed(t) => {
-                        args.push((&*t.ty).try_into().map_err(|_| {
-                            anyhow!("Could not convert arg {} into a SQL data type", i)
-                        })?);
+                        if let Some(vec_type) = Self::vec_inner_type(&*t.ty) {
+                            info!("arg {} is a vector", i);
```
mwylde, line 213: ditto

mwylde (approved):
Looks great, and love the PR description!

arroyo-sql/src/lib.rs
jacksonrnewhouse: These were just me trying to understand the existing code, will delete.

Issue #307, opened 2023-09-12
AWS region for Kinesis


MarcoLugo:
I was using the Kinesis source earlier within Kubernetes and had an issue with the region. This can probably be worked around with an environment variable but maybe AWS region should be a Kinesis parameter?

Thanks

<img width="996" alt="image" src="https://github.com/ArroyoSystems/arroyo/assets/6843447/bffb1ec0-918c-42ab-a000-7f02efc9f156">

mwylde:
Fixed in #313

PR #313, opened 2023-09-13
enable configuration of aws_region for Kinesis sources and sinks.

jacksonrnewhouse:
This addresses #307  by allowing users to manually configure the AWS region for their Kinesis sources and sinks.


arroyo-worker/src/connectors/kinesis/source/mod.rs
```diff
@@ -370,6 +372,19 @@ impl<K: Data, T: Data + DeserializeOwned> KinesisSourceFunc<K, T> {
...
 
+    async fn init_client(&mut self) {
+        let config = match self.aws_region {
+            Some(ref region) => {
+                aws_config::from_env()
+                    .region(Region::new(region.to_string()))
+                    .load()
+                    .await
+            }
+            None => load_from_env().await,
```
mwylde, line 383: I'm a bit confused why here we're using load_from_env(), but above we're using `from_env` — is this the same aside from the region override?

mwylde (approved):

arroyo-worker/src/connectors/kinesis/source/mod.rs
jacksonrnewhouse: `load_from_env()` is a very thin wrapper around `from_env()`, probably makes sense to just use `from_env()`. Here's `load_from_env()`:
```
/// Load a default configuration from the environment
///
/// Convenience wrapper equivalent to `aws_config::from_env().load().await`
pub async fn load_from_env() -> aws_types::SdkConfig {
    from_env().load().await
}

PR #314, opened 2023-09-13
Better support for service accounts in helm and other helm improvements

mwylde:
This PR makes a small improvement to the helm chart to improve the process of configuring service accounts for k8s deployments. This was in service of making it easier to configure Arroyo to run on GKE and talk to GCS, which typically is done via a service account (as described [here](https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity#using_from_your_code)). Specifically, we previously would only forward the configured service account to the workers if `serviceAccount.create` was `true` in the helm chart, whereas now we will do so if `serviceAccount.name` is non-empty.

This PR also adds the ability to configure the deployments to use a pre-existing configmap as environment variables, which makes it easier to set arbitrary configs across our services.

Finally, it cleans up some leftover config from the old API gRPC service ports, as now only the HTTP port is needed.


arroyo-controller/src/schedulers/kubernetes.rs
```diff
@@ -83,15 +89,11 @@ impl KubernetesScheduler {
...
         }
     }
-}
-
-#[async_trait]
-impl Scheduler for KubernetesScheduler {
-    async fn start_workers(&self, req: StartPipelineReq) -> Result<(), SchedulerError> {
-        let api: Api<ReplicaSet> = Api::default_namespaced(self.client.clone());
 
+    fn make_replicaset(&self, req: StartPipelineReq) -> ReplicaSet {
```
jacksonrnewhouse, line 96: `make_replica_set`?

k8s/arroyo/templates/NOTES.txt
```diff
@@ -6,11 +6,11 @@ Once the release is fully deployed, you can navigate to the web ui by running
 
   $ open "http://$(kubectl get service/{{ .Release.Name }}-api -o jsonpath='{.spec.clusterIP}')"
 
-(note this might take a minute or so if you are also deploying Postgres)
+(note this might take a few minutes so if you are also deploying Postgres)
```
jacksonrnewhouse, line 9: wording. should be "note that this might take a few minutes if you are also deploying Postgres"

jacksonrnewhouse (commented):
Just a couple typos, looks good otherwise.

arroyo-controller/src/schedulers/kubernetes.rs
mwylde: no, replicaset is one word

jacksonrnewhouse (approved):

PR #321, opened 2023-09-14
Bugfix: skip lines from file in `single_file` source

rcjmurillo:
Lines weren't being actually skipped when restoring to previous source state.


jacksonrnewhouse:
Thanks for catching this. The source is pretty new and intended mainly for integration tests. We have testing checkpoint restoration with it scheduled, which probably would've surfaced it, but thanks for catching it!

jacksonrnewhouse (approved):

PR #325, opened 2023-09-14
adding support for custom consumer group

### LINKED ISSUE #308: Support configuring Kafka consumer group
Currently we set the Kafka consumer group based on a fixed pattern: https://github.com/ArroyoSystems/arroyo/blob/2ff997fd3bca7f25fcb7932d2395ee782d183092/arroyo-worker/src/connectors/kafka/source/mod.rs#L132-L137

For some use cases, it would be valuable to be able to set this to a particular static value.

We should enable this by adding an option to the kafka source connector.

harshit2283:
Closes #308


connector-schemas/kafka/connection.json
```diff
@@ -49,6 +49,11 @@
                     "additionalProperties": false
                 }
             ]
+        },
+        "groupId": {
+            "type": "string",
+            "title": "Group Id",
+            "description": "Custom Group ID (Consumer Group) to use"
```
mwylde, line 56: This might match the other description a bit better as "Sets the Kafka consumer group id".

It might also be good to add a warning that if you run multiple pipelines with the same consumer group they will not receive all of the data, as it will be shared between them.

connector-schemas/kafka/connection.json
```diff
@@ -49,6 +49,11 @@
                     "additionalProperties": false
                 }
             ]
+        },
+        "groupId": {
```
mwylde, line 53: For clustered systems like Kafka, we have two different configurations: the connection profile, and the table. The connection profile provides a single place to configure shared cluster configurations that will be reused across many different instances. So for Kafka, that includes stuff like the bootstrap servers and authentication.

The table configuration is used for specific instances of the cluster. For Kafka this is stuff like the topic name and whether it's a source or sink. 

Since the group id is something you might want to be specific to a topic rather than across the entire cluster, I'd move to the table.json. And since it's only used for sources, it should probably be under the source configuration there.

arroyo-worker/src/connectors/kafka/source/mod.rs
```diff
@@ -131,10 +135,11 @@ where
             .set("enable.auto.commit", "false")
             .set(
                 "group.id",
-                format!(
+                self.group_id.unwrap_or(format!(
```
mwylde, line 138: This is a nit since this isn't a performance-sensitive part of the code, but generally the closure form `upwrap_or_else` is prefered to `unwrap_or` when the None case is going to do something expensive like call format.

```suggestion
                self.group_id.unwrap_or_else(|| format!(
```

mwylde (changes requested):
A few comments but overall looks great! Thanks for the contribution.

connector-schemas/kafka/connection.json
harshit2283: Altered the description, added caution inside it. Any other way to emphasis it?

connector-schemas/kafka/connection.json
harshit2283: Makes sense, have implemented this.

mwylde (approved):
This looks great!

mwylde (changes requested):
Hey @harshit2283 — it looks like we're not currently running CI for PRs from contributors. I'm working on fixing that, but in the meantime it looks like there's a test build error.

Could you make sure that `cargo test` passes and push again?

harshit2283:
> Hey @harshit2283 — it looks like we're not currently running CI for PRs from contributors. I'm working on fixing that, but in the meantime it looks like there's a test build error.
> 
> Could you make sure that `cargo test` passes and push again?

Fixed it.

mwylde (approved):
Thanks, and apologies for the CI issues.

PR #327, opened 2023-09-19
S3 source connector

rcjmurillo:
A new connector to the filesystem to read files from a prefix on S3, it supports reading plain files or compressed in zstd or gzip format.

It will recursively list every file from the prefix and stream each line in the specified format.

**Directly from query:**
```sql
CREATE TABLE file_lines (
  value string
) WITH (
  connector = 's3',
  bucket = 'my-bucket',
  prefix = 'some/prefix/to/list',
  region = 'us-west-2',
  compression_format = 'zstd'
);

SELECT * FROM file_lines;
```


arroyo-connectors/src/s3.rs
```diff
@@ -0,0 +1,129 @@
...
+
+    fn name(&self) -> &'static str {
+        "s3"
+    }
+
+    fn metadata(&self) -> arroyo_rpc::types::Connector {
+        arroyo_rpc::types::Connector {
+            id: "s3".to_string(),
+            name: "S3".to_string(),
+            icon: "".to_string(),
```
rcjmurillo, line 33: i'll need help to get a nice image for the UI :)

arroyo-worker/src/connectors/filesystem/s3/source.rs
```diff
@@ -0,0 +1,214 @@
...
+use crate::{engine::Context, SourceFinishType};
+
+use super::S3Table;
+
+#[derive(StreamNode)]
+pub struct S3SourceFunc<K: Data, T: DeserializeOwned + Data> {
+    bucket: String,
+    prefix: String,
+    region: String,
+    compression: String,
```
jacksonrnewhouse, line 23: This could be CompressionFormat.

arroyo-connectors/src/s3.rs
```diff
@@ -0,0 +1,129 @@
...
+    fn from_options(
+        &self,
+        name: &str,
+        opts: &mut std::collections::HashMap<String, String>,
+        schema: Option<&ConnectionSchema>,
+    ) -> anyhow::Result<crate::Connection> {
+        let bucket = pull_opt("bucket", opts)?;
+        let prefix = pull_opt("prefix", opts)?;
+        let region = pull_opt("region", opts)?;
+        let compression_format = match pull_opt("compression_format", opts)?.as_str() {
```
jacksonrnewhouse, line 116: This block should probably support at least `gzip` and `none`.

arroyo-connectors/src/s3.rs
```diff
@@ -0,0 +1,129 @@
...
+
+use super::Connector;
+
+const TABLE_SCHEMA: &str = include_str!("../../connector-schemas/s3/table.json");
+
+import_types!(schema = "../connector-schemas/s3/table.json");
+
+pub struct S3Connector {}
+
+impl Connector for S3Connector {
```
jacksonrnewhouse, line 20: We're trying to be somewhat agnostic about which blob store is used for storing files. We use the object_store crate for this, and it'd be great if this new source could do the same. Right now blob store sinks fall under the `filesystem` name. A common pattern for connectors that have both source and sink implementations is to take a `type` argument which can be either `source` or `sink`, and then each one has its own set of options.

arroyo-worker/src/connectors/filesystem/s3/source.rs
```diff
@@ -0,0 +1,214 @@
...
+        SourceFinishType::Final
+    }
+
+    async fn read_file<R: AsyncBufRead + Unpin>(
+        &mut self,
+        ctx: &mut Context<(), T>,
+        mut reader: Lines<R>,
+        obj_key: &str,
+        prev_lines_read: usize,
+    ) -> Option<SourceFinishType> {
```
jacksonrnewhouse, line 150: This function should probably return an error, in case it can't read the file, for instance due to a network issue.

arroyo-worker/src/connectors/filesystem/s3/source.rs
```diff
@@ -0,0 +1,214 @@
...
+        let mut lines_read = 0;
+        while let Ok(Some(s)) = reader.next_line().await {
```
jacksonrnewhouse, line 152: Given that next_line() is an async function, you might be able to rewrite this as a tokio select loop. Something like
```
loop {
    select! {
        line_res = reader.next_line() => {
            let ok_result = line_res?;
            match ok_result {
                Some(line) => self.handle_line(line),
                None => return Some(SourceFinishType::Final),
            }
        },
        msg_res = ctx.control_rx.try_recv() => {
            if let Ok(control_message) = msg_res {
                match self.process_control_message(ctx, control_message, &obj_key, lines_read).await {
                    None => {},
                    finish_type => return finish_type,
                }
            }
        }
    }
}

```

arroyo-worker/src/connectors/filesystem/s3/source.rs
```diff
@@ -0,0 +1,214 @@
...
+            // if we're restoring from checkpoint, skip until we find the line we were on
+            if lines_read < prev_lines_read {
+                lines_read += 1;
+                continue;
+            }
+            // json! will correctly handle escaping
+            let value = serde_json::from_value(serde_json::json!({
```
jacksonrnewhouse, line 159: Rather than manually mapping it to raw_string, you should be able to use the Format attached to the table. See the Kafka source for an example.

jacksonrnewhouse (changes requested):

arroyo-connectors/src/s3.rs
rcjmurillo: Got it, will take a look 👍🏼

arroyo-worker/src/connectors/filesystem/s3/source.rs
rcjmurillo: cool, will update it.

jacksonrnewhouse:
Looks like you need to run `cargo fmt` and merge in master.

jacksonrnewhouse:
Found some time to review this and resolve conflicts. I found merging in commits from master was growing increasingly challenging, so I rebased on top of current HEAD and pushed to https://github.com/ArroyoSystems/arroyo/tree/jnewhouse/s3-conn-source. It looks good overall, I'm going to make a few additional changes, starting with using paths for the location and the generic object_store API.

rcjmurillo:
@jacksonrnewhouse thank you! definitely not easy to keep up to date with the fast progress on `master` 😄 , i was going to ask for another review but i'm having issues with the UI and it seems the it doesn't support more then 2 levels of nesting for objects from the connection schema, the UI won't respond after selecting the 2nd nested option, and also based on what saw in `arroyo-console/src/routes/connections/JsonForm.tsx`, maybe this is incorrect. Not sure what other approach could be taken for that. 
But this does work to use it as a source directly in SQL definitions.

rcjmurillo:
Closing in favor of #403.

PR #328, opened 2023-09-20
Allow compaction on previously compacted epochs

jbeisen:
If the controller is restarted immediately after a compaction it will try to compact the same epochs as its last compaction because the latest checkpoint specifies the min epoch, and that hasn't yet been updated. This allows compaction to continue even if the files it tries to delete don't exist.

### DIFF
--- a/arroyo-state/src/parquet.rs
+++ b/arroyo-state/src/parquet.rs
@@ -265,7 +265,7 @@ impl BackingStore for ParquetBackend {
                     &operator_id,
                 ));
                 debug!("deleting {}", path);
-                storage_client.lock().await.delete(path).await?;
+                storage_client.lock().await.delete_if_present(path).await?;
             }
             debug!(
                 message = "Finished compacting operator",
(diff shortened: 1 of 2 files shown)


arroyo-state/src/parquet.rs
```diff
@@ -265,7 +265,9 @@ impl BackingStore for ParquetBackend {
                     &operator_id,
                 ));
                 debug!("deleting {}", path);
-                storage_client.lock().await.delete(path).await?;
+                if let Err(e) = storage_client.lock().await.delete(path).await {
```
jacksonrnewhouse, line 268: Do we wanna check the error to make sure it is the "not present" error?

jbeisen: Like this?
```
match storage_client.lock().await.delete(path).await {
    Err(StorageError::ObjectStore(object_store::Error::NotFound {
        path, ..
    })) => {
        warn!("Failed to delete file: {}", path);
    }
    Err(e) => return Err(anyhow!("Failed to delete file: {}", e)),
    _ => {}
}
```

We might want to put that into `StorageProvider::delete` instead.

jacksonrnewhouse: Yeah, given that S3 will never fail on a missing file, maybe it makes sense to adjust the object store API. The object_store docs say 
> If the object did not exist, the result may be an error or a success, depending on the behavior of the underlying store. For example, local filesystems, GCP, and Azure return an error, while S3 and in-memory will return Ok. If it is an error, it will be [`Error::NotFound`].

jbeisen: Makes sense. I but that logic in the storage provider and renamed the function to `delete_if_present`

jacksonrnewhouse (approved):

PR #330, opened 2023-09-22
Trigger checkpoints and restoring in smoke tests

jbeisen:
Modify the single file source to block on receiving control messages, which enables the smoke tests to deterministically trigger checkpoints, and use a Noop control message to advance the source.

Running successive engines in the same process required changing how Prometheus metric registration works because metrics cannot be registered multiple times. This changes the worker to statically register counter vectors and apply labels when the counter is used.

This also splits up the job controller state into separate modules.

A later commit will add state compaction to the smoke tests.


arroyo-controller/src/job_controller/checkpoint_state.rs
jbeisen, line 1: This is all just moved from `job_controller/checkpointer.rs`.

arroyo-sql-testing/src/smoke_tests.rs
```diff
@@ -21,14 +121,56 @@ async fn run_pipeline_and_assert_outputs(
...
+    let engine = Engine::for_local(program, job_id.clone());
+    let (running_engine, mut control_rx) = engine
         .start(StreamConfig {
             restore_epoch: None,
         })
         .await;
-    // wait for the engine to finish
-    while let Some(control_resp) = control_rx.recv().await {}
+
+    if checkpoints {
```
jacksonrnewhouse, line 133: Can this be factored out into a method?

arroyo-sql-macro/src/lib.rs
```diff
@@ -209,6 +209,7 @@ impl Parse for PipelineCase {
 
 struct CorrectnessTestCase {
     test_name: LitStr,
+    checkpoints: LitStr,
```
jacksonrnewhouse, line 212: You can make this a `syn::LitBool` and then your macros can just take `true` and `false`.

arroyo-worker/src/engine.rs
```diff
@@ -149,33 +148,30 @@ impl OutQueue {
...
             let bytes = bincode::encode_to_vec(&message, config::standard()).unwrap();
-            sent_bytes.iter().for_each(|c| c.inc_by(bytes.len() as u64));
+            TaskCounters::BytesSent
+                .for_task(task_info)
+                .inc_by(bytes.len() as u64);
 
             QueueItem::Bytes(bytes)
         } else {
-            QueueItem::Data(Box::new(message))
+            QueueItem::Data(Box::new(message.clone()))
```
jacksonrnewhouse, line 161: This clone should be avoided.

arroyo-sql-testing/src/smoke_tests.rs
jbeisen: Done

arroyo-sql-macro/src/lib.rs
jbeisen: Done

arroyo-worker/src/connectors/websocket.rs
```diff
@@ -115,6 +115,7 @@ where
             ControlMessage::LoadCompacted { compacted } => {
                 ctx.load_compacted(compacted).await;
             }
+            ControlMessage::Noop => {}
```
jacksonrnewhouse, line 118: I think this would be a NoOp typically

jbeisen: Changed

arroyo-controller/src/job_controller/checkpoint_state.rs
```diff
@@ -0,0 +1,396 @@
...
+            start_time: to_micros(self.start_time),
+            finish_time: to_micros(finish_time),
+            min_epoch: self.min_epoch,
+            operator_ids: self.completed_operators.iter().cloned().collect(),
+        })
+        .await;
+
+        let operator_state = serde_json::to_value(&self.operator_details).unwrap();
+
+        if let Some(pool) = pool {
```
jacksonrnewhouse, line 355: This seems a bit of an anti-pattern. This is now two methods "finish saving operator state" and "update DB with new checkpoint status" and depending on how it is called you only do the second sometimes. Can you try factoring it out into two methods?

arroyo-sql-testing/Cargo.toml
```diff
@@ -27,3 +28,7 @@ tokio = { version = "1.16", features = ["full"] }
 arrow = {workspace = true }
 arrow-array = {workspace = true }
 arrow-schema = {workspace = true }
+tracing = "0.1.37"
+test-log = { version = "0.2.12", default-features = false, features = ["trace"] }
+tracing-subscriber = {version = "0.3.17"}
+#env_logger = "0.10.0"
```
jacksonrnewhouse, line 34: delete

arroyo-controller/src/job_controller/checkpoint_state.rs
```diff
@@ -0,0 +1,396 @@
...
+                .any(|table| table.write_behavior() == TableWriteBehavior::CommitWrites)
+        {
+            self.subtasks_to_commit
+                .insert((c.operator_id.clone(), metadata.subtask_index));
+        }
+
+        let subtasks = self
+            .tasks
+            .get_mut(&c.operator_id)
+            .unwrap_or_else(|| panic!("Received finish event without start for {}", c.operator_id));
```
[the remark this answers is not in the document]
jbeisen: Sure, done

arroyo-controller/src/job_controller/checkpoint_state.rs
jbeisen: Good idea, I've split this into two functions and changed `finish_checkpoint_if_done` to use them.

jacksonrnewhouse (approved):

PR #331, opened 2023-09-26
Making the raw identifier robust for numeric aliases

### LINKED ISSUE #286: Codegen panics when SQL aliases are not valid Rust identifiers
For example, this query:

```sql
SELECT 'a' as "1"
from nexmark;
```

produces a panic:

```
ERROR arroyo_server_common: panicked at 'Ident cannot be a number; use Literal instead', /root/.cargo/registry/src/index.crates.io-6f17d22bba15001f/proc-macro2-1.0.66/src/fallback.rs:756:9 panic.file="/root/.cargo/registry/src/index.crates.io-6f17d22bba15001f/proc-macro2-1.0.66/src/fallback.rs" panic.line=756 panic.column=9
```

edmondop:
Closes #286

### DIFF
--- a/arroyo-sql-testing/src/full_query_tests.rs
+++ b/arroyo-sql-testing/src/full_query_tests.rs
@@ -436,3 +436,9 @@ full_pipeline_codegen! {"polling_http_source",
 SELECT value
 FROM polling_source;
 "}
+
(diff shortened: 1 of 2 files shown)


jacksonrnewhouse (approved):
Looks great!

PR #332, opened 2023-09-26
Smoke test compaction and fix compaction logic

jbeisen:
Change the existing smoke tests to perform a compaction and load the
compacted data. This uncovered a mistake in the compaction logic for
TimeKeyMap tables.

### DIFF
--- a/arroyo-state/src/tables/mod.rs
+++ b/arroyo-state/src/tables/mod.rs
@@ -44,7 +44,8 @@ impl Compactor {
                 // keep only the latest entry for each key
                 let mut reduced = HashMap::new();
                 for tuple in tuples.into_iter() {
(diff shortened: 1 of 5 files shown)


arroyo-state/src/tables/mod.rs
```diff
@@ -44,7 +44,8 @@ impl Compactor {
                 // keep only the latest entry for each key
                 let mut reduced = HashMap::new();
                 for tuple in tuples.into_iter() {
-                    reduced.insert(tuple.key.clone(), tuple);
+                    let memory_key = (tuple.timestamp, tuple.key.clone());
+                    reduced.insert(memory_key, tuple);
```
jbeisen, line 48: @jacksonrnewhouse this is the main compaction logic

jacksonrnewhouse (approved):

PR #338, opened 2023-09-29
Support delete operations for KeyTimeMultiMap

jbeisen:
Add DeleteKey, DeleteValue, and DeleteRange operations for KeyTimeMultiMap tables.

### DIFF
--- a/arroyo-state/src/lib.rs
+++ b/arroyo-state/src/lib.rs
@@ -77,11 +77,37 @@ pub fn key_time_multi_map_table(
 }
 
 #[derive(Debug, Encode, Decode, PartialEq, Eq, Clone)]
+pub struct DeleteTimeKeyOperation {
+    pub timestamp: SystemTime,
+    pub key: Vec<u8>,
+}
+
+#[derive(Debug, Encode, Decode, PartialEq, Eq, Clone)]
+pub struct DeleteKeyOperation {
+    pub key: Vec<u8>,
+}
+
+#[derive(Debug, Encode, Decode, PartialEq, Eq, Clone)]
+pub struct DeleteValueOperation {
+    pub key: Vec<u8>,
+    pub timestamp: SystemTime,
+    pub value: Vec<u8>,
+}
+
+#[derive(Debug, Encode, Decode, PartialEq, Eq, Clone)]
+pub struct DeleteTimeRangeOperation {
+    pub key: Vec<u8>,
+    pub start: u64,
+    pub end: u64,
+}
+
+#[derive(Debug, Encode, Decode, PartialEq, Eq, Clone)]
 pub enum DataOperation {
     Insert,
-    DeleteKey,                 // delete single key of a TimeKeyMap or KeyTimeMultiMap
-    DeleteValue,               // delete single value of a KeyTimeMultiMap
-    DeleteTimeRange(u64, u64), // delete all values for key in range (only for KeyTimeMultiMap)
+    DeleteTimeKey(DeleteTimeKeyOperation), // delete single key of a TimeKeyMap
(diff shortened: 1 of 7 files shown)


arroyo-state/src/tables/key_time_multi_map.rs
```diff
@@ -124,10 +151,24 @@ impl<K: Key, V: Data> KeyTimeMultiMapCache<K, V> {
...
-                    values
-                        .entry(tuple.key)
-                        .or_default()
-                        .remove(&tuple.timestamp);
+                    values.remove(&tuple.key);
+                }
+                DataOperation::DeleteValue => {
+                    values.entry(tuple.key).and_modify(|map| {
+                        map.entry(tuple.timestamp).and_modify(|values| {
+                            values.retain(|value| value != &tuple.value.clone().unwrap());
```
jacksonrnewhouse, line 159: In general I think value-deletes should only remove a single value, even if there are more matches. For instance, if my state is keyed by campaign_id and has user_id, you could imagine a table of visits like

| campaign_id | user_id|
| 1 | 2 |
| 1 | 2 |
| 1 | 3 |
| 1 | 4 |

In this case (delete((1,2))) should only retract a single row.

arroyo-state/src/tables/mod.rs
```diff
@@ -50,9 +51,67 @@ impl Compactor {
...
+                                .push(tuple);
+                        }
+                        DataOperation::DeleteKey => {
+                            // Remove a key. Timestamp is not considered.
+                            values.remove(tuple.key.as_slice());
+                            deletes.insert(keep_deletes_key, tuple);
+                        }
+                        DataOperation::DeleteValue => {
+                            // Remove a single value from a (key -> time -> values).
+                            values.entry(tuple.key.clone()).and_modify(|map| {
```
jacksonrnewhouse, line 80: Same here, deleting a single value should only remove one entry from values.

arroyo-state/src/lib.rs
```diff
@@ -76,12 +76,38 @@ pub fn key_time_multi_map_table(
...
+}
+
 #[derive(Debug, Encode, Decode, PartialEq, Eq, Clone)]
 pub enum DataOperation {
     Insert,
-    DeleteKey,                 // delete single key of a TimeKeyMap or KeyTimeMultiMap
-    DeleteValue,               // delete single value of a KeyTimeMultiMap
-    DeleteTimeRange(u64, u64), // delete all values for key in range (only for KeyTimeMultiMap)
+    DeleteTimeKey(DeleteTimeKeyOperation), // delete single key of a TimeKeyMap
+    DeleteKey(DeleteKeyOperation),         // delete single key of a KeyTimeMultiMap
```
jacksonrnewhouse, line 108: is this the right comment?

arroyo-state/src/parquet.rs
```diff
@@ -793,10 +830,14 @@ impl ParquetBackend {
...
                 };
 
-                let operation: DataOperation =
-                    bincode::decode_from_slice(operation_array.value(index), BINCODE_CONFIG)
-                        .unwrap()
-                        .0;
+                let operation: Result<(DataOperation, usize), DecodeError> =
+                    bincode::decode_from_slice(operation_array.value(index), BINCODE_CONFIG);
+                let operation = match operation {
+                    Ok(o) => o.0,
```
jacksonrnewhouse, line 836: nit: don't love single character variable names.

jacksonrnewhouse (approved):
looks good, just a couple comments.

arroyo-state/src/lib.rs
jbeisen: I've made it more clear

PR #339, opened 2023-10-01
Add support for framing in message deserialization

mwylde:
This PR adds message framing into our deserialization process.

Currently, a single message read from a source (kafka, websocket, etc.) can only produce a single record in Arroyo. This is mostly ok for sources that provide their own framing mechanism, but can be a limitation for sources that do not have a standard protocol (for example, http endpoints can only return a single value).

With this PR, users can now set a framing method on their source schema that determines how messages are broken up into records. In this PR there's an initial method "newline" which breaks messages on newlines using the very fast memchar library, but the framework is extensible for other framing methods as well (for example, length-delimited framing will be useful once we support binary formats like protobuf).

This PR updates all sources that read custom data with the new framing support (via a new DataDeserializer struct that produces an iter of messages) and adds a UI for setting the framing method when creating a source.

It also includes a small bug fix to exclude generated fields from source schemas, which was preventing the creation of a raw_string source with generated fields.

---

This is particularly useful when working with the polling http source, which does not otherwise have any way to frame messages. For example, we can now directly read from prometheus endpoints:

```sql
create table raw_metrics (
    value TEXT,
    parsed TEXT generated always as (parse_prom(value))
) WITH (
    connector = 'polling_http',
    endpoint = 'http://localhost:9100/metrics',
    format = 'raw_string',
    framing = 'newline',
    emit_behavior = 'changed',
    poll_interval_ms = '1000'
);
```

```rust
fn parse_prom(s: String) -> Option<String> {
    let regex = regex::Regex::new(r"(?P<metric_name>\w+)\{(?P<labels>[^}]+)\}\s+(?P<metric_value>[\d.]+)").unwrap();
    let label_regex = regex::Regex::new(r##"(?P<label>[^,]+)="(?P<value>[^"]+)""##).unwrap();
    
    let captures = regex.captures(&s)?;
    
    let name = captures.name("metric_name").unwrap().as_str();
    let labels = captures.name("labels").unwrap().as_str();
    let value = captures.name("metric_value").unwrap().as_str();
    
    let labels: std::collections::HashMap<String, String> = label_regex.captures_iter(&labels)
        .map(|capture| (
            capture.name("label").unwrap().as_str().to_string(), 
            capture.name("value").unwrap().as_str().to_string()
        ))
        .collect();
    
    
    Some(serde_json::json!({
        "name": name,
        "labels": labels,
        "value": value
    }).to_string())
}
```


arroyo-connectors/src/kafka.rs
```diff
@@ -95,6 +95,7 @@ impl Connector for KafkaConnector {
             table: serde_json::to_value(table).unwrap(),
             rate_limit: None,
             format: Some(format),
+            framing: schema.framing.clone(),
```
jacksonrnewhouse, line 98: Why do we need framing on both the OperatorConfig and the ConnectionSchema?

jacksonrnewhouse (approved):
looks good, just one question.

arroyo-connectors/src/kafka.rs
mwylde: The connection schema is what's used in the API and database, while the operator config is what is used to initialize the operator (i.e., what's passed from the compiler to the `from_config` method of the operators).

PR #344, opened 2023-10-02
writing to file:/// should write to absolute paths.

jacksonrnewhouse:
Apparently converting `"file:///my/absolute/path"` to an ObjectStore Path will result in `my/absolute/path`. I've reworked how the path is managed so that it writes to an absolute location.

### DIFF
--- a/arroyo-connectors/src/filesystem.rs
+++ b/arroyo-connectors/src/filesystem.rs
@@ -135,7 +135,14 @@ impl Connector for FileSystemConnector {
         schema: Option<&ConnectionSchema>,
     ) -> anyhow::Result<crate::Connection> {
         let write_target = if let Some(path) = opts.remove("path") {
-            Destination::FolderUri { path }
+            if path.starts_with("file:///") {
+                let trimmed_path = path.trim_start_matches("file://");
+                Destination::LocalFilesystem {
+                    local_directory: trimmed_path.to_string(),
+                }
+            } else {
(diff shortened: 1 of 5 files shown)


arroyo-connectors/src/filesystem.rs
```diff
@@ -135,7 +136,13 @@ impl Connector for FileSystemConnector {
         schema: Option<&ConnectionSchema>,
     ) -> anyhow::Result<crate::Connection> {
         let write_target = if let Some(path) = opts.remove("path") {
-            Destination::FolderUri { path }
+            if let BackendConfig::Local(local_config) = BackendConfig::parse_url(&path, false)? {
+                Destination::LocalFilesystem {
+                    local_directory: local_config.path,
+                }
+            } else {
+                Destination::FolderUri { path }
```
mwylde, line 144: This should probably produce an error since this means that it's an s3/gcs path (or it would be straightforward to just configure an s3 bucket from that)

jacksonrnewhouse: We currently support s3 URIs via the FolderUri variant. This delegates parsing to ObjectStore, so this should fall through correctly. I've opened #345 as a fuller migration to arroyo-storage.

mwylde: I see, I was confused bythe FolderUri name that seems to imply a local filesystem

mwylde (approved):

PR #349, opened 2023-10-05
Remove arroyo-sql-testing's dependency on the controller

jbeisen:
This required moving the CheckpointState, CommittingState, and SubtaskState to the arroyo-state crate, and untangling those structs from the database queries performed by the controller.

### DIFF
--- a/arroyo-state/src/committing_state.rs
+++ b/arroyo-state/src/committing_state.rs
@@ -1,7 +1,4 @@
-use crate::queries::controller_queries;
-use deadpool_postgres::Pool;
 use std::collections::HashSet;
-use std::time::SystemTime;
 
 pub struct CommittingState {
     checkpoint_id: i64,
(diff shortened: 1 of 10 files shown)


arroyo-state/src/committing_state.rs
```diff
@@ -1,10 +1,7 @@
-use crate::queries::controller_queries;
-use deadpool_postgres::Pool;
 use std::collections::HashSet;
-use std::time::SystemTime;
 
 pub struct CommittingState {
-    checkpoint_id: i64,
+    pub checkpoint_id: i64,
```
jacksonrnewhouse, line 4: rather than making these pub, can you just make an accessor?

jacksonrnewhouse (approved):

arroyo-state/src/committing_state.rs
jbeisen: sure, done

PR #352, opened 2023-10-09
connectors: implement event time and field based partitioning for filesystem sink.

jacksonrnewhouse:
This change implements two kinds of partitioning for outputs to the filesystem sink: event time partitioning and field based partitioning. The semantics are as follows:

### Event Time Partitioning
This partitioning type takes in a time formatting string and appends it to the overall directory. The formatting strings are of strftime style, in particular using [this](https://docs.rs/chrono/latest/chrono/struct.DateTime.html#method.format) method. Rollover is handled by tracking the watermark of the operator and introducing a new WatermarkExpiration to RollingPolicy. In particular, if the timestamp is greater than `representative_timestamp` _and_ the watermark has a different formatted string, then it is presumed that the prior bucket is expired. It will be closed and flushed.

### Field-based partitioning
A list of fields can be provided. These will be added to the file location using Hive-style partition directories. Expiration will be the same as unpartitioned sinks, unless it is combined with an event time partitioning.

Example:

```sql
CREATE TABLE file_sink (
    counter bigint unsigned,
    counter_mod_ten bigint unsigned
) WITH (
    connector = 'filesystem',
    format = 'json',
    aws_region = 'us-west-2',
    s3_bucket = 'arroyo-sink',
    s3_directory = 'partitioning',
    rollover_seconds = '500000',
    time_partition_pattern = 'year=%Y/month=%m/day=%d/hour=%H/minute=%M',
    partition_fields = 'counter_mod_ten'
);
INSERT INTO file_sink SELECT counter, counter%10 from slow_infinite_impulse;
```

### DIFF
--- a/arroyo-connectors/src/filesystem.rs
+++ b/arroyo-connectors/src/filesystem.rs
@@ -78,7 +78,10 @@ impl Connector for FileSystemConnector {
         schema: Option<&ConnectionSchema>,
     ) -> anyhow::Result<crate::Connection> {
         let is_local = match &table.write_target {
-            Destination::FolderUri { path } => path.starts_with("file:/"),
+            Destination::FolderUri {
+                path,
+                storage_options: _,
+            } => path.starts_with("file:/"),
             Destination::S3Bucket { .. } => false,
             Destination::LocalFilesystem { .. } => true,
         };
@@ -138,12 +141,15 @@ impl Connector for FileSystemConnector {
         schema: Option<&ConnectionSchema>,
     ) -> anyhow::Result<crate::Connection> {
         let write_target = if let Some(path) = opts.remove("path") {
-            if let BackendConfig::Local(local_config) = BackendConfig::parse_url(&path, false)? {
-                Destination::LocalFilesystem {
-                    local_directory: local_config.path,
-                }
-            } else {
-                Destination::FolderUri { path }
+            let storage_options: std::collections::HashMap<String, String> = opts
+                .iter()
+                .filter(|(k, _)| k.starts_with("storage."))
+                .map(|(k, v)| (k.trim_start_matches("storage.").to_string(), v.to_string()))
+                .collect();
+            opts.retain(|k, _| !k.starts_with("storage."));
+            Destination::FolderUri {
+                path,
+                storage_options,
             }
         } else if let (Some(s3_bucket), Some(s3_directory), Some(aws_region)) = (
             opts.remove("s3_bucket"),
--- a/arroyo-connectors/src/filesystem.rs
+++ b/arroyo-connectors/src/filesystem.rs
@@ -9,7 +9,7 @@ use arroyo_rpc::types::{ConnectionSchema, ConnectionType, TestSourceMessage};
 use arroyo_rpc::OperatorConfig;
 use serde::{Deserialize, Serialize};
 
-use crate::{pull_option_to_i64, Connection, EmptyConfig};
+use crate::{pull_opt, pull_option_to_i64, Connection, EmptyConfig};
 
 use super::Connector;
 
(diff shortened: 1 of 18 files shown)


arroyo-connectors/src/filesystem.rs
```diff
@@ -165,18 +165,38 @@ impl Connector for FileSystemConnector {
...
             target_part_size,
+            partitioning,
         });
+
         let format_settings = match schema
             .ok_or(anyhow!("require schema"))?
             .format
             .as_ref()
-            .unwrap()
+            .ok_or(anyhow!("require format"))?
```
mwylde, line 199: can we make this error more understandable for users? what would they need to fix to get rid of it?

arroyo-worker/src/connectors/filesystem/mod.rs
```diff
@@ -128,24 +130,79 @@ impl<K: Key, T: Data + Sync, R: MultiPartWriter<InputType = T> + Send + 'static>
...
+
+fn formatted_time_from_timestamp(timestamp: SystemTime, pattern: &str) -> String {
+    let datetime: DateTime<Utc> = DateTime::from(timestamp);
+    datetime.format(pattern).to_string()
+}
+fn partition_string_for_fields<T: Data + Serialize>(
+    value: &T,
+    partition_fields: &Vec<String>,
+) -> Result<String> {
+    let value = serde_json::to_value(value)?;
```
mwylde, line 187: this shouldn't fail (it would indicate a bug in codegen), so I would unwrap instead of propagating it

arroyo-connectors/src/filesystem.rs
```diff
@@ -165,18 +165,38 @@ impl Connector for FileSystemConnector {
         let target_file_size = pull_option_to_i64("target_file_size", opts)?;
         let target_part_size = pull_option_to_i64("target_part_size", opts)?;
 
+        let formatting_string = opts.remove("time_partition_pattern");
+        let time_partition_pattern =
```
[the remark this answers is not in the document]
jacksonrnewhouse: I think every string is a potentially valid format. What did you have in mind?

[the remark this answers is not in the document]
mwylde: Ah, that make sense

connector-schemas/filesystem/table.json
```diff
@@ -129,6 +129,36 @@
...
+                },
+                "partitioning": {
+                    "title": "Partitioning",
+                    "type": "object",
+                    "properties": {
+                        "time_partition_pattern": {
+                            "title": "Time Partition Pattern",
+                            "type": "object",
+                            "properties": {
+                                "pattern": {
```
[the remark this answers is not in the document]
jacksonrnewhouse: fixed.

k8s/arroyo/values.yaml
```diff
@@ -15,7 +15,7 @@ api:
   image:
     repository: ghcr.io/arroyosystems/arroyo-services
     pullPolicy: IfNotPresent
-    tag: "0.6.0"
+    tag: "arm64-rc-0.7.0-docker"
```
mwylde, line 18: Did you mean to commit this change?

docker/build_base/Cargo.toml
```diff
@@ -17,9 +17,9 @@ arrow-schema = { version = "43.0.0" }
 parquet = { version = "43.0.0" }
 
 [patch.crates-io]
-parquet = {git = 'https://github.com/ArroyoSystems/arrow-rs', branch = '43.0.0/arroyo_patches'}
-arrow = {git = 'https://github.com/ArroyoSystems/arrow-rs', branch = '43.0.0/arroyo_patches'}
-arrow-buffer = {git = 'https://github.com/ArroyoSystems/arrow-rs', branch = '43.0.0/arroyo_patches'}
-arrow-array = {git = 'https://github.com/ArroyoSystems/arrow-rs', branch = '43.0.0/arroyo_patches'}
-arrow-schema = {git = 'https://github.com/ArroyoSystems/arrow-rs', branch = '43.0.0/arroyo_patches'}
-object_store = {git = 'https://github.com/ArroyoSystems/arrow-rs', branch = '43.0.0/arroyo_patches'}
+parquet = {git = 'https://github.com/ArroyoSystems/arrow-rs', branch = '43.0.0/arroyo_patches-1'}
```
mwylde, line 20: is this something that should be merged in before committing?

jacksonrnewhouse: I cut to a different branch to avoid caching issues. Is there a better approach?

mwylde: You mean lockfile issues?

I don't think that will affect users, since they have our lockfiles. We may want to start checking in the build_dir lockfile for us.

mwylde (approved):

k8s/arroyo/values.yaml
jacksonrnewhouse: no

PR #353, opened 2023-10-09
Add session window smoke tests

jbeisen:


arroyo-worker/src/operators/windows.rs
```diff
@@ -425,6 +425,14 @@ impl<K: Key, T: Data, OutT: Data> SessionWindowFunc<K, T, OutT> {
             .clear_time_range(&mut key, window.start, window.end)
             .await;
     }
+
+    async fn handle_checkpoint(
```
jbeisen, line 429: Under what conditions does an operator need to define `handle_checkpoint` itself? Should we remove the empty one in the macro?

arroyo-worker/src/operators/windows.rs
```diff
@@ -324,15 +324,15 @@ impl<K: Key, T: Data, OutT: Data> SessionWindowFunc<K, T, OutT> {
...
+                retention_micros: MAX_SESSION_SIZE.as_micros() as u64,
             },
             TableDescriptor {
                 name: "s".to_string(),
                 description: "sessions".to_string(),
                 table_type: TableType::TimeKeyMap as i32,
                 delete_behavior: TableDeleteBehavior::NoReadsBeforeWatermark as i32,
                 write_behavior: TableWriteBehavior::NoWritesBeforeWatermark as i32,
-                retention_micros: MAX_SESSION_SIZE.as_micros() as u64,
+                retention_micros: 0 as u64,
```
jbeisen, line 335: Why not `MAX_SESSION_SIZE`?

arroyo-worker/src/operators/windows.rs
jacksonrnewhouse: If it is using table types that buffer data, it needs to flush those. In the state store changes I have planned there's a change that pushes this back into the state store itself, which will be nice.

arroyo-worker/src/operators/windows.rs
jacksonrnewhouse: The timestamp is the end time of the final session, so we don't need to keep any data before the watermark, as it'll have already been processed.

jacksonrnewhouse (approved):
Looks good!

PR #354, opened 2023-10-09
Implement unnest

mwylde:
This PR adds initial support for `unnest`, which unrolls an array into individual records. This is often useful when interacting with HTTP apis via the polling http source, as it allows json arrays to be handled as individual records.

For example, emitting on new Github PRs:

```sql
CREATE TABLE prs (
    value TEXT
) WITH (
    connector = 'polling_http',
    endpoint = 'https://api.github.com/repos/ArroyoSystems/arroyo/pulls?per_page=50&state=all',
    poll_interval_ms = '5000',
    emit_behavior = 'changed',
    headers = 'User-Agent:arroyo/0.6',
    format = 'json',
    'json.unstructured' = 'true'
);
 
SELECT distinct(unnest(extract_json(value, '$[*].url'))) as url
from prs;
```

image

Unnest is currently only implemented for projections. The implementation works by registering unnest as a UDF with datafusion, then looking for it in projection expressions. If it's found, the expression is split into two parts around the unnest. This is then turned into an UnnestProjection, which compiles into a flatmap that maps over the array produced by the inner expression and projects it uses the outer expression, and finally collects it into a vec. This vec is then unrolled by a new FlatMapOperator (the existing, misnamed FlatMapOperator has been renamed to ArrayMapOperator).

This PR also includes a UX improvement to json functions, which now will validate the json path expression at plan time if it's a literal string (previously bad json paths were just ignored leading to a confusing user experience). I've also added a couple of aliases for the json functions to better match other sql dialects.

### DIFF
--- a/arroyo-sql/src/pipeline.rs
+++ b/arroyo-sql/src/pipeline.rs
@@ -27,6 +27,7 @@ use crate::{
     types::{interval_month_day_nanos_to_duration, StructDef, StructField, TypeDef},
     ArroyoSchemaProvider,
 };
+use crate::operators::UnnestProjection;
 
 #[derive(Debug, Clone)]
 pub enum SqlOperator {
@@ -44,6 +45,7 @@ pub enum SqlOperator {
 pub enum RecordTransform {
     ValueProjection(Projection),
     KeyProjection(Projection),
+    UnnestProjection(UnnestProjection),
     TimestampAssignment(Expression),
     Filter(Expression),
 }
@@ -75,6 +77,9 @@ impl RecordTransform {
             }
             RecordTransform::KeyProjection(_) | RecordTransform::Filter(_) => input_struct,
             RecordTransform::TimestampAssignment(_) => input_struct,
+            RecordTransform::UnnestProjection(projection) => {
+                projection.expression_type(&ValuePointerContext::new())
+            }
         }
     }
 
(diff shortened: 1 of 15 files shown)


arroyo-sql/src/pipeline.rs
```diff
@@ -473,13 +488,50 @@ impl<'a> SqlPipelineBuilder<'a> {
         let input = self.insert_sql_plan(&filter.input)?;
         let struct_def = input.return_type();
         let ctx = self.ctx(&struct_def);
-        let predicate = ctx.compile_expr(&filter.predicate)?;
+        let mut predicate = ctx.compile_expr(&filter.predicate)?;
+
+        Self::assert_no_unnest("where", &mut predicate)?;
```
jacksonrnewhouse, line 493: Would this be clearer with something like
```
if Self::has_unnest(&mut predicate) {
    bail!("can't have unnest in WHERE clause");
}

jacksonrnewhouse (approved):
Looks great!

arroyo-sql/src/pipeline.rs
mwylde: given this appears in a bunch of places, I preferred to move the bail call into the function so that I wasn't repeating it over and over again. To me at least, having `assert` and the `?` makes it clear that this can early-exit.

PR #359, opened 2023-10-10
Run cargo check on udfs

jbeisen:
Add an API endpoint that accepts the contents of the UDFs editor and
sends it to the controller, and in turn to the compiler service, which
runs 'cargo check'. Pass any errors back to the API and show them in the
console. Also show the query validation errors in the Errors tab for
consistency.

Note that the UDFs are put into a new 'udfs' crate in the build
directory for the purpose of cargo check, and written again the
'pipeline' crate for the actual pipeline. A future commit will change
the pipeline to use the udfs crate.

---

Query error:

image

UDF error:

image


arroyo-controller/src/lib.rs
```diff
@@ -427,6 +434,64 @@ impl ControllerGrpc for ControllerServer {
...
+
+fn get_udfs_mod(udfs: &Vec<String>) -> Result<String, Status> {
+    let imports = quote! {
+        #![allow(warnings)]
+        use types::*;
+        use chrono;
+        use std::time::SystemTime;
+        use std::str::FromStr;
+        use serde::{Deserialize, Serialize};
+    };
```
jbeisen, line 473: @mwylde I looked at the imports in `compile_pipeline_main` and kind of guessed what we want for udfs. What do you think?

arroyo-api/src/pipelines.rs
```diff
@@ -339,35 +348,82 @@ pub async fn validate_pipeline(
...
+            errors: Some(check_udfs_resp.errors),
+        }
+    } else {
+        UdfsValidationResult {
+            udfs: Some(validate_pipeline_post.udfs.unwrap_or(vec![])),
+            errors: None,
+        }
+    };
+
+    let pipeline_graph_validation_result = match compile_sql(&sql, &auth_data, &client).await {
```
mwylde, line 383: Does it make sense to validate the sql first, so that we can report errors back more quickly?

arroyo-api/src/pipelines.rs
```diff
@@ -98,11 +108,7 @@ where
         },
     )
     .await
-    .with_context(|| "failed to generate SQL program")
-    .map_err(|err| {
-        warn!("{:?}", err);
-        bad_request(format!("{}", err.root_cause()))
-    })?;
+    .with_context(|| "failed to generate SQL program")?;
```
mwylde, line 111: this loses the detailed sql errors. 

For example, `select * from blah;` which on master tells me the actual error (blah doesn't exist) now just returns "failed to generate SQL program"

arroyo-compiler-service/src/main.rs
```diff
@@ -261,4 +263,54 @@ impl CompilerGrpc for CompileService {
                 .unwrap_or_else(|e| Status::internal(e.to_string()))
         })
     }
+
+    async fn check_udfs_mod(
+        &self,
+        request: Request<CheckUdfsModReq>,
+    ) -> Result<Response<CheckUdfsResp>, Status> {
+        // write udf to build_dir udfs module
+        let build_dir = &self.build_dir;
```
mwylde, line 272: this could use some logging so it's clear what the compiler service is doing (and timing information is also helpful)

mwylde (changes requested):
This is a good start. I think it needs a couple more iterations before being ready to merge.

* We no longer get useful SQL error messages, which is a big degradation in the UX
* I think we can avoid slowing down SQL checking when there are no UDFs (probably the most common case)
* The UDF errors are often too vague to be useful (like "mismatched type" — which types? where?)
* This will need changes in the docker images so this can work outside of our dev environment — I would specifically build a docker-single image and make sure that works
* It might make sense to have udf checking be a separate API call from graph validation, which would allow caching of UDF results (that way you wouldn't need to recompile the udfs every time, even when they haven't changed)

arroyo-rpc/src/types.rs
```diff
@@ -13,6 +13,27 @@ pub struct ValidatePipelinePost {
...
+#[serde(rename_all = "camelCase")]
+pub struct PipelineGraphValidationResult {
+    pub graph: Option<PipelineGraph>,
+    pub errors: Option<Vec<String>>,
+}
+
+#[derive(Serialize, Deserialize, Clone, Debug, ToSchema)]
+#[serde(rename_all = "camelCase")]
+pub struct UdfsValidationResult {
+    pub udfs: Option<Vec<Udf>>,
```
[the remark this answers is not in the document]
jbeisen: The logic here is that `udfs` (and `PipelineGraphValidationResult::graph`) are only Some if the validation was successful and there are no errors, i.e. only one of these fields is Some. I originally had this as an enum like this:

```
pub enum UdfsValidationResult {
    Ok(Vec<Udf>),
    Error(Vec<String>),
}
```

but this creates a bit of a weird API json schema, since there isn't really a standard way of representing in json the Rust style of enums variants with values.

arroyo-rpc/src/types.rs
```diff
@@ -13,6 +13,27 @@ pub struct ValidatePipelinePost {
     pub udfs: Option<Vec<Udf>>,
 }
 
+#[derive(Serialize, Deserialize, Clone, Debug, ToSchema)]
+#[serde(rename_all = "camelCase")]
+pub struct PipelineGraphValidationResult {
+    pub graph: Option<PipelineGraph>,
```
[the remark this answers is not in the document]
jbeisen: See above

arroyo-rpc/proto/rpc.proto
```diff
@@ -425,6 +430,21 @@ message CompileQueryResp {
...
+}
+
+message CheckUdfsResp {
+  ValidationResult result = 1;
+  repeated string errors = 2;
+}
+
 service CompilerGrpc {
   rpc CompileQuery(CompileQueryReq) returns (CompileQueryResp);
+  rpc CheckUdfsMod(CheckUdfsModReq) returns (CheckUdfsResp);
```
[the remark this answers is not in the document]
jbeisen: The string that the controller sends to the compiler service is the pretty printed source code that goes in the udfs _module's_ lib.rs - 
```
prettyplease::unparse(&parse_quote! {
    #imports
    
    #(#udf_tokens )*
}
```

so I named it like this to differentiate it from the `CheckUdfsReq` that the api sends to the controller, which is just the udfs strings.

arroyo-api/src/pipelines.rs
```diff
@@ -339,35 +348,82 @@ pub async fn validate_pipeline(
...
+
+    let check_udfs_resp = controller
+        .check_udfs(CheckUdfsReq { udfs: udf_strings })
+        .await
+        .map_err(log_and_map)?
+        .into_inner();
+
+    let udf_validation_result = if check_udfs_resp.result == ValidationResult::Error as i32 {
+        UdfsValidationResult {
+            udfs: None,
```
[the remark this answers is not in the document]
jbeisen: Exactly one of these two fields will be Some, the other None, so it behaves like an enum. I think this format translates better to JSON for the api.

arroyo-api/src/pipelines.rs
```diff
@@ -57,15 +60,24 @@ where
             Some(UdfLanguage::Rust) => {
                 schema_provider
                     .add_rust_udf(&udf.definition)
-                    .map_err(|e| bad_request(format!("Could not process UDF: {:?}", e)))?;
+                    .map_err(|e| anyhow!(format!("Could not process UDF: {:?}", e)))?;
```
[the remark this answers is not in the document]
jbeisen: This function no longer uses those functions, but they're used elsewhere in the API, where I think they're useful because they give us some error message consistency.

arroyo-api/src/pipelines.rs
```diff
@@ -339,35 +348,82 @@ pub async fn validate_pipeline(
...
-        .graph
-        .node_weights()
-        .map(|node| PipelineNode {
-            node_id: node.operator_id.to_string(),
-            operator: format!("{:?}", node),
-            parallelism: node.clone().parallelism as u32,
-        })
+    let udf_strings = validate_pipeline_post
+        .udfs
+        .clone()
```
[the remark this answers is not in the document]
jbeisen: Removed.

arroyo-api/src/pipelines.rs
```diff
@@ -339,35 +348,82 @@ pub async fn validate_pipeline(
         preview: false,
     };
 
-    let (mut program, _) = compile_sql(&sql, &auth_data, &client).await?;
-    optimizations::optimize(&mut program.graph);
+    // check the udfs
+    let mut controller = ControllerGrpcClient::connect(state.controller_addr.clone())
+        .await
+        .map_err(log_and_map)?;
```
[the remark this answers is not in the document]
jbeisen: Yup, makes sense. I've cleaned this so that the controller and compiler service both return appropriate `Status` responses and the API logs warnings.

arroyo-api/src/pipelines.rs
```diff
@@ -339,35 +348,82 @@ pub async fn validate_pipeline(
...
-            let src = program.graph.node_weight(edge.source()).unwrap();
-            let target = program.graph.node_weight(edge.target()).unwrap();
-            PipelineEdge {
-                src_id: src.operator_id.to_string(),
-                dest_id: target.operator_id.to_string(),
-                key_type: edge.weight().key.to_string(),
-                value_type: edge.weight().value.to_string(),
-                edge_type: format!("{:?}", edge.weight().typ),
+
+    let check_udfs_resp = controller
```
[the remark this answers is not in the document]
jbeisen: Since it's now two separate endpoints, this endpoint will only be called if UDFs are defined.

build_dir/udfs/Cargo.toml
```diff
@@ -0,0 +1,23 @@
+[package]
```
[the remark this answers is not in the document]
jbeisen: Done.

arroyo-rpc/proto/rpc.proto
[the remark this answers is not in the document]
jbeisen: I've renamed this to `CheckUdfs` since it now uses the same `CheckUdfsReq` as the controller.

arroyo-api/src/pipelines.rs
jbeisen: Yup, these are now separate endpoints.

arroyo-compiler-service/src/main.rs
jbeisen: Sure, I've added some logging. Are you suggesting something like timing how long `cargo check` takes?

arroyo-api/src/pipelines.rs
jbeisen: Fixed

arroyo-controller/src/lib.rs
jbeisen: I've changed it so the exact contents of the UDFs editor are written to the build directory, so it's now up to the user to write whatever `use` statements they need.

arroyo-rpc/src/types.rs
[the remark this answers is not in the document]
mwylde: That still doesn't really make sense to me as an api consumer. Why are we returning the udfs at all?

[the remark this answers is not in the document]
mwylde: I guess the weird thing is that we're using rust-style errors for an http endpoint, where you'd expect an error response in the error case.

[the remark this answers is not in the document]
mwylde: Anyways I think this is fine for now but we should probably think about how to more naturally express this in REST for the next version

arroyo-compiler-service/src/main.rs
mwylde: Yes, like the existing timing information we have for compilation: https://github.com/ArroyoSystems/arroyo/blob/master/arroyo-compiler-service/src/main.rs#L208-L210

build_dir/udfs/Cargo.toml
```diff
@@ -0,0 +1,23 @@
...
+
+[dependencies]
+types = { path = "../types" }
+petgraph = "0.6"
+bincode = "=2.0.0-rc.3"
+bincode_derive = "=2.0.0-rc.3"
+chrono = "0.4"
+serde = "1.0"
+serde_json = "1.0"
+arrow = { workspace = true}
```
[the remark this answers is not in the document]
mwylde: This doesn't seem to have been done — I'm still seeing all of those dependencies in the Cargo.toml

arroyo-api/src/pipelines.rs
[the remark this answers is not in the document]
mwylde: I don't quite follow what you're saying. I mean, previously we were able to distinguish between user errors (like invalid udf language) and system errors (like we failed to talk to the database), and now those all get collapsed into anyhow::error.

arroyo-api/src/pipelines.rs
```diff
@@ -306,29 +313,27 @@ impl Into<Job> for DbPipelineJob {
 /// Get a pipeline graph
 #[utoipa::path(
     post,
-    path = "/v1/pipelines/validate",
+    path = "/v1/pipelines/validate_query",
```
mwylde, line 316: maybe leave this as `validate` to avoid unnecessary changes to the API?

arroyo-api/src/pipelines.rs
```diff
@@ -339,35 +344,112 @@ pub async fn validate_pipeline(
...
+    responses(
+        (status = 200, description = "Validated query", body = UdfValidationResult),
+    ),
+)]
+pub async fn validate_udfs(
+    State(state): State<AppState>,
+    bearer_auth: BearerAuth,
+    WithRejection(Json(validate_udfs_post), _): WithRejection<Json<ValidateUdfsPost>, ApiError>,
+) -> Result<Json<UdfValidationResult>, ErrorResp> {
+    let _client = client(&state.pool).await?;
```
mwylde, line 405: shuold this be removed?

arroyo-compiler-service/src/main.rs
```diff
@@ -261,4 +270,70 @@ impl CompilerGrpc for CompileService {
...
+            &request.into_inner().udfs_rs,
+        )
+        .await
+        .unwrap();
+
+        let output = Command::new("cargo")
+            .current_dir(&self.build_dir)
+            .arg("check")
+            .arg("--package=udfs")
+            .arg("--message-format=json")
```
mwylde, line 293: ooh nice

arroyo-console/package.json
```diff
@@ -33,6 +33,7 @@
     "@types/react-syntax-highlighter": "^15.5.7",
     "ajv": "^8.12.0",
     "ajv-formats": "^2.1.1",
+    "ansi-to-react": "^6.1.6",
```
mwylde, line 36: is this used?

mwylde (commented):
Overall looking good, a few more comments.

From the UX side, can we switch automatically to the error tab if there are validation failures? Currently the experience is a bit annoying because every time you click "check" it moves to the graph, so you constantly have to switch back to the errors tab as you're iterating.

Nit: the scrolling is a bit weird:
image

I'd expect the whole thing to scroll, rather than individual errors

arroyo-api/src/pipelines.rs
mwylde: that was supposed to be for the line above `_client`

jbeisen: Sure

arroyo-api/src/pipelines.rs
jbeisen: We're changing the response format, and to be consistent with the new endpoint I think we should rename it.

arroyo-console/package.json
jbeisen: Unfortunately not, I've removed it.

arroyo-api/src/pipelines.rs
jbeisen: Yes, done.

arroyo-compiler-service/src/main.rs
jbeisen: Done

arroyo-api/src/pipelines.rs
mwylde: do we need to change the response format? since it's now doing the same thing as it was previously

jbeisen:
I've fixed the scroll issue and made it switch to the Errors tab when appropriate. I also moved the Alert into the tab - 

image

arroyo-rpc/src/types.rs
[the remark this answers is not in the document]
jbeisen: > Why are we returning the udfs at all

My thinking is that POST endpoints are typically supposed to return a representation of the object that was created, and even though we're not creating an object with this request, we can still return the thing that was validated, kind of just like saying _this was the thing we validated_.

> does it make sense to distinguish between the empty vec case and none cases for errors

This is debatable but I think most people would [say](https://stackoverflow.com/questions/1969993/is-it-better-to-return-null-or-empty-collection) an empty vector is preferred because it avoids the class of null/undefined related errors. 

> I guess the weird thing is that we're using rust-style errors for an http endpoint, where you'd expect an error response in the error case.

I'm not sure that's the proper use of HTTP error codes, since the _errors_ we're talking about here originate from the code provided by the client, not on the server. I think 200 responses are still appropriate when the thing we're validating ends up being invalid.

arroyo-api/src/pipelines.rs
jbeisen: In my opinion it's cleanest to have both validate endpoints as similar as possible, and like I say [here](https://github.com/ArroyoSystems/arroyo/pull/359#discussion_r1357562372), I think this type of "error" that we're reporting with these endpoints is distinct from the type of "error" that HTTP response codes are supposed to represent. But if you feel strongly I can put it back.

mwylde (approved):

PR #364, opened 2023-10-11
support numeric indexes off of arrays.

jacksonrnewhouse:
This fixes the return type on regexp_match and implements indexing into arrays.


arroyo-sql/src/expressions.rs
```diff
@@ -3345,6 +3361,29 @@ impl CodeGenerator<ValuePointerContext, TypeDef, syn::Expr> for DataStructureFun
...
+            } => {
+                let array_expr = array_expression.generate(input_context);
+                let array_nullable = array_expression
+                    .expression_type(input_context)
+                    .is_optional();
+                match array_nullable {
+                    true => parse_quote!({
+                        let array = #array_expr;
+                        if let Some(array) = array {
+                            array.get(#index - 1).cloned()
```
mwylde, line 3376: This produces a compilation error if the for a[0]:

```
error: this arithmetic operation will overflow
```

mwylde (changes requested):

mwylde (approved):

PR #366, opened 2023-10-12
Support multiple projections from a single unnest

mwylde:
This PR relaxes a restriction in the initial unnest implementation (#354) to allow multiple projection columns to reference the same unnested value.

This query is now allowed:

```sql
create table events AS (
    select 
        extract_json_string(event, '$.id') as id,
        extract_json_string(event, '$.type') as type, 
        extract_json_string(event, '$.actor.login') as "user",
        extract_json_string(event, '$.repo.name') as repo 
    FROM
        (select unnest(extract_json(value, '$[*]')) as event from raw_events));


select distinct(id), type, "user", repo
from events;
```

### DIFF
--- a/arroyo-sql/src/pipeline.rs
+++ b/arroyo-sql/src/pipeline.rs
@@ -19,7 +19,7 @@ use quote::quote;
 use crate::code_gen::{CodeGenerator, ValuePointerContext, VecAggregationContext};
 use crate::expressions::{AggregateComputation, AggregateResultExtraction, ExpressionContext};
 use crate::external::{ProcessingMode, SqlSink, SqlSource};
-use crate::operators::UnnestProjection;
+use crate::operators::{UnnestFieldType, UnnestProjection};
 use crate::schemas::window_type_def;
 use crate::tables::{Insert, Table};
 use crate::{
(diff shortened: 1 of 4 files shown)


arroyo-sql/src/pipeline.rs
```diff
@@ -553,32 +553,38 @@ impl<'a> SqlPipelineBuilder<'a> {
...
                     fields,
-                    unnest_col,
                     unnest_inner,
-                    unnest_outer,
                     format: None,
                 }),
             ))
         } else {
-            let projection = Projection::new(fields);
+            let projection = Projection::new(fields.into_iter().map(|(a, b, _)| (a, b)).collect());
```
jacksonrnewhouse, line 587: Maybe name these variables so it is easier to know what the triple corresponds to?

jacksonrnewhouse (approved):

arroyo-sql/src/pipeline.rs
mwylde: this is just a generic function that drops the last tuple, so I didn't think they needed distinctive names (you could imagine extracting this for example).

PR #368, opened 2023-10-13
Add regex dependency to UDF crate

mwylde:
This enables regex usage in UDFs without editing the crate dependencies and building a custom image.

### DIFF
--- a/docker/build_base/udfs/Cargo.toml
+++ b/docker/build_base/udfs/Cargo.toml
@@ -12,6 +12,7 @@ chrono = "0.4"
 serde = "1.0"
 serde_json = "1.0"
 arroyo-types = { path = "/opt/arroyo/src/arroyo-types" }
+regex = "1"
 
 [package.metadata.wasm-pack.profile.release]
 wasm-opt = false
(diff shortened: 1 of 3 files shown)


docker/build_base/udfs/Cargo.toml
```diff
@@ -12,6 +12,7 @@ chrono = "0.4"
 serde = "1.0"
 serde_json = "1.0"
 arroyo-types = { path = "/opt/arroyo/src/arroyo-types" }
+regex = "1"
```
jacksonrnewhouse, line 15: Can you also add this to the code that generates the whole project in `arroyo-controller/src/compiler.rs`?

jacksonrnewhouse (commented):
Need to also update compiler.rs. 

Maybe add a test to the smoke_test? There's an optional file argument for UDF definitions.

mwylde:
Do you think it's worth adding a smoke test? I'm not actually sure which build pathway it uses, but it won't be one used by our users.

jacksonrnewhouse (approved):

jbeisen (approved):

PR #369, opened 2023-10-15
Update dependencies to address dependabot alerts

mwylde:


copy-artifacts/.gitignore
```diff
@@ -0,0 +1 @@
+Cargo.lock
```
mwylde, line 1: copy-artifacts is now part of our main workspace, so it doesn't use a separate Cargo.lock file, but one was present

jbeisen (approved):

PR #373, opened 2023-10-17
sql: have regexp_match return all matches.

jacksonrnewhouse:


mwylde (commented):
There's a test you will also need to update

PR #379, opened 2023-10-20
Add headers and multiple subscription messages to Websocket source

jbeisen:
Add headers and subscription_messages fields to the Websocket schema and
include them in the request. The console uses a new ArrayWidget
component to render this type of schema.

---

The subscription messages can be set in sql like this:

```
    "subscription_messages.0" = 'first message',
    "subscription_messages.1" = 'second message',
```

Or if there's only 1, the old way:

```
    subscription_message = 'only message',
```

Or through the form like this:

image

### DIFF
--- a/connector-schemas/websocket/table.json
+++ b/connector-schemas/websocket/table.json
@@ -11,14 +11,35 @@
             ],
             "format": "uri"
         },
+        "headers": {
+            "title": "Headers",
+            "type": "string",
+            "description": "Comma separated list of headers to send with the request",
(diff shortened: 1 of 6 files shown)


connector-schemas/websocket/table.json
```diff
@@ -11,14 +11,25 @@
             ],
             "format": "uri"
         },
-        "subscription_message": {
```
mwylde, line 14: This will break existing websocket connections in your DB. Do you think it's worth keeping the old "subscription_message" around to provide backwards-compatibility?

mwylde (approved):
Looks great!

PR #383, opened 2023-10-24
arroyo-sql-testing: reduce the size of cars.json so tests run more qu…

jacksonrnewhouse:
…ickly.


arroyo-sql-testing/Cargo.toml
```diff
@@ -4,6 +4,7 @@ version = "0.7.0"
 edition = "2021"
 
 [features]
+default = ["integration-tests"]
```
mwylde, line 7: is this intentional?

jacksonrnewhouse: No, I just do that to get VS Code to resolve the feature.

mwylde (approved):

PR #385, opened 2023-10-27
sql: Implement Delta Lake Sink

jacksonrnewhouse:
This integrates with Delta Lake, using the deltalake crate to add parquet files to the commit log. This is done by subtask 0 and required additional message passing between subtasks. This message passing is implemented via the introduction of OperatorCommitData, which committing operators can send back to the controller. This will be written into the checkpoint metadata and broadcast to workers as part of the commit message. Obviously this should be small, not including data. For our purposes, this will be multipart files that are ready to be committed.

### DIFF
--- a/arroyo-state/src/committing_state.rs
+++ b/arroyo-state/src/committing_state.rs
@@ -1,15 +1,23 @@
-use std::collections::HashSet;
+use std::collections::{HashMap, HashSet};
+
+use arroyo_rpc::grpc::{OperatorCommitData, TableCommitData};
 
 pub struct CommittingState {
     checkpoint_id: i64,
     subtasks_to_commit: HashSet<(String, u32)>,
+    committing_data: HashMap<String, HashMap<String, HashMap<u32, Vec<u8>>>>,
 }
 
 impl CommittingState {
-    pub fn new(checkpoint_id: i64, subtasks_to_commit: HashSet<(String, u32)>) -> Self {
+    pub fn new(
+        checkpoint_id: i64,
+        subtasks_to_commit: HashSet<(String, u32)>,
+        committing_data: HashMap<String, HashMap<String, HashMap<u32, Vec<u8>>>>,
+    ) -> Self {
         Self {
             checkpoint_id,
             subtasks_to_commit,
+            committing_data,
         }
     }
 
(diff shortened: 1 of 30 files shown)


arroyo-state/src/committing_state.rs
```diff
@@ -25,13 +33,35 @@ impl CommittingState {
...
-        Self {
-            checkpoint_id,
-            subtasks_to_commit,
-        }
+    pub fn committing_data(&self) -> HashMap<String, OperatorCommitData> {
+        let operators_to_commit: HashSet<_> = self
+            .subtasks_to_commit
+            .iter()
+            .map(|(operator_id, _subtask_id)| operator_id.clone())
+            .collect();
```
mwylde, line 42: why do we collect this then immediately turn it back into an iterator?

arroyo-state/src/lib.rs
```diff
@@ -416,6 +417,12 @@ impl<S: BackingStore> StateStore<S> {
     pub async fn load_compacted(&mut self, compaction: CompactionResult) {
         self.backend.load_compacted(compaction).await;
     }
+
+    pub async fn insert_committing_data(&mut self, table: char, committing_data: Vec<u8>) {
```
mwylde, line 421: Is committing data created fresh for each checkpoint? I.e., when I call this am I inserting it specifically in this checkpoint?

Following on from that:
* Does it make sense to include the epoch here so that everyone is aligned on what checkpoint we're talking about?
* In the future we will likely want to support aborting/timing out checkpoints; how would this interact with this system?

mwylde (commented):
Overall looks good (as far as I follow it). A few questions / comments.

arroyo-state/src/committing_state.rs
jacksonrnewhouse: It's deduplicating them by a collecting into a hashset.

arroyo-state/src/lib.rs
jacksonrnewhouse: I've added the epoch and a check on it. For aborting/timing out checkpoints we'll have to do some additional design. Currently the system is pretty optimistic that the checkpoint data will make it to state. When cancelling a checkpoint we'll have some subtasks that have already written that data, while others will have not received the barrier and should basically push all the data into the next checkpoint. I think it's doable, but a separate project.

mwylde (approved):

PR #386, opened 2023-10-30
Initial support for Avro formats

mwylde:
This PR adds initial support for the Avro format, initially only for deserialization in sources:

```sql
CREATE TABLE pizza_orders_avro (
    value TEXT
) WITH (
    connector = 'kafka',
    topic = 'pizza_orders_avro',
    type = 'source',
    format = 'avro',
    'schema_registry.endpoint' = 'http://localhost:8081',
    'avro.confluent_schema_registry' = 'true'
);
```

Initially supported:
* Reading Avro data with either an embedded schema, a pre-registered schema, or with schemas fetched from Confluent Schema Repository
* Struct generation from a subset of Avro features (records, primitives, and unions of `[T, null]`
* "Unstructured" avro (via avro-to-json conversion)
* Schema evolution (via confluent schema registry or pre-registered schema)

Followup work for more complete avro support will include:
* Serialization (sink) support
* Complete union/array support

The current implementation is pretty inefficient. In order to support avro records with features that we don't currently support statically, currently avro is parsed, then converted to JSON, then deserialized into the struct via our normal JSON deserialization pathway. This allows us to utilize the existing RawJson approach for unsupported features.

This PR also reworks how we interact with Confluent Schema Registry. The schema registry is now configured as part of the Kafka connection, so as to not require redefining it each time you create a table. We also now to the schema resolution as part of connection table creation, rather than having the frontend request the schema and then fill it in.


arroyo-rpc/src/schema_resolver.rs
```diff
@@ -0,0 +1,167 @@
...
+use std::time::Duration;
+use tracing::warn;
+
+#[async_trait]
+pub trait SchemaResolver: Send {
+    async fn resolve_schema(&self, id: u32) -> Result<Option<String>, String>;
+}
+
+/// A schema resolver that return errors when schemas are requests; this is intended
+/// to be used when schemas as embedded into the message and we do not expect to
```
jacksonrnewhouse, line 15: There's a wording issue here.

arroyo-rpc/src/schema_resolver.rs
```diff
@@ -0,0 +1,167 @@
...
+use apache_avro::Schema;
+use async_trait::async_trait;
+use reqwest::{Client, StatusCode, Url};
+use serde::{Deserialize, Serialize};
+use std::time::Duration;
+use tracing::warn;
+
+#[async_trait]
+pub trait SchemaResolver: Send {
+    async fn resolve_schema(&self, id: u32) -> Result<Option<String>, String>;
```
jacksonrnewhouse, line 11: Why not use anyhow errors?

jacksonrnewhouse (approved):
Just a couple of comments

arroyo-rpc/src/schema_resolver.rs
mwylde: Because the errors are meant to be used as the message in a UserError, which takes a string. You could rely on `anyhow::Error::to_string`, but that includes various formatting logic that I don't want; the SchemaResolver should be responsible for producing a good String error message that can be shown to the user.

PR #387, opened 2023-10-31
Add global UDFs feature

jbeisen:
This change consists of the following:

In the API, this adds new endpoints for creating and managing global UDFs. The global UDFs are stored in a new table, while local UDFs are still stored in the pipelines table.

In the controller and compiler service, this changes the UDF validation and compilation logic to include global UDFs. We traverse the job graph to to determine which UDFs are used, and include only those.

In the console, this adds adds several new components to the create pipeline page to support browsing local and global UDFs, viewing UDFs in multiple tabs, and creating and saving new UDFs.

---
A few issues left to fix:
- ~fix smoke tests~
- ~fix React focus issue with causing the arrow keys not to work in the Globalize UDF modal~
---

The create pipeline page:
image

Edit tab popover:
image

Modal when globalizing a UDF:
image

Global UDFs display in a folder-like structure with a popover showing the description and definition
image

Error tab shows query and UDF errors together:
image


mwylde:
A couple of design nits:

The UDF editor could use a bit of padding at the top:
image
(compare to the query editor)

Can we have the popup to the right instead of down, and get rid of the animation:

image

I think it's a bit frustrating when you switch to the UDF page you have to wait for the popup to clear before you can see the function (and I keep doing it by accident when I'm trying to edit)

There's an extra line at the top of the UDFs tab:
image

arroyo-api/src/udfs.rs
```diff
@@ -0,0 +1,248 @@
...
+pub async fn validate_udf(
+    State(state): State<AppState>,
+    bearer_auth: BearerAuth,
+    WithRejection(Json(validate_udf_post), _): WithRejection<Json<ValidateUdfPost>, ApiError>,
+) -> Result<Json<UdfValidationResult>, ErrorResp> {
+    let auth_data = authenticate(&state.pool, bearer_auth).await?;
+
+    // Return an ok (valid) if the controller is not available or if it fails to validate the UDFs
+
+    let mut controller = match ControllerGrpcClient::connect(state.controller_addr.clone()).await {
```
mwylde, line 172: not necessary now, but eventually we should store the connection so that we're not re-establishing it on every request

arroyo-api/src/udfs.rs
```diff
@@ -0,0 +1,248 @@
...
+)]
+pub async fn create_udf(
+    State(state): State<AppState>,
+    bearer_auth: BearerAuth,
+    WithRejection(Json(req), _): WithRejection<Json<UdfPost>, ApiError>,
+) -> Result<Json<GlobalUdf>, ErrorResp> {
+    let client = client(&state.pool).await.unwrap();
+    let auth_data = authenticate(&state.pool, bearer_auth).await.unwrap();
+
+    let pub_id = generate_id(IdTypes::Udf);
```
mwylde, line 55: Should we be validating the UDF before we create it? It looks like right now we rely on the console / api consumer to do the right thing. It would also be good to do this within a transaction with the creation so that we can avoid race co conditions (like two requests to create the same UDF both succeeding).

arroyo-controller/src/lib.rs
```diff
@@ -447,35 +447,69 @@ impl ControllerGrpc for ControllerServer {
...
+        let function = match function {
+            Some(f) => f,
+            None => {
+                return Ok(Response::new(CheckUdfsResp {
+                    result: ValidationResult::Error as i32,
+                    errors: vec!["No function found in UDF".to_string()],
+                    udf_name: None,
+                }))
+            }
+        };
```
mwylde, line 501: Let-else is really nice for this pattern:

```suggestion
        let Some(function) = function else {
              return Ok(Response::new(CheckUdfsResp {
                  result: ValidationResult::Error as i32,
                  errors: vec!["No function found in UDF".to_string()],
                  udf_name: None,
              }))
          }
        };
```

arroyo-sql/src/expressions.rs
```diff
@@ -3617,6 +3628,14 @@ pub struct RustUdafExpression {
     ret_type: TypeDef,
 }
 impl RustUdafExpression {
+    pub(crate) fn name(&self) -> String {
+        self.name.clone()
+    }
+
+    pub fn expressions(&mut self) -> Vec<&mut Expression> {
+        self.args.iter_mut().map(|(_, e)| e).collect()
+    }
```
mwylde, line 3637: maybe better to return this as an iter so that you're not creating a vec that's not needed

```suggestion
    pub fn expressions(&mut self) -> impl Iterator<Item = &mut Expression> {
        self.args.iter_mut().map(|(_, e)| e)
    }
```

arroyo-sql/src/lib.rs
```diff
@@ -688,3 +705,22 @@ pub fn get_test_expression(
         expected_result,
     )
 }
+
+pub fn has_duplicate_udf_names(udf_defs: &Vec<String>) -> Result<bool> {
+    let mut udf_names = HashMap::new();
```
mwylde, line 710: You can also use a HashSet instead of a HashMap<K, ()>.

arroyo-sql/src/lib.rs
```diff
@@ -688,3 +705,22 @@ pub fn get_test_expression(
...
+    for udf_def in udf_defs {
+        let file = syn::parse_file(udf_def)?;
```
mwylde, line 712: This will cause everything to fail if you have a single bad UDF; is that what we want?

arroyo-sql/src/operators.rs
```diff
@@ -34,6 +34,13 @@ impl Projection {
     pub fn output_struct(&self) -> StructDef {
         self.expression_type(&ValuePointerContext::new())
     }
+
+    pub fn expressions(&mut self) -> Vec<&mut Expression> {
```
mwylde, line 38: ditto here — we probably want an iterator instead of Vec

mwylde (changes requested):

arroyo-api/src/pipelines.rs
```diff
@@ -56,19 +56,51 @@ where
...
+        .await
+        .map_err(|e| anyhow!(format!("Error global global UDFs: {}", e)))?
+        .into_iter()
+        .map(|u| u.into())
+        .collect::<Vec<GlobalUdf>>();
+
+    // error if there are duplicate local or duplicate global UDF names,
+    // but allow global UDFs to override local ones
+
+    if has_duplicate_udf_names(
```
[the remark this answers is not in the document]
jbeisen: Or you can just delete one of the global UDFs? I could also make it impossible to create duplicate global UDFs.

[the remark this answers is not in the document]
mwylde: Yeah, I think that would be better

arroyo-api/src/udfs.rs
jbeisen: I've made this transactional and added a unique constraint on the name in the database (this required splitting out a `prefix` column, so you'll need to reapply the migration). In the UI it presents like this:

image

If we want to do the full validation as part of this transaction then we need to decide what to do about _failure to validate_ errors, like when the compiler service isn't running.

arroyo-controller/src/lib.rs
jbeisen: Thanks, some day I'll get the hang of this.

arroyo-sql/src/expressions.rs
jbeisen: Done

jbeisen:
@mwylde I've pushed a few changes:

- I made UDF creation transactional so it now only completes after it successfully does the same validation as the validation endpoint. Separate from that the validation it also checks for duplicates within the local and global UDFs.
- I made both the create UDF and validate UDF endpoints hard fail (return 500 or 503) if the controller fails to do the validation or the controller isn't available. So we're essentially requiring the compiler service to use UDFs. I think this is necessary to prevent invalid global UDFs.
- I simplified the gRPC methods for UDF validation, getting rid of the `ValidationResult` enum, and just relying on `errors` to determine validity, which I think is cleaner

arroyo-sql/src/lib.rs
jbeisen: I've changed it to log a warning and continue

mwylde (approved):

PR #393, opened 2023-11-06
Define dependencies in UDF definition

jbeisen:
Allow the user to define a special comment in the UDF definition containing the [dependencies] block of Cargo.toml. The controller parses the comment, and builds a separate crate for each individual UDF.

This commit also:
 - Removes local compilation by the controller and makes running the compiler service required.
 - Removes the UDF language field since we only support Rust and it was inconvenient to maintain.

---

image


arroyo-sql/src/lib.rs
```diff
@@ -338,7 +338,7 @@ impl ArroyoSchemaProvider {
                 UdfDef {
                     args,
                     ret,
-                    def: function.to_token_stream().to_string(),
+                    def: body.to_string(),
```
jbeisen, line 341: Note that we're now writing the whole definition to `lib.rs`. This way `use` statements don't need to be inside the function.

arroyo-compiler-service/src/main.rs
```diff
@@ -209,6 +209,13 @@ impl CompileService {
             }
         }
 
+        // clean up the udfs_dir
+        if !self.debug {
+            tokio::fs::remove_dir_all(&self.build_dir.join("udfs_dir"))
+                .await
+                .unwrap();
```
mwylde, line 216: These should return errors instead of unwrapping so that errors are properly propagated to the client

arroyo-compiler-service/src/main.rs
```diff
@@ -243,6 +250,61 @@ impl CompileService {
...
+    async fn write_udf_crates(&self, udf_crates: &Vec<UdfCrate>) -> anyhow::Result<()> {
+        let build_dir = &self.build_dir;
+
+        // write individual udf crates
+        for udf_crate in udf_crates {
+            let udf_build_dir = self.build_dir.join("udfs_dir").join(&udf_crate.name);
+
+            tokio::fs::create_dir_all(&udf_build_dir.join("src"))
+                .await
+                .unwrap();
```
mwylde, line 263: ditto here

arroyo-compiler-service/src/main.rs
```diff
@@ -268,33 +330,44 @@ impl CompilerGrpc for CompileService {
...
+            .arg(format!("--package={}", &name))
             .arg("--message-format=json")
             .output()
             .await?;
 
+        // clean up the udfs_dir
+        if !self.debug {
+            tokio::fs::remove_dir_all(&self.build_dir.join("udfs_dir"))
+                .await
+                .unwrap();
```
mwylde, line 365: and here

arroyo-compiler-service/src/main.rs
```diff
@@ -268,33 +330,44 @@ impl CompilerGrpc for CompileService {
...
+            .current_dir(&udf_build_dir)
             .arg("check")
-            .arg("--package=udfs")
+            .arg(format!("--package={}", &name))
             .arg("--message-format=json")
             .output()
             .await?;
 
+        // clean up the udfs_dir
+        if !self.debug {
```
mwylde, line 362: why do we have this check?

jbeisen: Looking at the generated UDF crates was helpful during development, so I figured I'd make that possible in debug mode.

arroyo-console/src/udf_state.ts
```diff
@@ -135,9 +135,12 @@ export const getLocalUdfsContextValue = () => {
     const id = generate_udf_id();
     const functionName = `new_udf`;
     const definition =
+      `/*\n` +
+      `[dependencies]\n\n` +
```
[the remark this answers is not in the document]
jbeisen: I actually did remove the default dependencies with this change. I personally think this is clear as is, since anyone writing rust code and who knows about a dependency existing should be somewhat familiar with a Cargo.toml. Maybe a better place for that kind of hand holding is through the Catalog thing. Like we can add an example in here with a UDF + dependencies: 
image

arroyo-compiler-service/src/main.rs
jbeisen: I've changed it to always clean up the directory, and to do it before checking/compiling.

mwylde (approved):

PR #395, opened 2023-11-07
Refactor join processing

jbeisen:
Simplify some conditionals and change some function headers.


jbeisen:
Decided to include this in #420

PR #403, opened 2023-11-10
S3 source connector + Parquet

jacksonrnewhouse:
This takes #327 and adds on Parquet support. Doing this directly instead of https://github.com/rcjmurillo/arroyo/pull/1 so we can get it into the next release.


mwylde (approved):

mwylde:
We'll want to either figure out the issues with the web creation of file sources/sinks (maybe @jbeisen can help here?) or disable it

rcjmurillo:
thank you for following up the work, will close #327.

PR #407, opened 2023-11-13
Make UDFs public when written to build dir

jbeisen:
When the compiler service writes the UDF to the build dir, parse the definition and make the function public.


arroyo-controller/src/lib.rs
```diff
@@ -459,45 +457,10 @@ impl ControllerGrpc for ControllerServer {
...
-                    }
-                    Item::Use(_) => {}
-                    _ => {
-                        return Ok(Response::new(CheckUdfsResp {
-                            errors: vec![
-                                "Only functions and use statements are allowed in UDFs".to_string()
-                            ],
-                            udf_name: None,
-                        }))
-                    }
```
jbeisen, line 490: As far as I know we don't actually need this check any more, so I didn't implement it in the ArroyoSchemaProvider.

jacksonrnewhouse (approved):

PR #408, opened 2023-11-13
arroyo-connectors: Test code-gen for JSON Schema and Avro.

jacksonrnewhouse:
This updates arroyo-sql-testing by registering two new kafka sources, one with a schema defined by JsonSchema, the other using an Avro definition. It is invoked from full_query_testing, and just confirms code is generated without issue. This required two changes:
1. In arroyo-sql avro.rs and json_schema.rs were importing `crate::*`. This only works if the definitions they need are at the top level of the crate. This is true for real pipelines, but not the test infrastructure.
2. We changed how we specify timestamp deserialization from JSON, and json_schema.rs was not migrated at that time. The tests will stop this from happening again.

### DIFF
--- a/arroyo-sql-macro/Cargo.toml
+++ b/arroyo-sql-macro/Cargo.toml
@@ -10,10 +10,14 @@ proc-macro = true
 arroyo-sql = { path = "../arroyo-sql" }
 arroyo-datastream = { path = "../arroyo-datastream" }
 arroyo-connectors = { path = "../arroyo-connectors" }
+arroyo-rpc = { path = "../arroyo-rpc" }
+
+anyhow = "1"
 
 syn = {version = "2", features = ["full"]}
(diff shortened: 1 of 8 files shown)


arroyo-sql-macro/Cargo.toml
```diff
@@ -7,13 +7,18 @@ edition = "2021"
 proc-macro = true
 
 [dependencies]
+arroyo-api = { path = "../arroyo-api" }
```
mwylde, line 10: This dependency is a bit unfortunate, as now to build the macro we need to build essentially everything. It also doesn't appear that any imports from arroyo_api are used, so maybe this can be removed?

jacksonrnewhouse: Done

mwylde (approved):

PR #412, opened 2023-11-15
Clean up file system table schema and support maps in UI

jbeisen:
Remove some unnecessary nesting in the file system table schema,
standardize the field capitalization, and add a MapWidget component to
support a Storage Options string map. Also renders all helper text in
the JsonForm as markdown so it can include links.

image

Duplicate keys error handling:
image


jacksonrnewhouse (approved):
Looks great!

PR #415, opened 2023-11-17
Update dependencies

mwylde:
Updates dependencies and removes unused dependencies.

An earlier version of this PR had removed an unused dependency on arroyo-sql from arroyo-controller, but in the mean time this started being used in #359.

However, the reliance on arroyo-sql was fairly minor, so I refactored a bit to move the code that used it—a transformation on UDFs before sending to the compiler service—into the API instead of having the controller do it. This allows us to keep our dependency tree flatter and should help (moderately) with compile times by allowing more parallelism.

### DIFF
--- a/arroyo-sql/src/lib.rs
+++ b/arroyo-sql/src/lib.rs
@@ -48,6 +48,7 @@ use datafusion_common::DataFusionError;
 use prettyplease::unparse;
 use regex::Regex;
 use std::collections::HashSet;
+use std::sync::OnceLock;
 use std::time::{Duration, SystemTime};
 use std::{collections::HashMap, sync::Arc};
 use syn::{parse_file, parse_quote, parse_str, FnArg, Item, ReturnType, Visibility};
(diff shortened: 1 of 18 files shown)


arroyo-sql/src/lib.rs
```diff
@@ -354,9 +355,11 @@ impl ArroyoSchemaProvider {
     }
 }
 
-pub fn parse_dependencies(definition: &str) -> Result<String> {
+pub fn parse_dependencies(definition: &str) -> anyhow::Result<String> {
     // get content of dependencies comment using regex
-    let re = Regex::new(r"\/\*\n(\[dependencies\]\n[\s\S]*?)\*\/").unwrap();
+    static REGEX: OnceLock<Regex> = OnceLock::new();
+    let re = REGEX.get_or_init(|| Regex::new(r"\\*\n(\[dependencies\]\n[\s\S]*?)\*/").unwrap());
```
jbeisen, line 361: nice

jbeisen (approved):

PR #416, opened 2023-11-20
Implement json sink for schema registry, use connection types for sinks, add auth for schema registry.

jacksonrnewhouse:
This implements a number of features in order to write to a confluent topic which has a JsonSchema in a confluent registry. In particular,
* Add confluent_schema_registry version to JsonFormat. This will be written in the 5 byte header of messages written to kafka for sinks using the schema registry.
* Add an optional `struct_name` to arroyo-sql's Projection, so that it will use structs generated with the correct format.
* Add basic auth for kafka schema registry, currently named "API Key" and "API Secret", which is how Confluent Cloud names them.
* unify treatment of sources and sinks in arroyo-sql as "connections". This means that two queries emitting to the same sink will share a sink node. Also properly tracks dependencies between pipelines and sink connections.

### DIFF
--- a/arroyo-worker/src/formats/mod.rs
+++ b/arroyo-worker/src/formats/mod.rs
@@ -180,7 +180,7 @@ impl<T: SchemaData> DataSerializer<T> {
     pub fn to_vec(&self, record: &T) -> Option<Vec<u8>> {
         match &self.format {
             Format::Json(json) => {
-                let mut writer: Vec<u8> = vec![];
+                let mut writer: Vec<u8> = Vec::with_capacity(128);
                 if json.confluent_schema_registry {
                     writer.push(0);
                     writer.extend(json.confluent_schema_version.expect("must have computed schema version to write using confluent schema registry").to_be_bytes());
--- a/arroyo-worker/src/formats/mod.rs
+++ b/arroyo-worker/src/formats/mod.rs
@@ -180,22 +180,22 @@ impl<T: SchemaData> DataSerializer<T> {
     pub fn to_vec(&self, record: &T) -> Option<Vec<u8>> {
         match &self.format {
             Format::Json(json) => {
-                let v = if json.include_schema {
+                let mut writer: Vec<u8> = vec![];
+                if json.confluent_schema_registry {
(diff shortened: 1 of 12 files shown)


arroyo-worker/src/formats/mod.rs
```diff
@@ -180,22 +180,22 @@ impl<T: SchemaData> DataSerializer<T> {
     pub fn to_vec(&self, record: &T) -> Option<Vec<u8>> {
         match &self.format {
             Format::Json(json) => {
-                let v = if json.include_schema {
+                let mut writer: Vec<u8> = vec![];
+                if json.confluent_schema_registry {
+                    writer.push(0);
+                    writer.extend(json.confluent_schema_version.expect("must have computed schema version to write using confluent schema registry").to_be_bytes());
+                }
+                if json.include_schema {
```
mwylde, line 188: I don't think confluent_schema_registry and include_schema are really compatible — they're two different ways to communicate the schema to kafka consumers

arroyo-worker/src/formats/mod.rs
```diff
@@ -180,22 +180,22 @@ impl<T: SchemaData> DataSerializer<T> {
     pub fn to_vec(&self, record: &T) -> Option<Vec<u8>> {
         match &self.format {
             Format::Json(json) => {
-                let v = if json.include_schema {
+                let mut writer: Vec<u8> = vec![];
```
mwylde, line 183: For performance, I think we should follow the serde_json::to_vec behavior of preallocating 128 bytes here; otherwise you could end up repeatedly reallocating the buffer as you write

jacksonrnewhouse: As the initial size? sure.

arroyo-worker/src/formats/mod.rs
jacksonrnewhouse: Happy to prevent this. Have an idea of where we should be validating?

mwylde: upfront, in JsonFormat::from_opts

PR #417, opened 2023-11-21
Add support for inferring schemas

mwylde:
This PR adds support for inferring source and sink connector schemas.

## Source schemas

On the source side, this addresses an issue with connectors that have a fixed schema, like the impulse source or nexmark source. Currently, the user is required to exactly specify the schema in order to create the connector in SQL. For impulse, this looks like:

```sql
CREATE TABLE periodic (
  counter BIGINT UNSIGNED,
  subtask_index BIGINT UNSIGNED
)  WITH (
  connector = 'impulse',
 ...
);
```

which isn't _too_ bad, but for nexmark it's not actually possible to write the schema due to the limitations of datafusion DDL syntax.

Now impulse (and nexmark) sources can be created with an _inferred_ schema by simply omitting the fields:

```sql
CREATE TABLE periodic WITH (
  connector = 'impulse',
 ...
);
```

## Sink

Schema inference is even more useful on the sink side, as it avoids the need to specify schemas when creating the sink and (implicitly) in the query:

```sql
CREATE TABLE blackhole (
  f1  INT,
  f2 INT
) WITH (
  connector = 'blackhole'
);

INSERT INTO blackhole 
SELECT f1, f2 FROM source;
```

becomes

```sql
CREATE TABLE blackhole  WITH 
( connector = 'blackhole');

INSERT INTO blackhole
select f1, f2 FROM source;
```

This also makes web ui-created sinks easier to work with, as it's often awkward to define the schema when creating them. Now they can be created with an inferred schema:

image

---

In this version, I've made inference always occur if there are no fields defined. An open question is whether it would be better if we required users to specify (in the with clause) whether inference should happen.


arroyo-sql/src/lib.rs
```diff
@@ -70,7 +72,7 @@ pub struct UdfDef {
 #[derive(Debug, Clone, Default)]
 pub struct ArroyoSchemaProvider {
     pub source_defs: HashMap<String, String>,
-    tables: HashMap<String, Table>,
+    tables: HashMap<UniCase<String>, Table>,
```
jacksonrnewhouse, line 75: This seems good, given the case insensitive nature of SQL. Can you add a check to when connections are created, or automatically lowercase them?

jacksonrnewhouse (approved):
Looks good, curious about casing on connection names.

arroyo-sql/src/lib.rs
mwylde: I've added a new index to the connections_table table that prevents creating multiple tables with the same (case-insensitive) name

PR #420, opened 2023-11-22
Support joining updating data

jbeisen:
Operators that produce updating data (insert/update/delete) previously could not be upstream from the JoinWithExpiration operator. This change modifies the JoinWithExpiration operator to correctly handle updating data.

All join types, excluding inner joins on non-updating inputs, now produce updating data.

The bulk of the changes are in `join_with_expiration.rs`, where we transforms/coerces all data into updating data so that we reuse the same 4 join processors regardless of input type.

Tested by adding new smoke tests and by running the queries through Flink and checking that the outputs are equivalent.

### DIFF
--- a/arroyo-worker/src/operators/join_with_expiration.rs
+++ b/arroyo-worker/src/operators/join_with_expiration.rs
@@ -13,6 +13,10 @@ use crate::engine::Context;
 #[derive(StreamNode)]
 pub struct JoinWithExpiration<
     K: Key,
+    InT1: Data,
+    InT2: Data,
+    P1: IncomingDataProcessor<InT1, T1>,
+    P2: IncomingDataProcessor<InT2, T2>,
     T1: Data,
     T2: Data,
     Output: Data,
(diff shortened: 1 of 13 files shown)


arroyo-worker/src/operators/join_with_expiration.rs
```diff
@@ -21,299 +25,506 @@ pub struct JoinWithExpiration<
     left_expiration: Duration,
     right_expiration: Duration,
     processor: P,
-    _t: PhantomData<(K, T1, T2, Output)>,
+    pub _t: PhantomData<(K, P1, P2, InT1, InT2, T1, T2, Output)>,
```
jacksonrnewhouse, line 28: I think it'd be a cleaner to just have a `new()` method that other parts of the code call, instead of making this pub or pub (crate)

jacksonrnewhouse (approved):
Looks good, one nit and simplifying the generated code.

PR #421, opened 2023-11-22
Connection profiles sql

jacksonrnewhouse:
This lets users specify the name of a saved connection profile via `connection_profile` key in WITH clause. I think we might want to also add a connection profiles view in the UI, as right now it is pretty hidden.

### DIFF
--- a/arroyo-connectors/src/lib.rs
+++ b/arroyo-connectors/src/lib.rs
@@ -1,6 +1,6 @@
 use anyhow::{anyhow, bail, Context};
 use arroyo_rpc::api_types::connections::{
-    ConnectionSchema, ConnectionType, FieldType, SourceField, SourceFieldType,
+    ConnectionProfile, ConnectionSchema, ConnectionType, FieldType, SourceField, SourceFieldType,
 };
 use arroyo_rpc::primitive_to_sql;
 use arroyo_types::string_to_map;
@@ -120,6 +120,7 @@ pub trait Connector: Send {
         name: &str,
         options: &mut HashMap<String, String>,
         schema: Option<&ConnectionSchema>,
+        profile: Option<&ConnectionProfile>,
     ) -> anyhow::Result<Connection>;
 
     fn from_config(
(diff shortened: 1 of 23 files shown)


arroyo-connectors/src/lib.rs
```diff
@@ -120,6 +120,7 @@ pub trait Connector: Send {
         name: &str,
         options: &mut HashMap<String, String>,
         schema: Option<&ConnectionSchema>,
+        profile: Option<&ConnectionProfile>,
```
mwylde, line 123: Could this be `Self::ProfileT` instead of `ConnectionProfile`, like it is in `from_config`? That would remove the responsibility of the individual connectors for having to do the deserialization and handle errors.

arroyo-sql/src/tables.rs
```diff
@@ -646,10 +648,32 @@ impl Table {
                             .collect(),
                     }))
                 }
-                Some(connector) => Ok(Some(Table::ConnectorTable(
-                    ConnectorTable::from_options(&name, connector, fields, &mut with_map)
+                Some(connector) => {
+                    let connection_profile = match with_map.remove("connection_profile") {
```
mwylde, line 652: It doesn't appear that we validate that the profile is for the correct connector. I think that's worth doing so we can give a reasonable error to the user.

mwylde (changes requested):

build_dir/pipeline/Cargo.toml
```diff
@@ -17,7 +17,7 @@ serde_json = "1.0"
 arrow = { workspace = true}
 arrow-array = { workspace = true}
 arroyo-types = { path = "../../arroyo-types" }
-arroyo-worker = { path = "../../arroyo-worker" }
+arroyo-worker = { path = "../../arroyo-worker" , features = ["kafka-sasl"] }
```
mwylde, line 20: Did you mean to commit this?

arroyo-sql/src/tables.rs
jacksonrnewhouse: Done

build_dir/pipeline/Cargo.toml
jacksonrnewhouse: nope, reverted.

mwylde (approved):

PR #424, opened 2023-11-27
bump versions to 0.8.0.

jacksonrnewhouse:


mwylde (changes requested):
arroyo-server-common/src/lib.rs#41 also needs to be updated

mwylde (approved):

PR #437, opened 2023-12-02
Switch rust builds to ubicloud runners

mwylde:
This is a 10x cost savings over github's runners


crohr:
@mwylde I know you're probably happy with Ubicloud, but just in case you're interested, I ran the same workflow using [RunsOn](https://runs-on.com) and got a build in less than 5 min. Same (or even cheaper pricing), with everything on-premise.

https://github.com/runs-on/arroyo/actions/runs/7723361513/job/21053291933

Issue #438, opened 2023-12-07
Group bys with floats fail to compile


mwylde:
```sql
create table nexmark with (
    connector = 'nexmark',
    event_rate = '5'
);

select cast(bid.price as float) 
from nexmark
group by 1;
```

```
error[E0277]: the trait bound `f32: std::cmp::Eq` is not satisfied
   --> pipeline/src/main.rs:290:5
    |
278 | #[derive(Eq, Hash, Ord)]
    |          -- in this derive macro expansion
...
290 |     pub nexmark_bid_price_: Option<f32>,
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `std::cmp::Eq` is not implemented for `f32`
    |
    = help: the following other types implement trait `std::cmp::Eq`:
              isize
              i8
              i16
              i32
              i64
              i128
              usize
              u8
            and 4 others
    = note: required for `std::option::Option<f32>` to implement `std::cmp::Eq`
```

mwylde:
Fixed in #442

PR #442, opened 2023-12-08
Fail on queries that use floats as keys

mwylde:
Addresses #438 by now producing an error at the planning stage when a float is used as a key in a group or join.


arroyo-sql/src/types.rs
```diff
@@ -901,6 +901,16 @@ impl TypeDef {
         }
     }
 
+    pub fn can_be_key(&self) -> bool {
```
jacksonrnewhouse, line 904: Can we have this be a `Result<()>` which reports that float can't be a key, rather than having the callers report that?

jacksonrnewhouse (approved):

arroyo-sql/src/types.rs
mwylde: That's a good idea

PR #443, opened 2023-12-08
Add control for deserialization error behavior

jbeisen:


mwylde (commented):
This is a great start.

A few general thoughts:
* `invalid_data_behavior` is pretty long — can we come up with a shorter name for this (at least in SQL?)
* I think the key design decision here will be what part of the logic is handled in the connectors vs the deserializer. In the very near future, we might want to support things like "allow 10 failures per second" (of possibly processing or event time) or "10 failures per 100 messages" or something. Where would that logic be possible/easy to implement?
* Longer term, we want to support redirecting invalid records elsewhere in the pipeline. For example, allowing users to define a sink as a dead-letter queue. That will likely need a more complex implementation of this feature that's more integrated with the dataflow, but we don't have to solve that now.

arroyo-worker/src/connectors/sse.rs
```diff
@@ -190,11 +191,13 @@ where
                                             for v in iter {
                                                 match v {
                                                     Ok(value) => {
-                                                        ctx.collector.collect(Record {
-                                                            timestamp: SystemTime::now(),
-                                                            key: None,
-                                                            value,
-                                                        }).await;
+                                                        if let Some(value) = value {
```
mwylde, line 194: The HTTP sources (SSE and websocket) have their own implementation of this that we probably want to replace

arroyo-formats/src/lib.rs
```diff
@@ -290,7 +298,7 @@ impl<T: SchemaData> DataDeserializer<T> {
     pub async fn deserialize_slice<'a>(
         &mut self,
         msg: &'a [u8],
-    ) -> impl Iterator<Item = Result<T, UserError>> + 'a + Send {
+    ) -> impl Iterator<Item = Result<Option<T>, UserError>> + 'a + Send {
```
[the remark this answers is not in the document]
jbeisen: True, changed.

arroyo-formats/src/lib.rs
```diff
@@ -335,7 +347,21 @@ impl<T: SchemaData> DataDeserializer<T> {
...
                 ),
             )
-        })
+        });
+
+        match deserialization {
+            Ok(t) => Ok(Some(t)),
+            Err(e) => match self.invalid_data_behavior {
+                None | Some(InvalidDataBehavior::Drop) => {
+                    warn!("Dropping invalid data: {}", e.details.clone());
```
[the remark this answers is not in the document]
jbeisen: I've moved this logic to `source_collector.rs` and added rate limiting.

jbeisen:
I've made a few changes:
-  moved the handling logic out of the `DataDeserializer` and put it a function `collect_source_record` invoked by the sources
- renamed the option to `bad_data`

About Avro data, I left the existing behavior for non-deserialization errors, which is that the job fails. Should we change that?

arroyo-worker/src/rate_limiter.rs
```diff
@@ -0,0 +1,26 @@
+use std::future::Future;
+use std::ops::Sub;
+use std::time::{Duration, Instant};
+
+pub struct RateLimiter {
```
mwylde, line 5: We don't generally move individual structs into their own modules unless they have very large implementations. Given this has just two methods, I think it fits better in lib.rs

arroyo-worker/src/metrics.rs
```diff
@@ -63,6 +72,13 @@ impl TaskCounters {
                 &task_info.task_index.to_string(),
                 &task_info.operator_name,
             ]),
+            TaskCounters::DeserializationErrors => DESERIALIZATION_ERRORS_COUNTER
+                .with_label_values(&[
+                    &task_info.operator_id,
+                    &task_info.task_index.to_string(),
+                    &task_info.operator_name,
+                ])
+                .clone(),
```
mwylde, line 81: I don't think you need to clone here

arroyo-worker/src/rate_limiter.rs
```diff
@@ -0,0 +1,26 @@
...
+            last: Instant::now().sub(Duration::from_secs(60)),
+        }
+    }
+
+    pub async fn rate_limit<F, Fut>(&mut self, f: F)
+    where
+        F: FnOnce() -> Fut,
+        Fut: Future<Output = ()> + Send,
+    {
+        if self.last.elapsed() > Duration::from_secs(30) {
```
mwylde, line 21: Once every 30 seconds seems a bit too long to me; maybe 5 seconds?

arroyo-worker/src/connectors/source_collector.rs
```diff
@@ -0,0 +1,46 @@
...
+use crate::rate_limiter::RateLimiter;
+use arroyo_formats::SchemaData;
+use arroyo_rpc::formats::BadData;
+use arroyo_types::{Key, Record, SourceError, UserError};
+use std::time::SystemTime;
+use tracing::warn;
+
+/// Collects a source record, handling errors and rate limiting.
+/// Considers the `bad_data` option to determine whether to drop or fail on bad data.
+pub async fn collect_source_record<K: Key, T: SchemaData>(
```
mwylde, line 12: Similarly here — this doesn't need to be its own module for one function

mwylde: Maybe this is actually better as a method on Context? Since we're taking in a &mut Context as the first argument anyways

arroyo-worker/src/connectors/kafka/source/mod.rs
```diff
@@ -272,11 +272,20 @@ where
                                 let iter = self.deserializer.deserialize_slice(v).await;
 
                                 for value in iter {
-                                    ctx.collector.collect(Record {
-                                        timestamp: from_millis(timestamp as u64),
-                                        key: None,
-                                        value: value?,
-                                    }).await;
+                                    match value {
```
mwylde, line 275: Does this need to be updated to use collect_source_record?

arroyo-worker/src/connectors/polling_http.rs
```diff
@@ -270,17 +277,8 @@ where
...
-                                            ctx.collect(Record {
-                                                timestamp: SystemTime::now(),
-                                                key: None,
-                                                value,
-                                            }).await;
-                                        }
-                                        Err(e) => {
-                                            ctx.report_user_error(e).await;
-                                        }
+                                    if let Err(e) = collect_source_record(ctx, SystemTime::now(), record, &self.bad_data, &mut self.rate_limiter).await {
```
mwylde, line 280: I think this needs to be updated to remove the existing bad data handling code

arroyo-worker/src/connectors/sse.rs
```diff
@@ -188,15 +197,7 @@ where
...
-                                                match v {
-                                                    Ok(value) => {
-                                                        ctx.collector.collect(Record {
-                                                            timestamp: SystemTime::now(),
-                                                            key: None,
-                                                            value,
-                                                        }).await;
-                                                    }
-                                                    Err(e) => {
+                                                if let Err(e) = collect_source_record(ctx, SystemTime::now(), v, &self.bad_data, &mut self.rate_limiter).await {
```
mwylde, line 200: Same here — there's existing error handling code that is now duplicative with what collect_source_record is doing

mwylde (changes requested):

arroyo-rpc/src/formats.rs
```diff
@@ -260,6 +260,29 @@ impl Format {
     }
 }
 
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Eq, Hash, PartialOrd, ToSchema)]
+#[serde(rename_all = "snake_case")]
+pub enum BadData {
```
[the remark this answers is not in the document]
jbeisen: I think if/when we add policies like that we'll want something like this:
```
struct BadData {
    limit: i32,
    per_seconds: i32,
}
```

or we could even represent it as a percentage, like an SLA.

But either way, I'm not sure there's a clean way of implementing something backwards compatible with the current binary option.

arroyo-worker/src/connectors/polling_http.rs
jbeisen: It looks like most sources panic on user errors, except for this one and SSE. Is there a reason for that or should we make them all consistent?

arroyo-worker/src/connectors/source_collector.rs
jbeisen: Good idea

arroyo-worker/src/connectors/sse.rs
jbeisen: I've removed this

arroyo-worker/src/connectors/polling_http.rs
jbeisen: I've removed this and made all the sources consistent in handling user errors.

arroyo-worker/src/connectors/kafka/source/mod.rs
jbeisen: Yup, missed this. It's done now.

arroyo-rpc/src/formats.rs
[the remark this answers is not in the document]
jbeisen: I've made the enum variants structs so it gets serialized as a json object so we can add more complex ones in the future.

mwylde (approved):

PR #445, opened 2023-12-08
state: Reduce the panics in the parquet flusher, add retries to StorageProvider, send flusher failure.

jacksonrnewhouse:


arroyo-storage/src/lib.rs
```diff
@@ -119,6 +119,28 @@ fn matchers() -> &'static HashMap<Backend, Vec<Regex>> {
...
+        use std::time::Duration;
+        let mut retries = 0;
+        let max_retries: u32 = 5;
+        let backoff_factor: u64 = 2;
+        loop {
+            match $e {
+                Ok(value) => break Ok(value),
+                Err(e) if retries < max_retries => {
+                    retries += 1;
+                    println!("Error: {}. Retrying...", e);
```
mwylde, line 134: should this be an `error!` instead of println?

arroyo-storage/src/lib.rs
```diff
@@ -119,6 +119,28 @@ fn matchers() -> &'static HashMap<Backend, Vec<Regex>> {
...
+                    let backoff_time = Duration::from_millis(10 * backoff_factor.pow(retries));
```
mwylde, line 135: What do you think about placing a bound on this so it can't get way too large (for example if we later increase the number of retries)?

arroyo-storage/src/lib.rs
```diff
@@ -119,6 +119,28 @@ fn matchers() -> &'static HashMap<Backend, Vec<Regex>> {
     })
 }
 
+macro_rules! retry {
+    ($e:expr) => {{
+        use std::thread::sleep;
+        use std::time::Duration;
+        let mut retries = 0;
+        let max_retries: u32 = 5;
```
mwylde, line 127: If I've done my math right, this is about 600ms from first to last retry. That seems a bit low for a potentially short-lived transient issue. Maybe something closer to 10 seconds would be good?

mwylde (approved):

PR #452, opened 2023-12-14
Add bad data option to create connection form

jbeisen:
Add a bad data option selector to the Define schema stage of the create connection flow.

image

### DIFF
--- a/arroyo-console/src/gen/api-types.ts
+++ b/arroyo-console/src/gen/api-types.ts
@@ -197,10 +197,17 @@ export interface components {
   schemas: {
     AvroFormat: {
       confluentSchemaRegistry?: boolean;
-      embeddedSchema?: boolean;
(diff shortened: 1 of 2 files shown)


arroyo-console/src/gen/api-types.ts
```diff
@@ -197,10 +197,14 @@ export interface components {
   schemas: {
     AvroFormat: {
       confluentSchemaRegistry?: boolean;
-      embeddedSchema?: boolean;
       intoUnstructuredJson?: boolean;
+      rawDatums?: boolean;
       readerSchema?: string;
+      /** Format: int32 */
+      schemaId?: number | null;
```
jbeisen, line 204: @mwylde looks like we missed this in #422

mwylde:
I think this could use more explanation in the UI about what this means and what the options are

mwylde (approved):

PR #453, opened 2023-12-14
Fix issues with enums in json form

jbeisen:
Formik was not properly setting the values when switching between JSON shema 'oneof' variants. This adds a call to reset the field when switching variants.

This change also removes the option of leaving optional enums as null, and instead defaults it to the 'default_value' (for string enums if defined), or the first oneof variant.

For example:

image


mwylde (commented):
This looks good. I think you'll need to also go through all of the schemas and adjust the ordering of the options to align with the defaults. For example, in Kafka we need the non-transactional options to be default for both sources and sinks.

mwylde (approved):

PR #454, opened 2023-12-15
Redesign cluster profile UI and add validation

mwylde:
This PR updates the connection creation flow for connectors that have a separate connection profile (like Kafka and Redis). Currently, the connection profile can be selected from a drop-down list or created as a modal, however there are some issues:
* There's currently no way to inspect an existing connection profile or delete it
* The modal gets unwieldy as configurations get more complex
* We don't validate the profile until the last step of connection creation, at which point it's too late to fix it without just creating a new one

In the new flow, connection profiles have been split out into a separate step. It starts with a UI to select a profile:
image

This UI also allows inspecting and deleting existing profiles.

If the user chooses to create a new profile, they go to this screen:

image

which has an option to validate the connection profile before creating it.

This also involved adding a new API, /connection_profiles/test, and a new method to Connector.

### DIFF
--- a/arroyo-console/src/lib/data_fetching.ts
+++ b/arroyo-console/src/lib/data_fetching.ts
@@ -146,7 +146,7 @@ const pingFetcher = async () => {
 
 export const usePing = () => {
   const { data, error, isLoading } = useSWR('ping', pingFetcher, {
-    refreshInterval: 10000000,
+    refreshInterval: 1000,
     onErrorRetry: (error, key, config, revalidate, {}) => {
       // explicitly define this function to override the exponential backoff
       setTimeout(() => revalidate(), 1000);
--- a/arroyo-console/src/lib/data_fetching.ts
+++ b/arroyo-console/src/lib/data_fetching.ts
@@ -146,7 +146,7 @@ const pingFetcher = async () => {
 
 export const usePing = () => {
   const { data, error, isLoading } = useSWR('ping', pingFetcher, {
-    refreshInterval: 1000,
+    refreshInterval: 10000000,
     onErrorRetry: (error, key, config, revalidate, {}) => {
       // explicitly define this function to override the exponential backoff
       setTimeout(() => revalidate(), 1000);
--- a/arroyo-console/src/routes/connections/ConfigureProfile.tsx
+++ b/arroyo-console/src/routes/connections/ConfigureProfile.tsx
@@ -18,8 +18,9 @@ import {
   LinkBox,
   LinkOverlay,
   useToast,
+  Flex,
 } from '@chakra-ui/react';
-import { useState } from 'react';
+import React, { useState } from 'react';
 import { JsonForm } from './JsonForm';
 import {
   AddIcon,
@@ -118,9 +119,7 @@ const ClusterChooser = ({
 
   return (
     <>
-      <Stack spacing={4} padding={4} maxW={800}>
-        <Heading size={'xx-small'}>Cluster config</Heading>
-
+      <Stack spacing={6} padding={4} maxW={800}>
         <Text fontSize={'sm'}>
           Select an existing {connector}{' '}
           <dfn
@@ -133,7 +132,7 @@ const ClusterChooser = ({
           or create a new one
         </Text>
 
-        <Stack spacing={4} w={800}>
+        <Stack spacing={3} w={800}>
           {connections
             .filter(c => c.connector == connector)
             .map(c => (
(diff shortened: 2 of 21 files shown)


jbeisen:
Nice, a couple thoughts:
- Clicking on a profile and being brought to step 2 feels a bit abrupt, especially since to me they look more like text boxes than buttons. How about first selecting one, then clicking a 'Next' button?
- I think it's a bit odd to have two separate views that are part of the same step. I suggest bringing back the modal for creating the profile with some styling changes to make it less unwieldy. 
- I think we can have a line somewhere explaining what a profile is

arroyo-console/src/routes/connections/ConfigureProfile.tsx
```diff
@@ -0,0 +1,232 @@
...
+      <Stack spacing={8}>
+        <Box>
+          <Button leftIcon={<ArrowBackIcon />} onClick={() => setCreatingCluster(false)}>
+            Select existing
+          </Button>
+        </Box>
+        <CreateProfile
+          connector={connector}
+          addConnectionProfile={c => {
+            mutateConnectionProfiles();
```
jbeisen, line 200: I think it's cleaner to use the `useConnectionProfiles` hook in the components where it's needed rather than passing these things around.

arroyo-console/src/routes/connections/ConfigureProfile.tsx
```diff
@@ -0,0 +1,232 @@
...
+}
+
+const ClusterChooser = ({
+  connector,
+  connections,
+  onSubmit,
+  schema,
+}: {
+  connector: string;
+  connections: Array<ConnectionProfile>;
```
jbeisen, line 73: You can get this from the `useConnectionProfiles` hook below.

arroyo-console/src/lib/data_fetching.ts
```diff
@@ -146,7 +146,7 @@ const pingFetcher = async () => {
 
 export const usePing = () => {
   const { data, error, isLoading } = useSWR('ping', pingFetcher, {
-    refreshInterval: 1000,
+    refreshInterval: 10000000,
```
jbeisen, line 149: Did you mean to commit this?

mwylde:
Yeah, feel the abruptness. I'll add a second step, and some explanation of profiles.

I originally built the whole thing with the modal, but found it was awkward for a few reasons:
* When the content got longer the the page (requiring scrolling) the UX gets wonky
* If the users clicks out of the modal, they lose their work
* Integrating it with testing and feedback made the UI very cluttered
* If you're creating a new profile, 99% that's the one you want to use, but with a modal it feels weird to move automatically to the next step

arroyo-console/src/routes/connections/ConfigureProfile.tsx
```diff
@@ -0,0 +1,232 @@
...
+        title: 'Connection profile deleted',
+        description: `Successfully deleted connection profile ${profile.name}`,
+        status: 'success',
+        duration: 9000,
+        isClosable: true,
+      });
+    }
+  };
+
+  return (
```
jbeisen, line 106: I think it's a good practice to break things down into variables so that the return statements are relatively small, so you can get an idea of the overall layout/content by looking at it.

arroyo-console/src/lib/data_fetching.ts
mwylde: No, I add that to stop the UI from losing state when I restart the api server

arroyo-console/src/routes/connections/ConfigureProfile.tsx
mwylde: I think passing it is is nicer in this case, because otherwise this connector has to handle the possibility that it hasn't loaded yet. By handling that at the higher level and passing in the already fetched connectors, this component gets a bit simpeler.

arroyo-console/src/routes/connections/ConfigureProfile.tsx
mwylde: Hmm, it's not clear to me how much clearer that would make this, since it would just be a couple of variable declarations and then returning them. In the end, it's the same amount of code that needs to be on screen to understand what's happening.

mwylde:
Updated with a two-step selection process:
image

and explanation of connection profiles

jbeisen:
If you select a profile, then delete it, it still lets you continue.

arroyo-console/src/routes/connections/ConfigureProfile.tsx
```diff
@@ -0,0 +1,254 @@
...
+  schema,
+  setCreatingCluster,
+}: {
+  connector: string;
+  connections: Array<ConnectionProfile>;
+  onSubmit: (c: string) => void;
+  schema: JSONSchema7;
+  setCreatingCluster: (b: boolean) => void;
+}) => {
+  const [editingProfile, setEditingProfile] = useState<any>({});
```
jbeisen, line 85: Does this need to be `any`?

jbeisen:
What do you think of this?
image

arroyo-console/src/routes/connections/ConfigureProfile.tsx
```diff
@@ -0,0 +1,254 @@
...
+                  w={'100%'}
+                  p={4}
+                  bg={c.id == selected ? 'blue.700' : 'gray.800'}
+                  border={'1px solid #777'}
+                  _hover={{ bg: c.id == selected ? 'blue.700' : 'gray.600', borderColor: 'black' }}
+                >
+                  <LinkOverlay href={'#'} onClick={() => setSelected(c.id)} />
+                  <HStack spacing={4}>
+                    <Heading size={'14px'}>{c.name}</Heading>
+                    <Text fontStyle={'italic'}>{c.description}</Text>
```
jbeisen, line 148: Gotta handle overflow - 

image

arroyo-console/src/routes/connections/ConfigureProfile.tsx
mwylde: Ah, not any more; previously this was used for creating profiles as well

mwylde:
Thanks for the thorough review! Addressed all of the feedback.

jbeisen (approved):

PR #460, opened 2023-12-19
Validate schema and encoding issues when testing kafka connections

mwylde:
At the end of the connection creation flow we have a testing process, which currently for Kafka is able to validate issues connecting to and reading messages. However, it does not currently validate that the messages themselves are well-formed according to the format that's been configured.

This adds such support, which can catch issues like the wrong format being selected or incorrectly selecting/not selecting confluent schema registry.

### DIFF
--- a/arroyo-console/src/routes/connections/CreateProfile.tsx
+++ b/arroyo-console/src/routes/connections/CreateProfile.tsx
@@ -27,7 +27,7 @@ export const CreateProfile = ({
   const [error, setError] = useState<string | null>(null);
   const [valid, setValid] = useState<boolean | null>(null);
   const [validating, setValidating] = useState<boolean>(false);
-  const [state, setState] = useState<any | null>(null);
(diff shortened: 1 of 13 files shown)


arroyo-console/src/routes/connections/CreateProfile.tsx
```diff
@@ -27,7 +27,7 @@ export const CreateProfile = ({
   const [error, setError] = useState<string | null>(null);
   const [valid, setValid] = useState<boolean | null>(null);
   const [validating, setValidating] = useState<boolean>(false);
-  const [state, setState] = useState<any | null>(null);
+  const state = useRef<any | null>(null);
```
mwylde, line 30: fixes a bug introduced in #454 which caused changes to the form to not always propagate to the state that was actually saved

jacksonrnewhouse (approved):

PR #462, opened 2023-12-19
kafka: correctly detect errors in topic creation and then retry.

jacksonrnewhouse:
The create_topic and delete_topic APIs return a Result<Vec<Result<>>, and we weren't inspecting that inner result. As such, we would occasionally fail to create a topic and then the KafkaSourceFunc would just emit a Watermark::Idle. This checks the API response and if there's a failure waits a second and retries.

### DIFF
--- a/arroyo-worker/src/connectors/kafka/sink/test.rs
+++ b/arroyo-worker/src/connectors/kafka/sink/test.rs
@@ -13,6 +13,7 @@ use rdkafka::consumer::{Consumer, StreamConsumer};
 use rdkafka::producer::Producer;
 use rdkafka::{ClientConfig, Message};
 use tokio::sync::mpsc::channel;
+use tracing::error;
 
 use super::KafkaSinkFunc;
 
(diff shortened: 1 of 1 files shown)


arroyo-worker/src/connectors/kafka/sink/test.rs
```diff
@@ -30,22 +30,40 @@ impl KafkaTopicTester {
...
-            )
-            .await
-            .expect("new topic should be present");
+        let mut tries = 0;
+        while tries < 5 {
+            let delete_result = &admin_client
+                .delete_topics(&[&self.topic], &AdminOptions::new())
+                .await
+                .expect("deletion should have worked")[0];
+            tokio::time::sleep(Duration::from_secs(1)).await;
```
mwylde, line 39: this always adds a second, maybe move this into the retry handler?

arroyo-worker/src/connectors/kafka/sink/test.rs
```diff
@@ -30,22 +30,40 @@ impl KafkaTopicTester {
...
+                        &self.topic,
+                        num_partitions,
+                        rdkafka::admin::TopicReplication::Fixed(1),
+                    )],
+                    &AdminOptions::new(),
+                )
+                .await
+                .expect("new topic should be present")[0];
+            if create_result.is_err() {
+                tries += 1;
```
mwylde, line 61: Do we want a delay here?

arroyo-worker/src/connectors/kafka/sink/test.rs
jacksonrnewhouse: This sleep had already been in there, I think because of some issues with the delete taking some time to propogate on the kafka side of things.

arroyo-worker/src/connectors/kafka/sink/test.rs
jacksonrnewhouse: Added it, which I think let it pass

PR #467, opened 2023-12-20
Handle empty strings in JSON form

jbeisen:
In the StringWidget component, if the string is empty then reset the field to undefined. This prevents validation false positives.

Also use the JSON schema 'dependentRequired' keyword in the Confluent schema, and surface object-level formik errors.

image

### DIFF
--- a/arroyo-console/src/routes/connections/JsonForm.tsx
+++ b/arroyo-console/src/routes/connections/JsonForm.tsx
@@ -43,6 +43,7 @@ function StringWidget({
   errors,
   onChange,
   readonly,
+  resetField,
 }: {
   path: string;
   title: string;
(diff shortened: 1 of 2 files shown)


arroyo-console/src/routes/connections/JsonForm.tsx
```diff
@@ -56,7 +57,16 @@ function StringWidget({
   errors: any;
   onChange: (e: React.ChangeEvent<any>) => void;
   readonly?: boolean;
+  resetField?: (field: string) => any;
 }) {
+  const onChangeWrapper = (e: React.ChangeEvent<any>) => {
+    onChange(e);
+
+    if (resetField && e.target.value == '') {
```
mwylde, line 65: Should this be before the parent onchange? For the connection page, we rely on onchange to synchronize the current state for validation

jbeisen: The issue that since we want to set the value to undefined, we need to do that after onChange sets it to the empty string.

arroyo-console/src/routes/connections/JsonForm.tsx
```diff
@@ -659,7 +669,7 @@ export function FormInner({
       // @ts-ignore
       onChange({ target: { name: path, value: {} } });
     }
-  }, [schema]);
+  }, []);
```
jbeisen, line 672: @mwylde This should make it add the empty object on only the first render

mwylde (approved):

PR #471, opened 2023-12-27
Arrow dataflow

mwylde:
Migrates the dataflow to arrow and removes AOT compilation of queries


arroyo-api/src/pipelines.rs
```diff
@@ -308,31 +311,34 @@ pub(crate) async fn create_pipeline<'a>(
         }
     };
 
-    optimizations::optimize(&mut compiled.program.graph);
+    // TODO: graph optimizations?
```
jacksonrnewhouse, line 314: I think we should be able to move these all into the various datafusion optimization passes.

arroyo-datastream/src/lib.rs
```diff
@@ -36,6 +39,38 @@ use rand::rngs::SmallRng;
 use rand::{Rng, SeedableRng};
 use regex::Regex;
 
+pub const TIMESTAMP_FIELD: &str = "_timestamp";
+
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub struct ArroyoSchema {
+    pub schema: Arc<Schema>,
+    pub timestamp_col: usize,
+    pub key_cols: Vec<usize>,
```
jacksonrnewhouse, line 48: very minor nit: I like `key_indices` a bit better.

arroyo-df/src/lib.rs
```diff
@@ -630,8 +615,35 @@ impl Debug for AggregateCalculation {
 
 #[derive(Debug)]
 struct DataFusionEdge {
```
jacksonrnewhouse, line 617: Should this include the _timestamp column?

arroyo-df/src/lib.rs
```diff
@@ -821,16 +839,20 @@ impl TreeNodeRewriter for QueryToGraphVisitor {
...
                 let table_name = format!("{}", aggregate_index.index());
                 self.local_logical_plan_graph.add_edge(
                     key_index,
                     aggregate_index,
                     DataFusionEdge {
-                        value_schema: input.schema().clone(),
-                        key_schema: Some(key_schema),
+                        schema: input_df_schema,
+                        edge_type: LogicalEdgeType::Shuffle,
+                        // start at 1 to exclude the window
```
jacksonrnewhouse, line 854: I'm not sure I understand this? I think the window field can appear anywhere in the group by list, not at the beginning. The approach I'd taken was to remove the window_field from the list and track where it should be inserted after the aggregate is computed.

arroyo-df/src/plan_graph.rs
```diff
@@ -61,10 +77,11 @@ pub(crate) async fn get_arrow_program(
             .local_logical_plan_graph
             .node_weight(node_index)
             .unwrap();
-        match logical_extension {
+
+        let new_node = match logical_extension {
             crate::LogicalPlanExtension::TableScan(logical_plan) => {
                 let LogicalPlan::TableScan(table_scan) = logical_plan else {
-                    bail!("expected table scan")
+                    panic!("expected table scan")
```
jacksonrnewhouse, line 84: If we have a Result return, I generally perefer bailing. I'm guessing this was to better identify where the error was happening?

arroyo-rpc/proto/api.proto
```diff
@@ -409,3 +410,33 @@ message OperatorCheckpointDetail {
...
+  repeated uint32 key_cols = 3;
+}
+
+message ArrowNode {
+  int32 node_index = 1;
+  string node_id = 2;
+  uint32 parallelism = 3;
+  string description = 4;
+  string operator_name = 5;
+  bytes operator_config = 6;
```
jacksonrnewhouse, line 433: Does it make senes for this to be optional? I guess the operator_name could decide to just not look at it.

arroyo-worker/src/connectors/filesystem/source/mod.rs
```diff
@@ -236,97 +253,97 @@ impl<K: Data, T: SchemaData> FileSystemSourceFunc<K, T> {
...
-                    ))))
-                        as Pin<Box<dyn Stream<Item = Result<T, SourceError>> + Send>>,
-                }))
-                    as Box<dyn Stream<Item = Result<T, SourceError>> + Send + Unpin>;
-                Ok(result)
-            }
-            arroyo_rpc::formats::Format::RawString(_) => todo!(),
-        }
-    }
+    // async fn get_item_stream(
```
jacksonrnewhouse, line 256: Can we delete this? I guess we need to convert JSON into Arrow?

jacksonrnewhouse (approved):
Looks good overall. Some questions/comments, but go ahead to merge after.

arroyo-controller/src/schedulers/mod.rs
```diff
@@ -170,19 +146,24 @@ impl Scheduler for ProcessScheduler {
             println!("Starting in path {:?}", path);
             let workers = self.workers.clone();
             let env_map = start_pipeline_req.env_vars.clone();
+            let program = start_pipeline_req.program.clone();
             tokio::spawn(async move {
-                let mut command = Command::new("./pipeline");
+                let mut command = Command::new("target/debug/arroyo-worker");
```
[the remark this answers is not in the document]
mwylde: This is the path to the worker binary, which by default will be there (for debug). My thinking here is that eventually everything should be bundled into a single binary so the controller can use its own location as the binary for process mode.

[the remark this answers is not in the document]
mwylde: It would be useful to allow it to be used in both debug and release mode so I'll add the equivalent DEBUG flag that we currently have on the compiler

arroyo-datastream/src/lib.rs
mwylde: 👍  makes sense to reserve "col" for the actual array

arroyo-df/src/lib.rs
mwylde: Ok, I'd misread the code here. I'll make it more general.

arroyo-df/src/plan_graph.rs
mwylde: In my mind, they serve two different purposes and it's useful to have both. Returning a result makes sense when it's a user-caused error that we want to report back to them. But when it's a programming error (there should be no way to end up with a LogicalPlanExtension::TableScan that doesn't contain a TableScan) there's no good reason to return that error message to the user; panicking is a sign to us that there's a bug we need to fix.

Ideally stuff like this would be encoded in the type system though

arroyo-rpc/proto/api.proto
mwylde: In the current structure, everything needs a config because there's a unified way to construct operators. What kind of operator are you thinking of that doesn't have any config?

arroyo-df/src/lib.rs
mwylde: Yeah, right now it's implicit in the schema field called _timestamp, but probably better to be explicit

arroyo-df/src/plan_graph.rs
jacksonrnewhouse: Gotchya, I like `unreachable!()` to indicate that we don't believe it is possible to get there.

arroyo-rpc/proto/api.proto
jacksonrnewhouse: I was doing without a config for e.g. the GrpcSink, but having a config with an empty table and connection also works.

mwylde: All connectors have the same config (ConnectorOp) like today

PR #473, opened 2023-12-28
Arrow networking

mwylde:
This is the initial approach for networking with record batches, which relies on the arrow ipc format. There's currently a copy happening on the recv side that can probably be removed in the future, making this much more efficient than our existing networking.


arroyo-worker/src/network_manager.rs
```diff
@@ -177,39 +217,60 @@ impl OutNetworkLink {
...
+                    Some(((quad, dictionary_tracker, msg), s)) = sel.next() => {
+                        match msg {
+                            ArrowMessage::Signal(signal) => {
+                                let data = bincode::encode_to_vec(&signal, config::standard()).unwrap();
+                                let header = Header::from_quad(quad, data.len(), MessageType::Signal);
+                                header.write(&mut Pin::new(&mut self.stream)).await;
+                                self.stream.write_all(&data).await.unwrap();
+                            }
+                            ArrowMessage::Data(data) => {
+                                let (_, encoded_message) = {
```
jacksonrnewhouse, line 258: I think it wouldn't be too hard to pass along the dictionaries emitted here in the future.

jacksonrnewhouse (approved):

arroyo-worker/src/network_manager.rs
mwylde: My concern with dictionaries in the context of streaming is that there's no garbage collection. Over time, they would become unbounded in size without another protocol to manage GCing them.

PR #476, opened 2024-01-02
Fix #252 kafka sink user error

### LINKED ISSUE #252: Kafka sink should produce UserErrror when it cannot connect to cluster
Currently if the user misconfigures their Kafka sink, they will get a panic at runtime:

```2023-08-15T03:33:12.267402Z ERROR arroyo_server_common: panicked at 'Producer creation failed: Transaction error: Failed to initialize Producer ID: Local: Timed out

Caused by:
    Failed to initialize Producer ID: Local: Timed out', /opt/arroyo/src/arroyo-worker/src/connectors/kafka/sink/mod.rs:119:14 panic.file="/opt/arroyo/src/arroyo-worker/src/connectors/kafka/sink/mod.rs" panic.line=119 panic.column=14```

This should be handled and produced to the UI as we do in the Kafka source.

### PR DESCRIPTION
**Pull request by breezykermo**
Closes #252 by communicating the error to the controller (following [the pattern in the Kafka source](https://github.com/ArroyoSystems/arroyo/blob/master/arroyo-worker/src/connectors/kafka/source/mod.rs#L237-L252)). 

When running in development with a basic job that plugs data from the nexmark source to a (badly configured) kafka sink, i.e.:
```sql
insert into badkaf
select process_string(person.name) from nex where person.name is not null;
```
(where `process_string` simply ensures a `String` is always returned, rather than an `Option<String>` -- as my `badkaf` sink expects a raw string.)

Arroyo seems to immediately attempt to go into a checkpoint, and eventually (after several minutes) times out through a failed `self.flush`.  When this happens, the error is appropriately displayed in the UX in my setup:

pic-selected-240102-0422-39

I am not familiar enough with the lifecycle to understand how to test the error code path in `self.process_element` given that the setup described seems to immediately go into a checkmark. I would appreciate guidance on how to do this, and will be happy to update the PR accordingly. 

I haven't added a test to catch this error/panic specifically, as from what I can tell the testing in connectors only confirms happy path.

### DIFF
--- a/arroyo-worker/src/connectors/kafka/sink/mod.rs
+++ b/arroyo-worker/src/connectors/kafka/sink/mod.rs
@@ -174,20 +174,37 @@ impl<K: Key + Serialize, T: SchemaData + Serialize> KafkaSinkFunc<K, T> {
     }
 
     async fn handle_checkpoint(&mut self, _: &CheckpointBarrier, ctx: &mut Context<(), ()>) {
-        self.flush(ctx).await;
-        if let ConsistencyMode::ExactlyOnce {
-            next_transaction_index,
-            producer_to_complete,
-        } = &mut self.consistency_mode
-        {
-            *producer_to_complete = self.producer.take();
-            ctx.state
-                .get_global_keyed_state('i')
-                .await
-                .insert(ctx.task_info.task_index, *next_transaction_index)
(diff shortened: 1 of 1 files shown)


mwylde (approved):
Thanks for the contribution, this looks great!

Issue #469, opened 2023-12-22
Add option to load offsets from a custom consumer group


mwylde:
Currently, the Kafka consumer has two options for initializing offsets on a new pipeline: `earliest` and `latest` which have the same behavior as the corresponding values for the Kafka consumer 'auto.offset.reset` config.

We also allow users to set a custom consumer group, in which case we will commit offsets to that (which can be used for progress tracking), but we do not have an option to load offsets from an existing group on startup.

We should add a third option, `group` to the offset enum here: https://github.com/ArroyoSystems/arroyo/blob/0eb9d9eb81b48cd1a210c99cd9697e1cc9860ec3/connector-schemas/kafka/table.json#L19
which will then be handled in the consumer creation code:
https://github.com/ArroyoSystems/arroyo/blob/0eb9d9eb81b48cd1a210c99cd9697e1cc9860ec3/arroyo-worker/src/connectors/kafka/source/mod.rs#L209-L215
such that if it's set, we'll attempt to load the offsets from the group, or fail if there are no existing offsets.

mwylde:
Fixed by @harshit2283 in #480

PR #480, opened 2024-01-07
fix: offsets picking when using custom group

harshit2283:
Based on https://github.com/ArroyoSystems/arroyo/issues/469
Adds a new enum `group` which ensures that stored offsets are picked.

### DIFF
--- a/arroyo-worker/src/connectors/kafka/source/mod.rs
+++ b/arroyo-worker/src/connectors/kafka/source/mod.rs
@@ -190,7 +190,6 @@ where
 
         // did we restore any partitions?
         let has_state = !state.is_empty();
-        let offset_mode = self.offset_mode.get_offset();
 
         let state: HashMap<i32, KafkaState> = state.iter().map(|s| (s.partition, **s)).collect();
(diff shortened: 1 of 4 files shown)


arroyo-worker/src/connectors/kafka/source/mod.rs
```diff
@@ -207,12 +208,14 @@ where
                         .get(&p.id())
                         .map(|s| Offset::Offset(s.offset))
                         .unwrap_or_else(|| {
-                            if has_state {
+                            if offset_mode == Offset::Stored {
+                                offset_mode
```
mwylde, line 212: I think that this should still go into the else block here. If `has_state` is true, that means that we are restoring an existing checkpoint and we should now be managing offsets based on our state. If we find a partition that doesn't exist in our state, that means it was added since our last checkpoint. In that case, I think the correct behavior is to always read it from the beginning.

harshit2283: apologies, this was an oversight on my part. I was thinking from the consumer point of view where I would just want to start from an x offset but the correct way would be to re-create the pipeline itself if offset differs from checkpoint.

mwylde (approved):
Thanks, looks great!

PR #481, opened 2024-01-07
Add avro support for arrow deserializer

mwylde:
This PR adds support for deserializing from avro (including all of the currently supported flavors) into arrow.

For structured deserialization, we roundtrip through JSON and use the existing JSON infrastructure, as we do today.


jacksonrnewhouse (approved):
looks good!

PR #483, opened 2024-01-10
Add Async UDFs

jbeisen:
Add a new 'AsyncMapOperator' that applies an async udf to its input.

Async UDFs use the existing interface for defining the UDF and are called in SQL the same as non-async UDFs. Options are specified as a TOML configuration block in the special comment with the dependencies.

Here's an example of an Async UDF:

```
/*
[dependencies]
tokio = { version = "1", features = ["full"] }
tokio-postgres = "0.7"

[udfs]
async_results_ordered = true
async_timeout_seconds = 5
async_max_concurrency = 10

*/

use tokio_postgres::{NoTls, Error};


pub async fn user_name_from_id(id: i64) -> String {
    let conn_str = "host=localhost user=arroyo password=arroyo dbname=my_db";

    let (client, connection) = tokio_postgres::connect(conn_str, NoTls).await.unwrap();

    tokio::spawn(async move {
        if let Err(e) = connection.await {
            println!("connection error: {}", e);
        }
    });

    let rows = client
        .query("SELECT name FROM users WHERE id = $1", &[&id])
        .await
        .unwrap();

    return if let Some(row) = rows.get(0) {
        row.get(0)
    } else {
        format!("Unknown-{}", id)
    };
}
```

which can be used in a query like this:
```
select user_name_from_id(person.id) as name from nexmark where person is not null;
```

I've left these things for a future commit:
- Retries: if the UDF panics we should allow retrying it certain number of times
- Async UDF Context: A trait that users can implement to perform setup and shutdown behavior, that each UDF invocation will have access to


arroyo-datastream/src/lib.rs
```diff
@@ -1622,9 +1636,75 @@ impl Program {
...
+                                Ok(#out_t {
+                                    #null_output_assignments
+                                })
+                            );
+                        }
+                    } else {
+                        quote! ()
+                    };
+
+                    let udf_wrapper = quote!({
```
jacksonrnewhouse, line 1676: I think there's too much logic here, and as a consequence too many fields on AsyncMapOperator. I think it'd be better to generate more of the code in arroyo-sql.

arroyo-sql/src/operators.rs
```diff
@@ -159,6 +160,35 @@ impl CodeGenerator<ValuePointerContext, StructDef, syn::Expr> for UnnestProjecti
...
+
+impl AsyncUdfProjection {
+    pub fn output_struct(&self) -> StructDef {
+        self.expression_type(&ValuePointerContext::new())
+    }
+}
+
+impl CodeGenerator<ValuePointerContext, StructDef, syn::Expr> for AsyncUdfProjection {
+    fn generate(&self, _input_context: &ValuePointerContext) -> syn::Expr {
+        unreachable!()
```
jacksonrnewhouse, line 178: This seems like not the approach. The idea behind the `CodeGenerator<Context, OutputValue, OutputType: ToTokens>` trait is that you specify how to produce some syn type ToTokens, given a Context and an OutputValue. I think you want the code after `  RecordTransform::AsyncUdfProjection(a) => {` to be specified in the generate() method of a CodeGenerator implementation.

arroyo-worker/src/operators/async_map.rs
```diff
@@ -0,0 +1,170 @@
...
+    }
+
+    async fn handle_checkpoint(&mut self, _: &CheckpointBarrier, ctx: &mut Context<InKey, OutT>) {
+        let mut gs = ctx
+            .state
+            .get_global_keyed_state::<String, Record<InKey, InT>>('a')
+            .await;
+
+        for (i, record) in self.inputs.iter().filter_map(|x| x.clone()).enumerate() {
+            let key = format!("{},{}", ctx.task_info.task_index, i);
```
jacksonrnewhouse, line 162: The key should just be a tuple. (usize, usize) is a valid key here.

jacksonrnewhouse (changes requested):
a couple stylistic things I'd like to see addressed.

mwylde:
This query is panicking in the planner:

```sql
create table logs (
    ip TEXT,
    city TEXT GENERATED ALWAYS AS (get_city(ip))
) with (
    connector = 'sse',
    endpoint = 'http://127.0.0.1:9563/sse',
    format = 'json'
);

select city
from logs;
```

with 

```
2024-01-11T23:52:08.418511Z ERROR arroyo_server_common: panicked at arroyo-sql/src/expressions.rs:269:21:
async udf appeared in a non-projection context panic.file="arroyo-sql/src/expressions.rs" panic.line=269 panic.column=21
2024-01-11T23:52:08.418876Z  WARN arroyo_api::pipelines: failed to generate SQL program
```

mwylde:
Another panic if you try to group by an async udf:

```sql
select 
    get_city(logs.ip) as city, 
    count(*)
from logs
group by 1;
```

```
2024-01-12T00:06:33.748816Z ERROR arroyo_server_common: panicked at arroyo-sql/src/expressions.rs:269:21:
async udf appeared in a non-projection context panic.file="arroyo-sql/src/expressions.rs" panic.line=269 panic.column=21
2024-01-12T00:06:33.749935Z  WARN arroyo_api::pipelines: failed to generate SQL program
```

arroyo-macro/src/lib.rs
```diff
@@ -502,6 +511,41 @@ fn impl_stream_node_type(
...
+                Some(f) = self.#futures_ident.next() => {
+                    let (id, value) = f;
+
+                    match value {
+                        Ok(value) => {
+                            let index = self.inputs.len() - (self.next_id - id);
+                            let input = self.inputs[index].clone().unwrap();
+                            ctx.collector.collect(
+                                Record {
+                                    timestamp: input.timestamp.clone(),
```
mwylde, line 528: this clone is a no-op because SystemTime implements Copy

mwylde:
This is really great! I'm so excited to see how people use this.

arroyo-datastream/src/lib.rs
```diff
@@ -375,6 +375,19 @@ pub enum Operator {
...
+        name: String,
+        ordered: bool,
+        fn_name: String,
+        defs: String,
+        args: String,
+        null_output_assignments: String,
+        output_assignments: String,
+        null_handlers: String,
+        return_nullable: bool,
+        timeout_seconds: u64,
```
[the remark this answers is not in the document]
jbeisen: I don't think so because it gets deserialized from the toml in the UDF definition.

arroyo-rpc/src/lib.rs
```diff
@@ -190,3 +190,22 @@ pub fn error_chain(e: anyhow::Error) -> String {
         .collect::<Vec<_>>()
         .join(": ")
 }
+
+fn default_async_timeout_seconds() -> u64 {
+    10
+}
+
+fn default_async_max_concurrency() -> u64 {
+    10
```
mwylde, line 199: This seems very low to me. If an async udf takes just 50ms to run (pretty typical for a DB lookup) this limits us to 200 events/sec. I'd make this at least 100.

jacksonrnewhouse (approved):

PR #484, opened 2024-01-10
feat(redis): support Redis username and password in RedisConnector #457

### LINKED ISSUE #457: Support Redis username and password in RedisConnector
Right now if you want to connect to a Redis cluster that has a username and password you have to inline it into the connection string, e.g. `redis://username:password@host:ip`. It'd be nice to be able to specify this on the Redis connection. This'd be added as two optional properties [here](https://github.com/ArroyoSystems/arroyo/blob/master/connector-schemas/redis/connection.json#L11). I think there should probably a single username/password for the Clustered case as well. 

Then, to update the operators, you'll need to use a different IntoConnectionInfo, likely just ConnectionInfo. That would happen [here](https://github.com/ArroyoSystems/arroyo/blob/master/arroyo-worker/src/connectors/redis/sink.rs#L261-L268).

Will also need to pick the sql parameters for these options.

chenquan:
resolved #457

### DIFF
--- a/arroyo-api/src/pipelines.rs
+++ b/arroyo-api/src/pipelines.rs
@@ -54,6 +54,7 @@ use crate::types::public::{PipelineType, RestartMode, StopMode};
 use crate::{connection_tables, to_micros};
 use crate::{handle_db_error, optimizations, AuthData};
 use create_pipeline_req::Config::Sql;
+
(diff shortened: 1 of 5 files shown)


mwylde (approved):
This is great! Made a few changes to fit into how we're now handling usernames and passwords (with env var support).

Thanks for the contribution!

PR #492, opened 2024-01-16
Have async operators flush on close(). Emit watermarks in order.

jacksonrnewhouse:
This addresses a couple of async udf issues, namely
* Flush the futures when EndOfData is encountered.
* Hold operator watermarks until all records received before them have been emitted.
* Pass down the final_message to on_close so operators can distinguish between Stop and EndOfData.

### DIFF
--- a/arroyo-worker/src/operators/async_map.rs
+++ b/arroyo-worker/src/operators/async_map.rs
@@ -1,8 +1,8 @@
 use crate::engine::{Context, StreamNode};
 use arroyo_macro::process_fn;
 use arroyo_rpc::grpc::TableDescriptor;
-use arroyo_types::Record;
 use arroyo_types::{CheckpointBarrier, Data, Key, UdfContext};
+use arroyo_types::{Message, Record};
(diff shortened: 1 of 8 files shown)


arroyo-worker/src/operators/async_map.rs
```diff
@@ -183,13 +199,69 @@ impl<
...
+                self.on_collect(id, ctx).await;
+            }
+            Err(_e) => {
+                if self.futures.is_ordered() {
+                    unimplemented!(
+                        "Ordered Async UDF timed out, currently panic to preserve ordering"
+                    );
+                }
+                warn!("Unordered Async UDF timed out, retrying");
+                self.futures.push_back((self.udf)(
```
jbeisen, line 257: Is there anything to prevent an infinite retry loop?

jacksonrnewhouse: There isn't currently, we should build out the options here in the next version.

jbeisen (approved):

PR #508, opened 2024-01-23
Record batch session windows

jacksonrnewhouse:
This implements Session Windows in the record batch ecosystem. The algorithm works as follows:

Each group by key is tracked separately in a KeyComputingHolder. This contains the configuration, potentially an ActiveSession, and a sorted map from minimum time to a Vec of RecordBatches. RecordBatches are always sorted by time order, and there is an invariant that if the active session is non-empty the data_end + gap is less than the first batches_by_start_time. This allows us to immediately flush the active session when the watermark advances.

### Adding a Batch
All batches are first added to the buffered batches, and then fill_active_session() is called to enforce the invariant. fill_active_session() compares the end of the active session with the fist batch in the buffer, and if they are within gap, adds as much of the batch as is definitely in the session. This is done in a loop until the invariant is reached.

### Advancing the watermark
When the watermark is advanced, two things may happen: The active session may be finished and a new active session may begin. These are done in a loop, first checking if there have been no events in the active session since `watermark - gap`. Then, if the first buffered event is within `gap` of the watermark, it will definitely be present in the next active session. As such, it is used to start a new session. When a new session is initialized, we call `fill_active_session()` to enforce the invariant.

This change also refactors the production of these windowed aggregates, so more of the SQL manipulation happens in planning, rather in the from_config() method of the operators.

### DIFF
--- a/arroyo-api/src/pipelines.rs
+++ b/arroyo-api/src/pipelines.rs
@@ -55,7 +55,7 @@ use crate::{connection_tables, to_micros};
 use crate::{handle_db_error, AuthData};
 use create_pipeline_req::Config::Sql;
 
-const DEFAULT_CHECKPOINT_INTERVAL: Duration = Duration::from_secs(2);
+const DEFAULT_CHECKPOINT_INTERVAL: Duration = Duration::from_secs(10);
 
 async fn compile_sql<'e, E>(
     query: String,
--- a/arroyo-api/src/pipelines.rs
+++ b/arroyo-api/src/pipelines.rs
@@ -55,7 +55,7 @@ use crate::{connection_tables, to_micros};
 use crate::{handle_db_error, AuthData};
 use create_pipeline_req::Config::Sql;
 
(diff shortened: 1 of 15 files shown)


arroyo-api/src/pipelines.rs
```diff
@@ -56,7 +56,7 @@ use crate::{connection_tables, to_micros};
 use crate::{handle_db_error, AuthData};
 use create_pipeline_req::Config::Sql;
 
-const DEFAULT_CHECKPOINT_INTERVAL: Duration = Duration::from_secs(10);
+const DEFAULT_CHECKPOINT_INTERVAL: Duration = Duration::from_secs(2);
```
mwylde, line 59: do we want to commit this?

mwylde (approved):

arroyo-api/src/pipelines.rs
jacksonrnewhouse: probably not

PR #511, opened 2024-01-25
Generate watermarks and set event time using expressions

jbeisen:
Modify the periodic watermark generator to accept a physical expressions that it uses to calculate the watermark. The expression can be set via a virtual field in the create table statement, or it defaults to a max lateness expression. The operator is inserted via a user-defined logical node.

 Add an expression to the projection in the SourceRewriter to set the event time if present.

### DIFF
--- a/arroyo-df/src/lib.rs
+++ b/arroyo-df/src/lib.rs
@@ -384,23 +384,19 @@ impl ArroyoSchemaProvider {
             .get_table(name)
             .ok_or_else(|| DataFusionError::Plan(format!("Table {} not found", name)))?;
 
-        let mut table_fields = vec![];
-        let fields_from_table = table.get_fields();
-        for field in fields {
-            let t_field = fields_from_table
-                .iter()
-                .find(|t_field| t_field.name() == field.name())
-                .ok_or_else(|| {
-                    return DataFusionError::Plan(format!(
-                        "Field {} not found in table {}",
-                        field.name(),
-                        name
-                    ));
-                })?;
-            table_fields.push(t_field.clone());
-        }
+        let fields = table
+            .get_fields()
+            .iter()
+            .filter_map(|field| {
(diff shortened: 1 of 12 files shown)


arroyo-df/src/plan_graph.rs
```diff
@@ -351,6 +328,73 @@ impl Planner {
                     node_mapping.insert(node_index, sink_index);
                     sink_index
                 }
+                crate::LogicalPlanExtension::WatermarkNode(w) => {
```
jacksonrnewhouse, line 331: longer names for these variants are more readable, e.g. `watermark_node`

arroyo-df/src/source_rewriter.rs
```diff
@@ -1,39 +1,59 @@
...
+                        if field.name() == &watermark_field {
+                            return Some(expression.clone());
+                        }
+                    }
+                    None
+                })
+                .ok_or_else(|| {
+                    DataFusionError::Plan(format!("Watermark field {} not found", watermark_field))
+                })?,
+            None => Expr::BinaryExpr(BinaryExpr {
```
jacksonrnewhouse, line 43: Might merit a comment explaining the default behavior

jacksonrnewhouse (approved):

arroyo-df/src/lib.rs
```diff
@@ -981,6 +988,43 @@ impl TreeNodeRewriter for QueryToGraphVisitor {
                 };
                 Ok(interred_plan.clone())
             }
+            LogicalPlan::Extension(extension) => {
+                let watermark_node = extension
+                    .node
+                    .as_any()
+                    .downcast_ref::<WatermarkNode>()
+                    .unwrap()
```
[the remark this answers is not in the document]
jbeisen: Done, returning a `DataFusionError`

PR #514, opened 2024-01-29
Add a new binary crate 'arroyo-bin' to package the entire system

mwylde:
This PR adds a new crate, "arroyo-bin" that packages the entire Arroyo system (api/controller/worker). This will allow for deployment via a single, self-contained binary.

```
$ arroyo-bin
Usage: arroyo-bin <COMMAND>

Commands:
  api         Starts an Arroyo API server
  controller  Starts an Arroyo Controller
  cluster     Starts a complete Arroyo cluster
  worker      Starts an Arroyo worker
  help        Print this message or the help of the given subcommand(s)

Options:
  -h, --help     Print help
  -V, --version  Print version
```

This affects the dev process a bit:
* Instead of running the api and the controller, run arroyo-bin (the `cluster` option, which will start both controller and api, or just one of the services)
* The worker also now runs from this binary, so there's no need to separately build the worker any more. For the process scheduler, the controller will use its own binary for the worker. This means that a controller compiled as debug mode will use the debug worker, and likewise for release.

As part of this work I've introduced a more sophisticated shutdown system (in arroyo-server-common/src/shutdown.rs) which handles the two phases of clean shutdowns in an async application:
1. Detecting that the system _should_ shut down, due to either a signal or because a subcomponent shut down or crashed)
2. Notifying the whole tree of tasks to stop via async cancellation, then waiting for all tasks to stop

I've also fixed and re-enabled the integration tests.


Cargo.toml
```diff
@@ -21,7 +22,7 @@ members = [
     "arroyo-types",
     "arroyo-worker",
     "copy-artifacts",
-    "integ",
+    "integ", "arroyo-bin",
```
jacksonrnewhouse, line 25: Put this on its own line?

jacksonrnewhouse (approved):
Do we need to update docs?

mwylde:
Yes, we'll need to update dev docs. I think it would make sense to do that when we're ready to merge into master.

PR #519, opened 2024-02-01
refactor filesystem source.

jacksonrnewhouse:
Adds JSON support to the filesystem source.

### DIFF
--- a/arroyo-connectors/src/filesystem/source.rs
+++ b/arroyo-connectors/src/filesystem/source.rs
@@ -1,11 +1,16 @@
+use std::collections::hash_map::DefaultHasher;
 use std::collections::HashMap;
 use std::future::ready;
+use std::hash::{Hash, Hasher};
 use std::time::SystemTime;
 
 use anyhow::Result;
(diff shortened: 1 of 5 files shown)


arroyo-connectors/src/filesystem/source.rs
```diff
@@ -160,15 +176,50 @@ impl FileSystemSourceFunc {
...
         &mut self,
         storage_provider: &StorageProvider,
         path: String,
+        out_schema: SchemaRef,
     ) -> Result<Box<dyn Stream<Item = Result<RecordBatch, UserError>> + Unpin + Send>, UserError>
     {
         match &self.format {
-            Format::Json(_) => Err(UserError::new("unsupported format:", "json")),
-            Format::Avro(_) => Err(UserError::new("unsupported format:", "avro")),
+            Format::Json(_) => Err(UserError::new("unsupported format", "json")),
```
mwylde, line 221: It seems like if you end up here this is due to a coding error, so maybe these should be panics?

mwylde (approved):

PR #521, opened 2024-02-02
Implement unnest in arrow land

mwylde:
for example:

```sql
select unnest(a.list) + 5
from source;
```

currently, this is implemented with the same restrictions as on master:
* only supported in select statements (projections)
* only one unnest per expression and only one column may have an unnest


arroyo-df/src/lib.rs
```diff
@@ -158,12 +156,14 @@ impl ArroyoSchemaProvider {
                 #[allow(deprecated)]
                 ScalarUDF::new(
                     "unnest",
-                    &Signature::any(1, Volatility::Immutable),
+                    &Signature::any(1, Volatility::Volatile),
```
jacksonrnewhouse, line 159: I think this is volatile so that DF doesn't replace a constant term. If that's the case, wanna leave a comment to that affect?

arroyo-df/src/rewriters.rs
```diff
@@ -0,0 +1,422 @@
...
+    type N = LogicalPlan;
+
+    fn mutate(&mut self, node: Self::N) -> DFResult<Self::N> {
+        let LogicalPlan::Projection(projection) = &node else {
+            if node.expressions().iter().any(|e| {
+                let e = Self::split_unnest(e.clone());
+                e.is_err() || e.unwrap().1.is_some()
+            }) {
+                return Err(DataFusionError::Plan(
+                    "unnest is only supported in SELECT statements".to_string(),
```
jacksonrnewhouse, line 320: Maybe print something that indicates where it was found instead? Possibly just the debug for the logical plan?

arroyo-df/src/rewriters.rs
```diff
@@ -0,0 +1,422 @@
...
+            .expr
+            .clone()
+            .into_iter()
+            .enumerate()
+            .map(|(i, expr)| {
+                let (expr, opt) = Self::split_unnest(expr)?;
+                let typ = if let Some(e) = opt {
+                    if let Some(prev) = unnest.replace((e, i)) {
+                        if &prev != unnest.as_ref().unwrap() {
+                            return Err(DataFusionError::Plan("Project contains multiple unnests, which is not currently supported".to_string()));
```
jacksonrnewhouse, line 337:
```suggestion
                            return Err(DataFusionError::Plan("Projection contains multiple unnests, which is not currently supported".to_string()));
```

arroyo-rpc/proto/api.proto
```diff
@@ -393,6 +393,27 @@ enum EdgeType {
...
+  // json-encoded
+  string schema = 2;
+}
+
+message UnnestExecNode {
+  // json-encoded
+  string schema = 1;
+}
+
+message ArroyoExecNode {
```
jacksonrnewhouse, line 408: This is nice.

jacksonrnewhouse (approved):

arroyo-df/src/lib.rs
mwylde: That's correct, I'll add a comment

PR #523, opened 2024-02-03
Load dynamically linked UDF libraries at runtime

jbeisen:
Build a wrapper library for each UDF that gets loaded by the worker at runtime. The library code is generated at pipeline creation time, and the worker calls the UDF via the foreign function interface.

This change also brings back the compiler service, converting it to a library and launching the service with arroyo-bin.


arroyo-storage/src/lib.rs
```diff
@@ -805,3 +805,11 @@ mod tests {
         );
     }
 }
+
+pub fn dylib_name(name: &str) -> String {
```
mwylde, line 809: does this belong in arroyo-storage? maybe arroyo-types or arroyo-rpc?

arroyo-df/src/physical.rs
```diff
@@ -156,7 +154,119 @@ pub fn window_scalar_function() -> ScalarUDF {
...
+        let signature = Signature::exact(
+            config
+                .arg_types
+                .iter()
+                .map(|arg| arrow_type_to_data_type(arg))
+                .collect(),
+            Volatility::Volatile,
+        );
+
+        let artifact_url =
```
mwylde, line 188: I think the full path needs to come from the job definition and should be set at UDF compile time. Imagine that I start a pipeline, then change my artifact directory—that would break all future runs for existing jobs.

arroyo-df/src/physical.rs
```diff
@@ -156,7 +154,119 @@ pub fn window_scalar_function() -> ScalarUDF {
...
+}
+
+impl Debug for UdfDylib {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("UdfDylib").finish()
+    }
+}
+
+impl UdfDylib {
+    pub async fn new(name: &str, config: &DylibUdfConfig) -> Self {
```
mwylde, line 178: I wouldn't expect something called "new" to go and get data from a remote environment and write it to the file system. possibly "download" or something?

arroyo-df/src/physical.rs
```diff
@@ -156,7 +154,119 @@ pub fn window_scalar_function() -> ScalarUDF {
...
+        Ok(self.return_type.clone())
+    }
+
+    fn invoke(&self, args: &[ColumnarValue]) -> DFResult<ColumnarValue> {
+        let args = args
+            .iter()
+            .map(|arg| match arg {
+                ColumnarValue::Array(array) => to_ffi(&(array.to_data())).unwrap(),
+                ColumnarValue::Scalar(_) => {
+                    unimplemented!("Scalar args not supported")
```
mwylde, line 239: We probably will need to support this or prevent it from planning (for example "select udf(5) from impulse;" will trigger this case)

arroyo-df/src/types.rs
```diff
@@ -117,6 +89,65 @@ pub fn rust_to_arrow(typ: &Type) -> Option<NullableType> {
...
+        DataType::UInt32 => ArrowTypeEnum::Uint32(EmptyMessage {}),
+        DataType::Float16 => ArrowTypeEnum::Float16(EmptyMessage {}),
+        DataType::Float32 => ArrowTypeEnum::Float32(EmptyMessage {}),
+        DataType::Float64 => ArrowTypeEnum::Float64(EmptyMessage {}),
+        DataType::Date32 => ArrowTypeEnum::Date32(EmptyMessage {}),
+        DataType::Date64 => ArrowTypeEnum::Date64(EmptyMessage {}),
+        DataType::Binary => ArrowTypeEnum::Binary(EmptyMessage {}),
+        DataType::LargeBinary => ArrowTypeEnum::LargeBinary(EmptyMessage {}),
+        DataType::LargeUtf8 => ArrowTypeEnum::LargeUtf8(EmptyMessage {}),
+        _ => panic!("Unsupported DataType"),
```
mwylde, line 113: We shouldn't panic here unless we guarantee we can never hit this case

arroyo-df/src/types.rs
```diff
@@ -117,6 +89,65 @@ pub fn rust_to_arrow(typ: &Type) -> Option<NullableType> {
...
+        Some(ArrowTypeEnum::Binary(_)) => DataType::Binary,
+        Some(ArrowTypeEnum::LargeBinary(_)) => DataType::LargeBinary,
+        Some(ArrowTypeEnum::Date32(_)) => DataType::Date32,
+        Some(ArrowTypeEnum::Date64(_)) => DataType::Date64,
+        Some(ArrowTypeEnum::Duration(_)) => DataType::Duration(TimeUnit::Second),
+        Some(ArrowTypeEnum::Timestamp(_)) => DataType::Timestamp(TimeUnit::Second, None),
+        Some(ArrowTypeEnum::Time32(_)) => DataType::Time32(TimeUnit::Second),
+        Some(ArrowTypeEnum::Time64(_)) => DataType::Time64(TimeUnit::Second),
+        Some(ArrowTypeEnum::Interval(_)) => DataType::Interval(IntervalUnit::YearMonth),
+        _ => panic!("Unsupported ArrowType"),
```
mwylde, line 147: And similarly here

mwylde (changes requested):
Really exciting that this is working!

Some notes from testing:

1. We've lost the logic that re-adds "pub" to the functions, so that's not required

2. This UDF fails to compile in the wrapper code (from the analytics parsing tutorial):

```
/*
[dependencies]
regex = "1"
serde_json = "1"
chrono = "0.4"
*/
 
use regex::Regex;
use serde_json::json;
use chrono::DateTime;
use std::str::FromStr;
 
pub fn parse_log(s: String) -> Option<String> {
    // use OnceLock to prevent re-compilation of the regexes on every request
    use std::sync::OnceLock;
    static REGEX: OnceLock<Regex> = OnceLock::new();
    let regex = REGEX.get_or_init(|| {
        Regex::new(r#"([0-9a-f\.:]+) ([\S\-]+) ([\S\-]+) \[([^\]]+)\] "(.*)" ([0-9]{3}) ([0-9]*).*"#).unwrap()
    });
 
    // try to parse the log, returning None if it fails
    let captures = regex.captures(&s)?;
 
    // try parsing the timestamp using the chrono library
    let date = DateTime::parse_from_str(captures.get(4).unwrap().as_str(), "%d/%b/%Y:%H:%M:%S %z").ok()?;
 
    // pull out the fields
    Some(json!({
        "host": captures.get(1).unwrap().as_str().to_string(),
        "ident": captures.get(2).unwrap().as_str().to_string(),
        "authuser": captures.get(3).unwrap().as_str().to_string(),
        "date": date.to_rfc3339(),
        "request": captures.get(5).unwrap().as_str().to_string(),
        "status": u16::from_str(captures.get(6).unwrap().as_str()).unwrap(),
        "bytes": usize::from_str(captures.get(7).unwrap().as_str()).unwrap(),
    }).to_string())
}
```

3. Calling a UDF `f(T)` (where T is any non-option type) on nullable inputs is causing a panic, when it should bypass the UDF and return NULL: 
image

Also, don't worry about docker for now—we're going to need to completely rework that anyways

arroyo-df/src/physical.rs
```diff
@@ -156,7 +154,119 @@ pub fn window_scalar_function() -> ScalarUDF {
     )
 }
 
-impl FunctionRegistry for EmptyRegistry {
+#[derive(WrapperApi)]
+struct UdfDylibInterface {
+    run: unsafe extern "C" fn(
+        args: Vec<(FFI_ArrowArray, FFI_ArrowSchema)>,
```
mwylde, line 160: I don't think Vec is a safe interface for FFI (i.e., there's no guarantee that memory layout will remain the same across rust versions). Conveniently, Rust just published new docs on ABI compatibility: https://doc.rust-lang.org/stable/std/primitive.fn.html#abi-compatibility.

arroyo-compiler-service/src/lib.rs
```diff
@@ -0,0 +1,226 @@
...
+            r#"
+[package]
+name = "udf_wrapper"
+version = "0.1.0"
+edition = "2021"
+
+
+[dependencies]
+udf = {{ path = "../udf" }}
+arrow = {{ version = "50.0.0", features = ["ffi"] }}
```
jacksonrnewhouse, line 127: This could probably be smaller, I think just arrow-array and arrow-data. Building with --release for me went from 36 seconds to 6.

arroyo-compiler-service/src/lib.rs
```diff
@@ -0,0 +1,226 @@
...
+        };
+
+        let name = udf_crate.name.clone();
+        self.write_udf_crate(udf_crate)
+            .await
+            .map_err(|e| Status::internal(format!("Writing UDFs failed: {}", e)))?;
+
+        let udf_build_dir = self.build_dir.join("udfs_dir").join("udf_wrapper");
+        let output = Command::new("cargo")
+            .current_dir(&udf_build_dir)
```
jacksonrnewhouse, line 165: Can this build with --release?

arroyo-compiler-service/src/lib.rs
```diff
@@ -0,0 +1,226 @@
...
+pub fn to_millis(time: SystemTime) -> u64 {
+    time.duration_since(UNIX_EPOCH).unwrap().as_millis() as u64
+}
+
+pub fn from_millis(ts: u64) -> SystemTime {
+    UNIX_EPOCH + Duration::from_millis(ts)
+}
+
+pub async fn start_service(service: CompileService) {
+    let shutdown = Shutdown::new("compiler-service");
```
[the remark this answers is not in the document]
jbeisen: I remove this and am using the shutdown to spawn the task in `start_control_plane`. Let me know if that's ok.

arroyo-controller/src/lib.rs
```diff
@@ -634,29 +641,201 @@ impl ControllerServer {
...
+
+    prettyplease::unparse(&parse_quote! {
+        use arrow::array;
+        use arrow::array::Array;
+        use arrow::ffi::{FFI_ArrowArray, FFI_ArrowSchema, from_ffi, to_ffi};
+        use udf;
+
+        #[no_mangle]
+        pub extern "C" fn run(args: Vec<(FFI_ArrowArray, FFI_ArrowSchema)>) -> (FFI_ArrowArray, FFI_ArrowSchema) {
+            let mut results = Vec::new();
```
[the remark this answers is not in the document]
jbeisen: Done. It required having some branching logic to handle strings because the `StringBuilder` has a different signature than `PrimitiveBuilder`.

arroyo-controller/src/lib.rs
```diff
@@ -634,29 +641,201 @@ impl ControllerServer {
...
+            quote!(Decimal128Array)
+        }
+        DataType::Decimal256(_, _) => {
+            quote!(Decimal256Array)
+        }
+        DataType::Map(_, _) => {
+            quote!(MapArray)
+        }
+        _ => {
+            panic!("Unsupported data type: {:?}", data_type);
```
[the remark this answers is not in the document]
jbeisen: Changed this to return a result

arroyo-df/src/physical.rs
jbeisen: Done. I'm now getting getting the storage provider with `StorageProvider::for_url("/")`. Is that what you had in mind?

arroyo-df/src/physical.rs
jbeisen: Renamed to `init` and added a comment.

arroyo-df/src/physical.rs
jbeisen: I've implemented this with a function `scalar_value_to_array`.

arroyo-compiler-service/src/lib.rs
jbeisen: Unfortunately the functions `from_ffi` and `to_ffi` are only exported by the `arrow` crate.

arroyo-df/src/physical.rs
mwylde: This is what StorageProvider::get_url is for. See the example here: https://github.com/ArroyoSystems/arroyo/blob/master/copy-artifacts/src/main.rs#L29.

arroyo-compiler-service/src/lib.rs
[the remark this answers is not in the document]
mwylde: We do need involve the shutdown system here, as that's responsible for cleanly shutting down the entire cluster. If something goes wrong on compiler service startup, that needs to trigger a shutdown of the rest of the system, and similarly if another part of the system errors (or the user initiates a shutdown) that needs to trigger a shutdown of the compiler service.

See how this works in the worker, api server, and controller — you should be able to use the same pattern here.

arroyo-compiler-service/src/lib.rs
jbeisen: Yes!

arroyo-df/src/types.rs
jbeisen: I've made sure the planning surfaces the error for unsupported types.

mwylde (approved):

mwylde:
It looks like it doesn't work out of the box running ./aroyo-bin cluster — you still need to manually pass in the compiler address.

jbeisen:
@mwylde I've changed it to make arroyo-bin pass the compiler address to the controller

mwylde (approved):

PR #524, opened 2024-02-05
Filesystem Sink + Two Phase Commits

jacksonrnewhouse:
Ready for review. I moved ConnectionSchema into OperatorConfig and made it non-optional in the APIs. I can revert either of these changes, but it did allow me access to the ArroyoSchemaRef on Operator creation.


mwylde (commented):
This generally looks good. I'm not sure it's wise to rely on the connection schema the way it's done here for the parquet writer. Generally speaking, the connection schema is an input to the system—it's what the user specified as the schema, but not necessarily the schema of the actual data. The source of truth for that is the graph edges.

mwylde (approved):

PR #533, opened 2024-02-17
Windowed Joins

jacksonrnewhouse:
This implements support for windowed joins, e.g. joining against two inputs that were previously windowed by the same window. Unfortunately the join exec is stateful, so we can't reuse the PhysicalExec across multiple joins. This required passing storing the Registry on InstantJoin, which motivated changing the with_config() signature from Arc<dyn FunctionRegistry> to Arc<Registry>, which will be Send and Sync. I also needed to move the Registry to arroyo-operator, which I don't love.

Test query:

```
CREATE TABLE impulse WITH(
  connector = 'impulse',
  event_rate = '100000000000',
  event_time_interval = '1',
  message_count = '500000000'
);
  

SELECT even_rows, all_rows FROM 
(SELECT count(*) as even_rows, hop(interval '1 second', '10 second') as sliding_window, counter % 100000 as counter_mod FROM impulse where counter % 2 = 0 GROUP BY 2,3) evens
JOIN (SELECT count(*) as all_rows, hop(interval '1 second', '10 second') as sliding_window, counter % 100000 as counter_mod FROM impulse GROUP BY 2,3) b on evens.sliding_window.start = b.sliding_window.start and b.counter_mod = evens.counter_mod

WHERE even_rows = b.counter_mod
```


crates/arroyo-operator/Cargo.toml
```diff
@@ -18,6 +18,7 @@ ahash = "0.8"
 async-trait = "0.1.68"
 bincode = "2.0.0-rc.3"
 datafusion = "35.0"
+datafusion-common = "35.0"
```
mwylde, line 21: Do we need this since we already have datafusion as a dependency?

mwylde (approved):

crates/arroyo-operator/Cargo.toml
jacksonrnewhouse: No, can use datafusion::common.

PR #534, opened 2024-02-17
Use hash of UDF definition for saving libraries

jbeisen:
Compute a hash of the UDF definition and use it as the file name of the library. Check if the library exists to avoid building libraries multiple times.

Also make the compiler service only run 'cargo build' when needed, otherwise running 'cargo check'.

### DIFF
--- a/crates/arroyo-compiler-service/src/lib.rs
+++ b/crates/arroyo-compiler-service/src/lib.rs
@@ -8,7 +8,7 @@ use arroyo_rpc::grpc::{
 };
 
 use arroyo_storage::StorageProvider;
-use arroyo_types::{dylib_name, grpc_port, ports, ARTIFACT_URL_DEFAULT, ARTIFACT_URL_ENV};
+use arroyo_types::{dylib_name, grpc_port, ports};
 use serde_json::Value;
 use tokio::{process::Command, sync::Mutex};
 use tonic::{Request, Response, Status};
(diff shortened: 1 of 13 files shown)


crates/arroyo-df/src/lib.rs
```diff
@@ -232,6 +231,10 @@ impl ArroyoSchemaProvider {
         None
     }
 
+    pub fn get_used_udfs() -> Vec<String> {
```
mwylde, line 234: is this used?

crates/arroyo-compiler-service/src/lib.rs
```diff
@@ -139,18 +145,22 @@ impl CompilerGrpc for CompileService {
...
+                    &udf_build_dir
+                        .join("target/release/")
+                        .join(dylib_name("libudf_wrapper")),
+                )
+                .await?;
+
+                self.storage
+                    .put(req.dylib_path.clone(), dylib)
+                    .await
+                    .map_err(|e| Status::internal(format!("Failed to save dylib: {}", e)))?;
```
mwylde, line 160: I'd avoid "dylib" in user-facing errors like this, since it may not be obvious to users what that means

crates/arroyo-types/Cargo.toml
```diff
@@ -10,3 +10,4 @@ serde = { version = "1.0", features = ["derive"] }
 arrow = { workspace = true }
 arrow-array = { workspace = true }
 async-trait = "0.1.74"
+blake3 = "1.5.0"
```
mwylde, line 13: do we need to add another hash dependency or can we use a built-in hash?

mwylde (approved):

crates/arroyo-df/src/lib.rs
jbeisen: No, removed.

crates/arroyo-types/Cargo.toml
jbeisen: Changed it to use `std::hash` and `base64`

PR #543, opened 2024-02-27
add synthetic key if there is no join key.

jacksonrnewhouse:
The overall_tumbling_table fails without this change.


jacksonrnewhouse:
This is superseded by #552.

PR #552, opened 2024-03-04
Make key_indices optional to distinguish from an empty group by and no group by.

jacksonrnewhouse:
This makes key_indices in ArroyoSchema an Option<Vec<usize>>, as when a query groups by only a time window we need to shuffle all data together. This approach won out over trying to mutate the SQL graph to insert a dummy key.

### DIFF
--- a/crates/arroyo-connectors/src/kafka/sink/mod.rs
+++ b/crates/arroyo-connectors/src/kafka/sink/mod.rs
@@ -171,8 +171,8 @@ impl ArrowOperator for KafkaSinkFunc {
     async fn process_batch(&mut self, batch: RecordBatch, ctx: &mut ArrowContext) {
         let values = self.serializer.serialize(&batch);
 
-        if !ctx.in_schemas[0].key_indices.is_empty() {
-            let k = batch.project(&ctx.in_schemas[0].key_indices).unwrap();
+        if let Some(key_indices) = &ctx.in_schemas[0].key_indices {
+            let k = batch.project(key_indices).unwrap();
 
             // TODO: we can probably batch this for better performance
             for (k, v) in self.serializer.serialize(&k).zip(values) {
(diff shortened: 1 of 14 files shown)


crates/arroyo-connectors/src/kafka/sink/mod.rs
```diff
@@ -171,8 +171,10 @@ impl ArrowOperator for KafkaSinkFunc {
     async fn process_batch(&mut self, batch: RecordBatch, ctx: &mut ArrowContext) {
         let values = self.serializer.serialize(&batch);
 
-        if !ctx.in_schemas[0].key_indices.is_empty() {
-            let k = batch.project(&ctx.in_schemas[0].key_indices).unwrap();
+        if ctx.in_schemas[0].key_indices.is_some() {
```
mwylde, line 174: Could this be an if let instead of if/unwrap?

crates/arroyo-operator/src/context.rs
```diff
@@ -170,13 +170,18 @@ pub struct ArrowCollector {
...
     record: &'a RecordBatch,
-    keys: &'a Vec<usize>,
+    keys: &'a Option<Vec<usize>>,
     qs: usize,
 ) -> impl Iterator<Item = (usize, RecordBatch)> + 'a {
     let mut buf = vec![0; record.num_rows()];
 
-    if !keys.is_empty() {
-        let keys: Vec<_> = keys.iter().map(|i| record.column(*i).clone()).collect();
+    if keys.is_some() {
```
mwylde, line 178: could this be an if-let?

mwylde (approved):

crates/arroyo-connectors/src/kafka/sink/mod.rs
jacksonrnewhouse: yep

crates/arroyo-operator/src/context.rs
jacksonrnewhouse: yep

PR #553, opened 2024-03-06
Reset ExecutionPlans

jacksonrnewhouse:
This uses a fork of DataFusion that allows for cheaply reseting execution plans, and calls it directly before calling execute(). Worth taking look at the commit in the fork: https://github.com/ArroyoSystems/arrow-datafusion/commit/18e1c45d48ab723fa17d29d97a658f028e1ff1e8. The biggest change there was in rewriting how the execute() method for the various joins function. Unlike all other ExecutionPlans it tries to cache work using OnceAsync, which can contaminate future calls. Additionally, it would defer calling execute() on some inputs, which will cause issues given how Arroyo reuses plans. I removed the use of OnceAsync and ensured that execute() is called on both left and right before the join returns.

### DIFF
--- a/k8s/arroyo/values.yaml
+++ b/k8s/arroyo/values.yaml
@@ -16,6 +16,7 @@ controller:
     tag: "dev"
   service:
     grpcPort: 9190
+    compilerPort: 9000
     adminPort: 8001
     httpPort: 80
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2816,8 +2816,7 @@ dependencies = [
 [[package]]
 name = "datafusion"
 version = "36.0.0"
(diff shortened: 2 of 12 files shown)


k8s/arroyo/values.yaml
```diff
@@ -13,9 +13,10 @@ controller:
   image:
     repository: ghcr.io/arroyosystems/arroyo
     pullPolicy: IfNotPresent
-    tag: "dev"
+    tag: "df_deltalake_docker"
```
mwylde, line 16: revert?

also—you can override this with a values.yaml file so it's not generally required to change the repo

mwylde (approved):

PR #560, opened 2024-03-12
Validate aggregate window selection, allow recalling windowing UDFs

jacksonrnewhouse:


crates/arroyo-df/src/plan.rs
```diff
@@ -171,43 +181,138 @@ impl TreeNodeRewriter for AggregateRewriter {
...
+                        if *existing_window != window {
+                            return Err(DataFusionError::Plan(
+                                "window expressions do not match".to_string(),
+                            ));
+                        }
+                        self.fields.insert(projection.schema.field(index).clone());
+                    } else {
+                        // If the input doesn't have an input window, we shouldn't be creating a window.
+                        return Err(DataFusionError::Plan(
+                            "can't call window function in a select statement".to_string(),
```
mwylde, line 260: As a user, what does this mean?

crates/arroyo-df/src/plan.rs
```diff
@@ -171,43 +181,138 @@ impl TreeNodeRewriter for AggregateRewriter {
...
+                            .transpose()
+                    })
+                    .collect::<DFResult<Vec<_>>>()?;
+                self.fields.clear();
+                for (index, window) in window_expressions {
+                    // if there's already a window they should match
+                    if let Some(existing_window) = &self.window {
+                        if *existing_window != window {
+                            return Err(DataFusionError::Plan(
+                                "window expressions do not match".to_string(),
```
mwylde, line 253: can this be explained more clearly?

mwylde (approved):
A few error messages that I think can be improved, but otherwise looks good!

crates/arroyo-df/src/plan.rs
jacksonrnewhouse: Gave it another shot. This is the error if you call one of our window UDFs without grouping by it.

mwylde: 👍 that's much more clear

PR #563, opened 2024-03-13
Implement window functions over already windowed inputs.

jacksonrnewhouse:
I needed to manually add a Sort to the logical plan. I think this might be because of the splitting of the prior aggregate and the window function between operators.


crates/arroyo-df/src/extension/window_fn.rs
```diff
@@ -0,0 +1,134 @@
...
+        let input_schema = input_schemas[0].clone();
+        let input_df_schema =
+            Arc::new(DFSchema::try_from(input_schema.schema.as_ref().clone()).unwrap());
+        let binning_function = planner.create_physical_expr(
+            &Expr::Column(Column::new_unqualified(TIMESTAMP_FIELD.to_string())),
+            &input_df_schema,
+        )?;
+        let binning_function_proto = PhysicalExprNode::try_from(binning_function)?;
+        let window_plan = planner.sync_plan(&self.window_plan)?;
+        info!(
```
mwylde, line 100: can these be downgraded to debug?

crates/arroyo-df/src/plan/window_fn.rs
```diff
@@ -0,0 +1,189 @@
...
+}
+
+impl TreeNodeRewriter for WindowFunctionRewriter {
+    type N = LogicalPlan;
+
+    fn mutate(&mut self, node: Self::N) -> DFResult<Self::N> {
+        let LogicalPlan::Window(window) = node else {
+            return Ok(node);
+        };
+        info!(
```
mwylde, line 43: ditto

mwylde (approved):

PR #568, opened 2024-03-19
Operator Life Cycle Overhaul

jacksonrnewhouse:
This reworks how our operators shutdown, both because of EndOfData messages and stopping checkpoints. It affects SourceOperators, ArrowOperators, and the controller.

### SourceOperator changes
Two main changes here. First, rather than having every source have a very similar match block on the control receiver, they all call `handle_control_message()`. This will call `flush_before_checkpoint()`, which each operator must implement. This did require moving some of the tracking structs into the main source struct.

Secondly, once the source finishes producing data, it now waits for either a stopping checkpoint or an immediate stop message. This lets checkpoints still be taken after some but not all sources are finished, where previously they would've hung.

### ArrowOperators
The previous `on_close()` is mainly replaced with an `on_end_of_data()` method, which is called when the operator is finishing. Mainly relevant for the `WatermarkGenerator`, which will insert the watermark that flushes downstream nodes. `on_close()` is still there, but now only used by the PreviewSink.

Like sources, it waits until an explicit signal to shut down, in this case receiving a SignalMessage::Shutdown. 

It also tracks whether the inputs were finished because they were stopped or because they don't have anymore data.

### Controller
The Finishing state has been overhauled. First, it is only transitioned to once all of the operators have finished processing their data. Then, it takes a stopping checkpoint which will commit any outstanding sinks.

Because we have several states that take stopping checkpoints I've factored that out into the `take_stopping_checkpoint()` method on the JobConfig. I'd appreciate a look if I translated all of the states correctly.


crates/arroyo-controller/src/states/mod.rs
```diff
@@ -381,6 +384,102 @@ impl<'a> JobContext<'a> {
...
+        loop {
+            match job_controller.checkpoint_finished().await {
+                Ok(done) => {
+                    debug!("checked checkpoint, got {}, job_controller.finished(): {}, final_checkpoint_started: {}", done, job_controller.finished(), final_checkpoint_started);
+
+                    if done && final_checkpoint_started {
+                        if job_controller.finished() {
+                            return Ok(StoppingCheckpointOutcome::SuccessfullyStopped);
+                        } else if !shutdown_started {
+                            info!("Starting shutdown");
```
mwylde, line 405: In the controller, logs should use structed logging like https://github.com/ArroyoSystems/arroyo/blob/master/crates/arroyo-controller/src/states/recovering.rs#L24 — this preserves context around job_id and makes it easier to build alerts / monitoring around the log lines

crates/arroyo-controller/src/states/mod.rs
```diff
@@ -381,6 +384,102 @@ impl<'a> JobContext<'a> {
...
+                        });
+                    }
+                }
+            }
+
+            match self
+                .rx
+                .recv()
+                .await
+                .expect("channel closed while receiving")
```
mwylde, line 441: we shouldn't panic in the controller

crates/arroyo-controller/src/states/rescaling.rs
```diff
@@ -12,59 +16,27 @@ impl State for Rescaling {
...
-                        ));
-                    }
-                }
-                JobMessage::ConfigUpdate(c) => {
-                    stop_if_desired_non_running!(self, &c);
-                }
-                _ => {
-                    // ignore other messages
-                }
+        match ctx.take_stopping_checkpoint().await {
```
mwylde, line 19: It looks like there might be some problems with rescaling. I got into a state where the the operators didn't shutdown cleanly, getting the controller stuck indefinitely in rescaling:
[rescaling.txt](https://github.com/ArroyoSystems/arroyo/files/14728293/rescaling.txt).

I also wasn't able to force stop out of rescaling — the pipeline was just wedged until I restarted the controller, then on recovery it went into CheckpointStopping where it also got stuck.

crates/arroyo-operator/src/operator.rs
```diff
@@ -232,20 +301,38 @@ async fn operator_run_behavior(
...
+                                        info!("finished stopping checkpoint in operator {}-{}, waiting for terminate",
+                                            ctx.task_info.operator_name, ctx.task_info.task_index);
+                                        operator_state = OperatorState::WaitingForTerminate;
                                     }
-                                    ControlOutcome::Finish => {
-                                        final_message = Some(SignalMessage::EndOfData);
-                                        break;
+                                    ControlOutcome::NoMoreData { end_of_data } => {
+                                        if operator_state != OperatorState::Running {
+                                            warn!("received no more data update in operator {}-{} while in state {:?}",
```
mwylde, line 313: I consistently see this warning when stopping pipelines — is it an actual issue, or should we make it a debug?

mwylde (changes requested):

crates/arroyo-controller/src/states/rescaling.rs
mwylde: I'm able to consistently reproduce this by doing two rescales in a row

PR #578, opened 2024-04-01
Add NATS source and sink connectors

gbto:
Opening the PR to get some feedbacks. This is a refactored version of https://github.com/gbto/arroyo/tree/feat/add-nats-source-connector which I've been using to run 6 pipelines in development for now one month on `arrow = { version = "0.9.0" }`. The source only support consuming from [NATS Streams](https://docs.nats.io/nats-concepts/jetstream/streams) and publishing to [NATS Subjects](https://docs.nats.io/nats-concepts/subjects). Both supports leverage connection profiles and tables configurations, with environment variable substitution for secrets. 

The tests are currently not implemented yet, neither are features that were note absolutely required to get it to work in our context of our architecture (e.g. autocomplete, schema validation and registry support, different semantics or deduplication..,). 

There still are a few flawed feature, in particular the way optional extra client configuration are parsed and the checkpointing in the sink connector. Comments on these 2 points would be really appreciated and implemented asap.

``` 
-- source configuration
create table demo_source (
    value text
    ) with (
    type = 'source',
    connector = 'nats',
    servers = 'nats-1:4222,nats-2:4222',
    'nats.stream' = 'demo-source,
    'auth.type' = 'credentials',
    'auth.username' = '{{ NATS_USER }}',
    'auth.password' = '{{ NATS_PASSWORD }}',
    format = 'json',
    'json.unstructured' = 'true'
);
```

```
-- sink configuration
create table demo_sink (
   value text
) with (
    type = 'sink',
    connector = 'nats',
    servers = 'nats-1:4222,nats-2:4222',
    'nats.subject' = 'demo.subject,
    'auth.type' = 'credentials',
    'auth.username' = '{{ NATS_USER }}',
    'auth.password' = '{{ NATS_PASSWORD }}',
    format = 'json',
    'json.unstructured' = 'true'
);
```

### DIFF
--- a/crates/arroyo-connectors/src/nats/mod.rs
+++ b/crates/arroyo-connectors/src/nats/mod.rs
@@ -46,7 +46,7 @@ impl NatsConnector {
     pub fn connection_from_options(
         options: &mut HashMap<String, String>,
     ) -> anyhow::Result<NatsConfig> {
-        let nats_servers = pull_opt("servers", options)?;
+        let nats_servers = VarStr::new(pull_opt("servers", options)?);
         let nats_auth = options.remove("auth.type");
         let nats_auth: NatsConfigAuthentication = match nats_auth.as_ref().map(|t| t.as_str()) {
             Some("none") | None => NatsConfigAuthentication::None {},
@@ -123,7 +123,12 @@ impl Connector for NatsConnector {
     }
 
     fn config_description(&self, config: Self::ProfileT) -> String {
-        (*config.servers).to_string()
+        (*config
+            .servers
+            .sub_env_vars()
+            .map_err(|e| e.context("servers"))
+            .unwrap())
+        .to_string()
     }
 
     fn table_type(&self, _: Self::ProfileT, table: Self::TableT) -> ConnectionType {
@@ -264,7 +269,11 @@ impl Connector for NatsConnector {
                     source_type: source_type
                         .clone()
                         .ok_or_else(|| anyhow!("`sourceType` is required"))?,
-                    servers: profile.servers.clone(),
+                    servers: profile
+                        .servers
+                        .sub_env_vars()
+                        .map_err(|e| e.context("servers"))?
+                        .clone(),
                     connection: profile.clone(),
                     table: table.clone(),
                     framing: config.framing,
@@ -284,7 +293,11 @@ impl Connector for NatsConnector {
                     sink_type: sink_type
                         .clone()
                         .ok_or_else(|| anyhow!("`sinkType` is required"))?,
-                    servers: profile.servers.clone(),
+                    servers: profile
+                        .servers
+                        .sub_env_vars()
+                        .map_err(|e| e.context("servers"))?
+                        .clone(),
                     connection: profile.clone(),
                     table: table.clone(),
                     publisher: None,
@@ -300,7 +313,11 @@ impl Connector for NatsConnector {
 async fn get_nats_client(connection: &NatsConfig) -> anyhow::Result<async_nats::Client> {
     let mut opts = async_nats::ConnectOptions::new();
 
-    let servers_str = connection.servers.clone();
+    let servers_str = connection
+        .servers
+        .sub_env_vars()
+        .map_err(|e| e.context("servers"))?
+        .clone();
     let servers_vec: Vec<ServerAddr> = servers_str
         .split(',')
         .map(|s| {
--- a/crates/arroyo-connectors/src/nats/mod.rs
+++ b/crates/arroyo-connectors/src/nats/mod.rs
@@ -268,7 +268,7 @@ impl Connector for NatsConnector {
                     connection: profile.clone(),
                     table: table.clone(),
                     framing: config.framing,
-                    format: config.format.unwrap(),
+                    format: config.format.expect("Format must be set for NATS source"),
                     bad_data: config.bad_data,
                     messages_per_second: NonZeroU32::new(
                         config
@@ -288,7 +288,9 @@ impl Connector for NatsConnector {
                     connection: profile.clone(),
                     table: table.clone(),
                     publisher: None,
-                    serializer: ArrowSerializer::new(config.format.unwrap()),
+                    serializer: ArrowSerializer::new(
+                        config.format.expect("Format must be set for NATS source"),
+                    ),
                 }))
             }
         })
@@ -299,7 +301,13 @@ async fn get_nats_client(connection: &NatsConfig) -> anyhow::Result<async_nats::
     let mut opts = async_nats::ConnectOptions::new();
 
     let servers_str = connection.servers.clone();
-    let servers_vec: Vec<ServerAddr> = servers_str.split(',').map(|s| s.parse().unwrap()).collect();
+    let servers_vec: Vec<ServerAddr> = servers_str
+        .split(',')
+        .map(|s| {
+            s.parse::<ServerAddr>()
+                .expect("Something went wrong while parsing servers string")
+        })
+        .collect();
 
     match &connection.authentication {
         NatsConfigAuthentication::None {} => {}
--- a/crates/arroyo-connectors/src/nats/mod.rs
+++ b/crates/arroyo-connectors/src/nats/mod.rs
@@ -17,7 +17,6 @@ use serde::{Deserialize, Serialize};
 use std::collections::HashMap;
 use std::num::NonZeroU32;
 use tokio::sync::mpsc::Sender;
-use tokio::sync::oneshot::Receiver;
 use typify::import_types;
 
 pub mod sink;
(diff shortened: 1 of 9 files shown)


crates/arroyo-connectors/src/nats/mod.rs
```diff
@@ -0,0 +1,316 @@
...
+        _: &str,
+        _: Self::ProfileT,
+        _: Self::TableT,
+        _: Option<&ConnectionSchema>,
+        _tx: Sender<TestSourceMessage>,
+    ) {
+        // TODO: Implement a full-fledge `NatsTester` struct for testing the client connection,
+        // the stream or subject existence and access permissions, the deserialization of messages
+        // for the specified format, the authentication to the schema registry (if any), etc.
+        let (tx, _rx) = tokio::sync::oneshot::channel();
```
mwylde, line 146: For testing, you should use the tx that's passed in to the test method rather than creating a new one (see the redis connector for a simple example of this). So basically in line 141 change that argument to `tx`, remove this line, and 148-151, and it shuold work.

crates/arroyo-connectors/src/nats/mod.rs
```diff
@@ -0,0 +1,316 @@
...
+
+    fn make_operator(
+        &self,
+        profile: Self::ProfileT,
+        table: Self::TableT,
+        config: OperatorConfig,
+    ) -> anyhow::Result<OperatorNode> {
+        Ok(match table.connector_type {
+            ConnectorType::Source { .. } => OperatorNode::from_source(Box::new(NatsSourceFunc {
+                stream: table.client_configs.get("nats.stream").cloned().unwrap(),
```
mwylde, line 243: This is expecting that client_configs will contain a `"nats.stream"` key, but when creating through the web ui that won't necessary be there. Instead, you can use the stream argument for the source like `ConnectorType::Source { stream }`.

We also want to avoid unwraps (and other panicking constructs) in this code, and instead return an error when the data isn't what we expect.

crates/arroyo-connectors/src/nats/mod.rs
```diff
@@ -0,0 +1,316 @@
...
+                messages_per_second: NonZeroU32::new(
+                    config
+                        .rate_limit
+                        .map(|l| l.messages_per_second)
+                        .unwrap_or(u32::MAX),
+                )
+                .unwrap(),
+            })),
+            ConnectorType::Sink { .. } => OperatorNode::from_operator(Box::new(NatsSinkFunc {
+                subject: table.client_configs.get("nats.subject").cloned().unwrap(),
```
mwylde, line 260: Similarly here, we probably want to use the subject in `ConnectorType::Sink { subject }`

crates/arroyo-connectors/src/nats/mod.rs
```diff
@@ -0,0 +1,316 @@
...
+async fn get_nats_client(
+    connection: &NatsConfig,
+    table: &NatsTable,
+) -> anyhow::Result<async_nats::Client> {
+    let client_config = get_nats_client_config(connection, table);
+    let servers_str = connection.servers.clone();
+    let servers_vec: Vec<ServerAddr> = servers_str.split(',').map(|s| s.parse().unwrap()).collect();
+    let client = async_nats::ConnectOptions::new()
+        .user_and_password(
+            client_config.get("nats.username").unwrap().to_string(),
```
mwylde, line 309: This currently panics if you don't specify a username and password; you might want to rework it so that the user_and_password method is only called if the username and password are Some values. For example

```rust
    let mut opts = async_nats::ConnectOptions::new();

    match (client_config.get("nats.username"), client_config.get("nats.password")) {
        (Some(username), Some(password)) => {
            opts = opts.user_and_password(username.clone(), password.clone());
        }
        _ => {}
    };
    
    
    let client = opts
        .connect(servers_vec)
        .await?;
```

crates/arroyo-connectors/src/nats/sink/mod.rs
```diff
@@ -0,0 +1,97 @@
...
+    async fn on_close(&mut self, _: &Option<SignalMessage>, ctx: &mut ArrowContext) {
+        if let Some(ControlMessage::Commit { epoch, commit_data }) = ctx.control_rx.recv().await {
+            self.handle_commit(epoch, &commit_data, ctx).await;
+        } else {
+            warn!("No commit message received, not committing")
+        }
+    }
+
+    async fn handle_checkpoint(&mut self, _: CheckpointBarrier, _ctx: &mut ArrowContext) {
+        // TODO: Is it necessary to insert any kind of checklpoint in the state for NATS sink?
```
mwylde, line 56: Short version: checkpointing in the sink isn't necessary for correctness; it's fine to just flush all in-progress messages. 

Longer version: there are two main strategies to achieve at-least once in a sink. The first (and usually simpler one) is what you've done here: flush the messages before finishing the checkpoint; this ensures that we won't drop any data because the downstream system will have acknowledged our writes before we mark the checkpoint as complete. If we fail during this process, we recover from the previous checkpoint and will replay the data that potentially got dropped.

The other strategy is to checkpoint our in-progress data. The downside of this is that we potentially have to write a lot more data to state. The upside is that we don't depend on the downstream system to be available in order to checkpoint. So if nats is down for 30 seconds, we're able to continue processing without blocking checkpointing.

So there's no one right answer, and often sinks will offer both options depending on which tradeoff the user prefers.

crates/arroyo-connectors/src/nats/nats.svg
mwylde, line 1: minor nit: our connector images are square and grayscale — I can put one together for you if you don't want to bother with the conversion

mwylde (commented):
This is a great start! I was able to get the source and sink working with a few tweaks.

crates/arroyo-connectors/src/nats/mod.rs
qgab-flowdesk: Thanks for pointing this out, looks like I introduced this when refactoring from `0.9` to `0.10`. This is fixed although a real tester still should be implemented

crates/arroyo-connectors/src/nats/nats.svg
qgab-flowdesk: Updated, I hope the format is correct now

crates/arroyo-connectors/src/nats/mod.rs
qgab-flowdesk: That's modified as well. Should other authentication be added to have it merged into master?

crates/arroyo-connectors/src/nats/mod.rs
qgab-flowdesk: Done 👍

crates/arroyo-connectors/src/nats/mod.rs
qgab-flowdesk: Done 👍

crates/arroyo-connectors/src/nats/table.json
```diff
@@ -0,0 +1,179 @@
...
+                                        },
+                                        "description": {
+                                            "title": "Consumer description",
+                                            "type": "string",
+                                            "description": "A description of the consumer. This can be particularly useful for ephemeral consumers to indicate their purpose since the durable name cannot be provided."
+                                        },
+                                        "filterSubjects": {
+                                            "title": "Filter subjects",
+                                            "type": "string",
+                                            "description": "A list of subjects that the consumer should filter on. If the list is empty, then no filtering is done.",
```
mwylde, line 61: I'd specify the format for this, whether it's comma-separated or what. You can also use an array which will render a nice UI.

crates/arroyo-connectors/src/nats/source/mod.rs
```diff
@@ -0,0 +1,548 @@
...
+                    AcknowledgmentPolicy::All => consumer::AckPolicy::All,
+                    AcknowledgmentPolicy::None => consumer::AckPolicy::None,
+                },
+                replay_policy: match replay_policy {
+                    ReplayPolicy::Original => consumer::ReplayPolicy::Original,
+                    ReplayPolicy::Instant => consumer::ReplayPolicy::Instant,
+                },
+                ack_wait: Duration::from_secs(ack_wait.clone() as u64),
+                description: description.clone(),
+                filter_subjects: filter_subjects
```
mwylde, line 244: It appears that this doesn't work properly if it's not set (i.e., `""`). 

I get 

> Something went wrong instantiating the NATS consumer.: Error { kind: JetStream(Error { code: 400, err_code: ErrorCode(10139), description: Some("consumer filter in FilterSubjects cannot be empty") }), source: None } panic.file="crates/arroyo-connectors/src/nats/source/mod.rs" panic.line=281 panic.column=14

crates/arroyo-connectors/src/nats/table.json
qgab-flowdesk: Now rendered as an array. The enumerated items descriptions are set to `undefined_n` though, not sure how to fix that?

crates/arroyo-connectors/src/nats/source/mod.rs
qgab-flowdesk: I think that's fixed, with single quotes at least you should be able to pass an empty string and define it as comma-separated list of strings

mwylde (approved):

PR #581, opened 2024-04-03
Implement UDAFs on datafusion

mwylde:
Adds proper support for UDAFs in datafusion by implementing an Accumulator which collects values for each UDAF argument, then on evaluate passes them into the UDAF dylib function. The state is the collectesd vector of records.

### DIFF
--- a/crates/arroyo-df/src/udafs.rs
+++ b/crates/arroyo-df/src/udafs.rs
@@ -87,7 +87,7 @@ impl Accumulator for ArroyoUdaf {
             })
             .collect();
 
-        let ColumnarValue::Scalar(scalar) = self.udf.invoke(&args?[..]).unwrap() else {
+        let ColumnarValue::Scalar(scalar) = self.udf.invoke(&args?[..])? else {
             return Err(DataFusionError::Execution(format!(
                 "UDAF {} returned an array result",
                 self.udf.name()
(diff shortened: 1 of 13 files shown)


crates/arroyo-df/src/udafs.rs
```diff
@@ -0,0 +1,228 @@
...
+                "UDAF {} returned an array result",
+                self.udf.name()
+            )));
+        };
+
+        Ok(scalar)
+    }
+
+    fn size(&self) -> usize {
+        std::mem::size_of_val(self)
```
jacksonrnewhouse, line 101: This is supposed to be the total size of the allocator, i.e. include the sizes of the arrays it owns.

jacksonrnewhouse (commented):
It looks like in parsing the UDAF code we allow a mixture of vectors and scalars, but then we throw in execution. Is that right, and are we happy with how we treat each case?

mwylde:
Addressed the comments.

With regards to the mix scalar/aggregate args, I would prefer this be a plan time failure but I'm not sure how best to accomplish that, since datafusion seems comfortable mixing argument types. Do you think it's possible for us to implement that check in a visitor?

Otherwise, we could support calling udafs with scalar args by making an array out of the scalar, but that probably isn't what the user wants and it would be better to error.

jacksonrnewhouse (approved):

PR #594, opened 2024-04-22
check that time window functions (hop, tumble, session) are moved into operators.

jacksonrnewhouse:


crates/arroyo-df/src/lib.rs
```diff
@@ -517,6 +517,8 @@ pub async fn parse_and_get_arrow_program(
         plan_rewrite.visit(&mut metadata)?;
         used_connections.extend(metadata.connection_ids.iter());
 
+        plan_rewrite.visit(&mut TimeWindowUdfChecker {})?;
```
mwylde, line 520: This should probably be in rewrite_plan, as otherwise it's only applied to inserts and not to views/memory tables.

mwylde (approved):

PR #601, opened 2024-04-24
Make the flush interval on updating aggregates configurable.

jacksonrnewhouse:


crates/arroyo-types/src/lib.rs
```diff
@@ -104,6 +104,7 @@ pub const GRPC_PORT_ENV: &str = "GRPC_PORT";
 pub const HTTP_PORT_ENV: &str = "HTTP_PORT";
 pub const COMPILER_PORT_ENV: &str = "COMPILER_PORT";
 
+pub const UPDATE_AGGREGATE_FLUSH_MICROS_ENV: &str = "UPDATE_AGGREGATE_FLUSH_MICROS";
```
mwylde, line 107: I think our other time configs are in milliseconds (like BATCH_LINGER_MS), and I think it's also a bit friendlier to someone configuring it as a reasonable number of micros is a very large number.

crates/arroyo-df/src/extension/updating_aggregate.rs
```diff
@@ -147,6 +148,11 @@ impl ArroyoExtension for UpdatingAggregateExtension {
             physical_plan_type: Some(PhysicalPlanType::Aggregate(Box::new(combine_aggregate))),
         };
 
+        let flush_interval_micros = std::env::var(UPDATE_AGGREGATE_FLUSH_MICROS_ENV)
+            .ok()
+            .map(|s| Duration::from_micros(s.parse().expect("invalid flush interval")))
```
mwylde, line 153: Can this return an error instead of panicking?

crates/arroyo-types/src/lib.rs
jacksonrnewhouse: Done

crates/arroyo-df/src/extension/updating_aggregate.rs
```diff
@@ -147,6 +148,17 @@ impl ArroyoExtension for UpdatingAggregateExtension {
             physical_plan_type: Some(PhysicalPlanType::Aggregate(Box::new(combine_aggregate))),
         };
 
+        let flush_interval = std::env::var(UPDATE_AGGREGATE_FLUSH_MILLIS_ENV)
+            .ok()
+            .map(|s| {
+                let Ok(millis): Result<u64, _> = s.parse() else {
+                    return plan_err!("Failed to parse {} to a number", s);
```
mwylde, line 155: This should still probably reference what the number is (i.e., that it's the UPDATE_AGGREGATE_FLUSH_MILLIS_ENV value)  otherwise it lacks context for the user.

crates/arroyo-types/src/lib.rs
```diff
@@ -104,6 +104,7 @@ pub const GRPC_PORT_ENV: &str = "GRPC_PORT";
 pub const HTTP_PORT_ENV: &str = "HTTP_PORT";
 pub const COMPILER_PORT_ENV: &str = "COMPILER_PORT";
 
+pub const UPDATE_AGGREGATE_FLUSH_MILLIS_ENV: &str = "UPDATE_AGGREGATE_FLUSH_MILLIS";
```
mwylde, line 107: This should be MS instead of MILLIS to match BATCH_LINGER_MS

mwylde (approved):

PR #628, opened 2024-05-20
ignore tests with deps

mxfactorial:
#### current
https://doc.arroyo.dev/developing/dev-setup#testing tasks the `cargo test` and `cargo nextest run` commands after a `master` clone but several tests with dependencies fail

#### expected
1. post `master` clone cargo test commands pass
2. `cargo test -- --ignored` to run tests with dependencies

welcome to change expected


mwylde:
Hey @mxfactorial — thanks for your contribution! I feel for the problem you're trying to solve here—it's annoying for tests to fail because of services that aren't running. However, I don't think using `ignore` is the best approach, since these are valuable tests that in general we do want to run. Instead, it might be possible to use cargo features to disable tests, since in CI we already run with --all-features.

mxfactorial:
can add `kafka = []` to Cargo.toml features, then `#[cfg_attr(not(feature = "kafka"), ignore)]` etc

but more errors appear since opening the pr

for example, `cargo test` fails when [crates/arroyo-controller/build.rs](https://github.com/ArroyoSystems/arroyo/blob/3d4bd7feab313b86e08074bb97dd0ac66ef567b0/crates/arroyo-controller/build.rs) doesnt connect to postgres

the devx intended by this branch is a pass after git clone && cargo test

which devs rely on for an initial sanity test when onboarding

but this project doesnt currently restrict `cargo test` to unit tests

add requirements at your leisure

can help test on a fresh machine if lowering the onboarding cost is a priority for you

other options:
1. increasing content in https://doc.arroyo.dev/developing/dev-setup#testing
2. add a cloud dev env (gitpod/codespace)

PR #633, opened 2024-05-23
Suppress NATS heartbeat error

benjamin-awd:
The `MissingHeartbeat` often indicates a temporary hiccup in networking/processing, and should not be treated as a breaking error by the worker/controller.

image

### DIFF
--- a/crates/arroyo-connectors/src/nats/source/mod.rs
+++ b/crates/arroyo-connectors/src/nats/source/mod.rs
@@ -422,7 +422,15 @@ impl NatsSourceFunc {
                                     }
                                 },
                                 Some(Err(msg)) => {
-                                    return Err(UserError::new("NATS message error", msg.to_string()));
+                                    match msg.kind() {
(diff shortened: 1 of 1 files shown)


crates/arroyo-connectors/src/nats/source/mod.rs
```diff
@@ -422,7 +422,15 @@ impl NatsSourceFunc {
                                     }
                                 },
                                 Some(Err(msg)) => {
-                                    return Err(UserError::new("NATS message error", msg.to_string()));
+                                    match msg.kind() {
+                                        consumer::pull::MessagesErrorKind::MissingHeartbeat => {
```
mwylde, line 426: Is it worth failing after some number of missed heartbeats?

benjamin-awd: That's a good idea, will have a think about how to best implement a check for that - probably something involving the NatsState cache

Issue #621, opened 2024-05-07
API throttle when using GCS as checkpoint backend


gbto:
We have observed on our Kubernetes deployments some responsiveness issue in the UI. After investigation, we've discovered the API calls made to fetch pipelines and jobs were pending and eventually timed out. Digging deeper, this was due to the controller calling the GCP metadata server too often:

`object_store::gcp::credential::fetching token from metadata server`

Which results in:

`object_store::client::retry::Encountered transport error (error sending request for url (http://metadata/computeMetadata/v1/instance/service-accounts/default/token?audience=https%3A%2F%2Fwww.googleapis.com%2Foauth2%2Fv4%2Ftoken): error trying to connect: operation timed out) backing off for 0.1 seconds, retry 1 of 10`

So after an attempt to reduce the number of `object_store::gcp::credential` call to the metadata server, I've switched from [Kubernetes SA authentication](https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity?cloudshell=false#kubernetes-sa-to-iam) to passing the serialised JSON of the SA credentials as `GOOGLE_SERVICE_ACCOUNT_KEY` environment variable, which did reduce the number of calls but did not resolve the unresponsive Arroyo API issue. I also tried hacking the object_store by forcing it to use the S3 client for authenticating to GCS with `AWS_DEFAULT_REGION`, `AWS_ENDPOINT` but had no luck for the authentication...

Here are the screenshots of what that looks like in practice:

<img width="1799" alt="Screenshot_2024-05-06_at_18 33 00" src="https://github.com/ArroyoSystems/arroyo/assets/29350919/153385f5-2c4b-4d1b-a249-36127f6f14bc">
<img width="1547" alt="Screenshot 2024-05-06 at 18 34 45" src="https://github.com/ArroyoSystems/arroyo/assets/29350919/f32d1176-6c8f-4ee1-af0b-9ec9bd094066">

Happy to help if there's anything I can do !

mwylde:
We believe we have found the root cause of this issue within the object_store crate, and have opened https://github.com/apache/arrow-rs/issues/6625 there.

mwylde:
Fixed in #770

PR #634, opened 2024-05-23
Add object store cache for GCS

benjamin-awd:
Resolves https://github.com/ArroyoSystems/arroyo/issues/621

This PR:
1. Changes the `construct_gcs` function to async, which should hopefully improve performance
2. Adds a cache so to prevent redundant DNS lookups
3. Adds an alternative way to create the GCS builder via a GOOGLE_SERVICE_ACCOUNT_KEY

Currently, every checkpoint instantiates a ObjectStore leading to a large number of calls to the metadata server. This can lead to a high number of concurrent DNS lookups, which may cause network latency and other undesirable effects. While the GCE metadata service endpoint has no [official rate limit](https://www.googlecloudcommunity.com/gc/Infrastructure-Compute-Storage/Rate-limits-on-metadata-server-access-token-requests/m-p/440525#M1001), we should avoid making unnecessary calls to it.

Took reference from Polars on this: https://github.com/pola-rs/polars/issues/14384#issuecomment-1948991697, https://github.com/pola-rs/polars/blob/main/crates/polars-io/src/cloud/object_store_setup.rs#L4

Note: the AWS metadata endpoint has a rate limit of 1024 packets per second, so it might be worth implementing this for the `construct_s3` function as well at some point.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -976,6 +976,7 @@ dependencies = [
  "bytes",
  "futures",
  "object_store",
+ "once_cell",
  "regex",
  "rusoto_core",
  "thiserror",
--- a/crates/arroyo-storage/Cargo.toml
+++ b/crates/arroyo-storage/Cargo.toml
@@ -23,3 +23,4 @@ tokio-util = {version = "0.7.9", features = ["io"]}
 async-trait = "0.1.73"
 futures = "0.3.28"
 webpki = ">=0.22.2"
+once_cell = "1.19.0"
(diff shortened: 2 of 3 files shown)


mwylde (approved):
This is great! Thanks for the contribution. Can you confirm that this addresses #621?

benjamin-awd:
> This is great! Thanks for the contribution. Can you confirm that this addresses https://github.com/ArroyoSystems/arroyo/issues/621?

It's looking a lot better so far on our end, but will keep an eye on it -- I think our remaining issues are due to how the NATS connector handles checkpointing

PR #652, opened 2024-06-10
setting imagePullSecrets for workers

chokosabe:
Setting the imagepullsecrets. Not sure this is the best location to do this in.

### DIFF
--- a/crates/arroyo-rpc/default.toml
+++ b/crates/arroyo-rpc/default.toml
@@ -64,7 +64,6 @@ resource-mode = "per-slot"
 name-prefix = "arroyo"
 image = "ghcr.io/arroyosystems/arroyo:latest"
 image-pull-policy = "IfNotPresent"
-image-pull-secrets = "IfNotPresent"
 service-account-name = "default"
 resources = { requests = { cpu = "900m",  memory = "500Mi" } }
 task-slots = 16
--- a/crates/arroyo-rpc/default.toml
+++ b/crates/arroyo-rpc/default.toml
@@ -64,6 +64,7 @@ resource-mode = "per-slot"
 name-prefix = "arroyo"
 image = "ghcr.io/arroyosystems/arroyo:latest"
 image-pull-policy = "IfNotPresent"
+image-pull-secrets = "IfNotPresent"
 service-account-name = "default"
 resources = { requests = { cpu = "900m",  memory = "500Mi" } }
 task-slots = 16
--- a/k8s/arroyo/templates/configmap.yaml
+++ b/k8s/arroyo/templates/configmap.yaml
@@ -59,6 +59,10 @@ data:
           {{- end }}    
         image: "{{ .Values.worker.image.repository }}:{{ .Values.worker.image.tag }}"
         image-pull-policy: "{{ .Values.worker.image.pullPolicy }}"
+        imagePullSecrets:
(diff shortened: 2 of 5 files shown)


mwylde (changes requested):
This is a great start but need a little more work

chokosabe:
Added the changes requested. There does seem to be an intermittent build error with mqtt.

chokosabe:
Updates pushed up. Resolved all the changes.

k8s/arroyo/templates/configmap.yaml
```diff
@@ -83,4 +87,4 @@ data:
           - name: arroyo-user-config
             mountPath: /config
           {{- end }}
-        command: "/app/arroyo-bin --config-dir /config worker"
\ No newline at end of file
+        command: ["/app/arroyo-bin --config-dir /config worker"]
```
mwylde, line 90: This still isn't quite right. With this, trying to run this in helm gives the error 

```
Configuration is invalid!
  • invalid type: found sequence, expected a string for key "default.kubernetes-scheduler.worker.command" in /root/.config/arroyo/config.yaml YAML file
```
It should be restored to what it currently is in master.

k8s/arroyo/templates/configmap.yaml
```diff
@@ -59,6 +59,10 @@ data:
           {{- end }}    
         image: "{{ .Values.worker.image.repository }}:{{ .Values.worker.image.tag }}"
         image-pull-policy: "{{ .Values.worker.image.pullPolicy }}"
+        image-pull-secrets:
+        {{- with .Values.imagePullSecrets }}
```
mwylde, line 63: `image_pull_secrets` is defined as a string, but here it's passed in an object which fails:
```
Configuration is invalid!
  • invalid type: found unit, expected a string for key "default.kubernetes-scheduler.worker.image-pull-secrets" in /root/.config/arroyo/config.yaml YAML file
```

crates/arroyo-rpc/default.toml
```diff
@@ -64,6 +64,7 @@ resource-mode = "per-slot"
 name-prefix = "arroyo"
 image = "ghcr.io/arroyosystems/arroyo:latest"
 image-pull-policy = "IfNotPresent"
+image-pull-secrets = "IfNotPresent"
```
mwylde, line 67: Is this a sensible default for this?

mwylde (changes requested):
This still isn't quite working. It might be useful to try testing it locally with docker desktop. With a local k8s environment, you can test it by creating a values.yaml file like this:

```
artifactUrl: "/tmp/arroyo-test"
checkpointUrl: "/tmp/arroyo-test"

volumes:
  - name: checkpoints
    hostPath:
      path: /tmp/arroyo-test
      type: DirectoryOrCreate

volumeMounts:
  - name: checkpoints
    mountPath: /tmp/arroyo-test
```

Then building an image from your branch and deploying it locally:

```
$ docker build . -f docker/Dockerfile -t ghcr.io/arroyosystems/arroyo:tip --target arroyo
$ helm install arroyo  k8s/arroyo -f values.yaml 
```

mwylde (approved):
Thanks for the contribution! I've gone ahead and rebased on master and fixed a couple of issues so that we can get this into the next release.

PR #677, opened 2024-07-04
allow cluster-domain for services to be configured in values.yaml

jr200:
the cluster domain is often overridden when multiple clusters are being managed. small patch to allow configuration of this field in the generated manifests.


mwylde (approved):
Thanks for the contribution!

PR #701, opened 2024-07-31
feat: add display_graphivz for LogicalGraph

zhuliquan:
It's helpful debug LogicalGraph like func `display_graphviz` in `LogicalPlan`


zhuliquan:
Sorry, I found similary func `Program.dot`

Issue #703, opened 2024-08-03
need set --experimental_allow_proto3_optional for protocbuf compiler


zhuliquan:
I found below error when cargo build,
```shell
error: failed to run custom build command for `arroyo-rpc v0.12.0-dev (/mnt/d/workspace/rust/github.com/zhuliquan/arroyo/crates/arroyo-rpc)`

Caused by:
  process didn't exit successfully: `/mnt/d/workspace/rust/github.com/zhuliquan/arroyo/target/debug/build/arroyo-rpc-6d47546660ec2985/build-script-build` (exit status: 1)
  --- stdout
  cargo:rerun-if-changed=proto/rpc.proto
  cargo:rerun-if-changed=proto

  --- stderr
  Error: Custom { kind: Other, error: "protoc failed: rpc.proto: This file contains proto3 optional fields, but --experimental_allow_proto3_optional was not set.\n" }
warning: build failed, waiting for other jobs to finish...
```

PR #704, opened 2024-08-03
fix: add experimental_allow_proto3_optional compile option

zhuliquan:
add experimental_allow_proto3_optional compile option
fixing issue #703


mwylde (approved):
Thanks for the contribution! While we have been directing developers to install a newer version of protoc (i.e., in https://doc.arroyo.dev/developing/dev-setup#dependencies) it's great that's not required.

PR #709, opened 2024-08-07
feat: add line number and file info in log message

zhuliquan:
Adding line number and file info in log message, it's helpful for debugging


mwylde (changes requested):
Thanks for the contribution! I don't think we want this enabled by default, as it's not relevant to most users. I'd be happy to accept it if it were enabled with a config setting.

zhuliquan:
> Thanks for the contribution! I don't think we want this enabled by default, as it's not relevant to most users. I'd be happy to accept it if it were enabled with a config setting.

Hi, I add two option for logging file line number and file name

https://github.com/zhuliquan/arroyo/blob/18cbc4d6ef0c2bad6585af32bc5d7a81339e54af/crates/arroyo-rpc/src/config.rs#L634-L641

crates/arroyo-rpc/default.toml
```diff
@@ -82,3 +82,7 @@ user = "arroyo"
 password = "arroyo"
 
 [logging]
+format = "plaintext" # option: plaintext / json / logfmt 
+nonblocking = true
```
mwylde, line 86: The current default is actually "nonblocking = false" since nonblocking mode causes a large increase in memory usage

zhuliquan: "nonblocking = false" will make logging as sync process. if it cause large memory usage. I suggest to decrease `buffered_lines_limit` of `tracing-appender` like below code (default is `pub const DEFAULT_BUFFERED_LINES_LIMIT: usize = 128_000`)
https://github.com/tokio-rs/tracing/blob/527b4f66a604e7a6baa6aa7536428e3a303ba3c8/tracing-appender/src/non_blocking.rs#L395-L398

mwylde (approved):
Thanks for the contribution!

PR #720, opened 2024-08-18
GitHub actions binaries

MarcoLugo:
This PR stems from a discussion about a missing macos intel binary in the releases page. It adds an automated to create that missing binary but also others which are already available but not automated at the moment.

It ran successfully (see [here](https://github.com/MarcoLugo/arroyo/actions/runs/10440758864)).

Please note that the linux arm runner still needs to be added or enabled (see the Linux matrix).


mwylde (approved):
Thanks, this is awesome!

PR #730, opened 2024-09-02
minor: change ouput_schema

zhuliquan:
It's may be not right for `KeyCalculationExtension`.output_schema. I notice that output_schema is depend on `self.input.schema`
https://github.com/ArroyoSystems/arroyo/blob/68fb412bfd15e706e59cbaac8295942d30badfee/crates/arroyo-planner/src/extension/key_calculation.rs#L112-L115
But according to func `new_named_and_trimmed`, the schema of `KeyCaclculationExtension` may not equal to `self.input.schema`
https://github.com/ArroyoSystems/arroyo/blob/68fb412bfd15e706e59cbaac8295942d30badfee/crates/arroyo-planner/src/extension/key_calculation.rs#L39-L57


zhuliquan:
Sorry, wrong understand

PR #745, opened 2024-09-23
feat: using NodePort export svc in k8s

zhuliquan:
original arroyo-controller, I can't access service outsize of k8s cluster. I think can using NodePort to export arroyo service. I notice that print NOTE using `kubectl port-forward service/arroyo 5115:80` It's hard to use, because `kubectl port-forward` won't return and not daemon cmd.


zhuliquan:
Sorry, It's not safe

PR #759, opened 2024-10-18
urlencode subject in confluent schema registry URLs

emef:
thank you for the recent protobuf support! while testing this, I received some 404 errors while creating a kafka source that are caused by subject names that include slashes. this happens when a protobuf schema definition includes imports to other proto files in different directories (the subject will be the file path of the import)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -877,6 +877,7 @@ dependencies = [
  "k8s-openapi",
  "log",
  "nanoid",
+ "percent-encoding",
  "prost 0.12.4",
  "prost-reflect 0.13.1",
  "rand 0.8.5",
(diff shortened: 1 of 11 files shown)


mwylde:
Hey @emef — appreciate your efforts debugging the CI issues. I don't think increasing the number of retries will help, as there seems to be some fatal error with the pipeline process; possibly it's being OOM killed, but unfortunately the CI system doesn't give us good logs on that.

It does seem to be running successfully now, so it may have been some temporary instability on the side of our action runner provider (ubicloud).

If you back out the retry changes we should be able to get this merged.

mwylde (approved):
Thanks for the fix, and your patience working through the CI challenges!

PR #763, opened 2024-10-22
skip the whole confluent schema header in proto deser

emef:
the header format for protobuf messages that use confluent schema registry expects there to be a message index list following the schema ID. this will always contain at least one varint which contains the length of the index list.


mwylde (approved):
Thanks!

PR #765, opened 2024-10-24
Non-data fields - Metadata fields to Kafka Connector

vaibhawvipul:
enabling users to query using the following syntax - 

```sql
create table users (
    id TEXT,
    name TEXT,
    offset BIGINT GENERATED ALWAYS AS (metadata('offset_id')) STORED
) with (
    connector = 'kafka',
    ...
);
```

Demo - 
Screenshot 2024-10-24 at 11 21 10 AM


crates/arroyo-planner/src/tables.rs
```diff
@@ -511,16 +534,55 @@ impl Table {
...
+                                                    ),
+                                                ),
+                                            )) => {
+                                                connector_metadata_columns.insert(
+                                                    column.name.value.to_string(),
+                                                    value.to_string(),
+                                                );
+                                            }
+                                            _ => {
+                                                "Unsupported argument format.".to_string();
```
mwylde, line 570: I'm a bit confusing by the intent here—this string is created by never used. Is this meant to return an error?

crates/arroyo-planner/src/tables.rs
```diff
@@ -511,16 +534,55 @@ impl Table {
...
-                    if let ColumnOption::Generated {
-                        generation_expr, ..
-                    } = &option.option
-                    {
-                        generation_expr.clone()
-                    } else {
-                        None
-                    }
-                });
+                let generating_expression =
```
mwylde, line 537: This function has very deep nesting and the logic is quite hard to follow. Is there a way to simplify it?

crates/arroyo-connectors/src/kafka/source/mod.rs
```diff
@@ -178,7 +180,15 @@ impl KafkaSourceFunc {
                                     .ok_or_else(|| UserError::new("Failed to read timestamp from Kafka record",
                                         "The message read from Kafka did not contain a message timestamp"))?;
 
-                                ctx.deserialize_slice(v, from_millis(timestamp as u64)).await?;
+                                let connector_metadata = ConnectorMetadata {
+                                    enable_metadata: self.enable_metadata.unwrap_or(false),
+                                    message_offset: msg.offset(),
+                                    message_partition: msg.partition(),
+                                    message_topic: self.topic.clone(),
+                                    metadata_fields: self.metadata_fields.clone(),
```
mwylde, line 188: Can we avoid these clones?

crates/arroyo-formats/src/de.rs
```diff
@@ -82,6 +83,9 @@ pub struct ArrowDeserializer {
     schema_registry: Arc<Mutex<HashMap<u32, apache_avro::schema::Schema>>>,
     proto_pool: DescriptorPool,
     schema_resolver: Arc<dyn SchemaResolver + Sync>,
+    enable_metadata: Option<bool>,
+    kafka_metadata_builder: Option<(Int64Builder, Int32Builder, StringBuilder)>,
```
mwylde, line 87: This approach is very specific to Kafka, but the goal for this project is to have a generic metadata system that any connector can opt into with its specific metadata fields. How can we make the deserialization support work for whatever metadata fields a connector has?

crates/arroyo-connectors/src/kafka/mod.rs
```diff
@@ -323,7 +326,19 @@ impl Connector for KafkaConnector {
 
         let table = Self::table_from_options(options)?;
 
-        Self::from_config(self, None, name, connection, table, schema)
+        let allowed_metadata_udf_args = ["offset_id", "partition", "topic"];
+        if let Some(fields) = &metadata_fields {
+            for key in fields.keys() {
+                if !allowed_metadata_udf_args.contains(&key.as_str()) {
```
mwylde, line 332: This should be checking the value rather than the key (since the key is the field name, not the metadata name)

crates/arroyo-connectors/src/kafka/source/mod.rs
```diff
@@ -178,7 +180,29 @@ impl KafkaSourceFunc {
                                     .ok_or_else(|| UserError::new("Failed to read timestamp from Kafka record",
                                         "The message read from Kafka did not contain a message timestamp"))?;
 
-                                ctx.deserialize_slice(v, from_millis(timestamp as u64)).await?;
+                                let connector_metadata = if self.metadata_fields.is_some() {
```
mwylde, line 183: This can be made a bit cleaner with an if let statement

crates/arroyo-formats/src/de.rs
```diff
@@ -195,6 +207,17 @@ impl ArrowDeserializer {
                     .map(|batch| {
                         let mut columns = batch.columns().to_vec();
                         columns.insert(self.schema.timestamp_index, Arc::new(timestamp.finish()));
+
+                        if self.additional_fields_builder.is_some() {
```
mwylde, line 211: Same here, this can be an if let

crates/arroyo-formats/src/de.rs
```diff
@@ -233,10 +257,20 @@ impl ArrowDeserializer {
             }) => {
                 self.deserialize_raw_string(buffer, msg);
                 add_timestamp(buffer, self.schema.timestamp_index, timestamp);
+                if additional_fields.is_some() {
```
mwylde, line 260: This can also be an if let

crates/arroyo-formats/src/de.rs
```diff
@@ -249,10 +283,63 @@ impl ArrowDeserializer {
...
+                                .collect(),
+                        );
+                    }
+                }
+
                 decoder
                     .decode(msg)
                     .map_err(|e| SourceError::bad_data(format!("invalid JSON: {:?}", e)))?;
                 timestamp_builder.append_value(to_nanos(timestamp) as i64);
+                if let Some(additional_fields) = additional_fields {
```
mwylde, line 310: This code is duplicated a couple of times, is it possible to abstract it out?

crates/arroyo-formats/src/de.rs
```diff
@@ -195,6 +207,17 @@ impl ArrowDeserializer {
...
                         let mut columns = batch.columns().to_vec();
                         columns.insert(self.schema.timestamp_index, Arc::new(timestamp.finish()));
+
+                        if self.additional_fields_builder.is_some() {
+                            for (field_name, mut builder) in
+                                self.additional_fields_builder.take().unwrap()
+                            {
+                                let (idx, _) =
+                                    self.schema.schema.column_with_name(&field_name).unwrap();
+                                columns.remove(idx);
```
mwylde, line 217: I find this logic a bit hard to follow—why are we removing columns?

mwylde (changes requested):
Great progress on the PR!

A few remaining issues.

We're now validating metadata names (although the key/value are switched), but we're missing two types of validations for the user:

* Data types, which if the user sets the wrong type they get a runtime error, for example if they declare

```
    offset_id INT GENERATED ALWAYS AS (metadata('offset_id')) STORED
```

and we're not validating that the metadata columns have only a metadata function in them, so this plans:

```
    offset_id BIGINT GENERATED ALWAYS AS (concat(metadata('offset_id'), '1')) STORED
```

Beyond that, I'd like to see some tests, both for the deserialization logic and for planning. It might also be worth modifying the kafka tests to use metadata fields.

vaibhawvipul:
@mwylde can you please trigger the CI again? I am asking because my last commit ran successfully, the failed CI is on master merge.

mwylde (approved):
Thanks, this is a great contribution!

PR #766, opened 2024-10-24
feat: make heap profiling as default

zhuliquan:
Sometimes, we need to analyze the memory usage of workers or controllers, and `pprof` is a very useful tool for this. I believe it should be included as a default feature rather than an optional compile-time setting. Additionally, it doesn't make much sense to expose an HTTP route if it can't be utilized effectively.

### DIFF
--- a/crates/arroyo/src/main.rs
+++ b/crates/arroyo/src/main.rs
@@ -33,7 +33,6 @@ use uuid::Uuid;
 static ALLOC: tikv_jemallocator::Jemalloc = tikv_jemallocator::Jemalloc;
 
 #[cfg(all(
-    feature = "profiling",
     not(target_env = "msvc"),
     not(target_os = "macos"),
     any(target_arch = "x86_64", target_arch = "aarch64")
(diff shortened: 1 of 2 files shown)


crates/arroyo/src/main.rs
```diff
@@ -25,6 +25,7 @@ use tracing::{debug, error, info};
 use uuid::Uuid;
 
 #[cfg(all(
+    feature = "profiling",
```
mwylde, line 28: This enables jemalloc only when profiling is enabled, which doesn't make sense to me. Profiling only works on certain environments, which is much more restrictive than jemalloc in general.

mwylde:
I think it would be good to disable the HTTP route if profiling isn't enabled, or return a more helpful error there. Also adding it as an option to the Dockerfile makes sense, although we probably wouldn't want it enabled by default.

crates/arroyo/src/main.rs
zhuliquan: Sorry, I will recover later

PR #774, opened 2024-10-30
adding metadata to mqtt connector

vaibhawvipul:
- adding metadata to mqtt connectors
- topic as a non-data field


mwylde (approved):
Thanks!

PR #777, opened 2024-11-01
Feature support set level of logger

zhuliquan:
Hello, Every time I want to debug problem and view the logs, I always see a large number of INFO logs, sometimes I only need to look at the warn and error level logs, meanwhile printing log in terminal frequently will affect the performance. If we can control the log level through the config file will be better, so that we can debug problem in the development stage, and ensure the performance in the production stage.


crates/arroyo-server-common/src/lib.rs
```diff
@@ -54,11 +54,29 @@ pub fn init_logging(name: &str) -> Option<WorkerGuard> {
...
         name,
         EnvFilter::builder()
-            .with_default_directive(LevelFilter::INFO.into())
+            .with_default_directive(parse_level_filter().into())
             .from_env_lossy(),
     )
 }
 
+fn parse_level_filter() -> LevelFilter {
+    let level_filter = config()
```
mwylde, line 63: I think we want the RUST_LOG env var to have the highest precedence, so that they can override what's the in the config file as needed.

With this change, it's not possible to use RUST_LOG at all because min_level is set in defaults.toml, and always will override the env var.

Additionally, the env var supports much more than just static levels (like "info" or "warn")—it's able to completely configure the logger on a per-crate basis. For example:


```
RUST_LOG=info,arroyo_worker=debug,arroyo_controller=debug,arroyo_state=debug
```

I think an approach would be to:

* First check if the env var is set, and if so use that with `from_env_lossy`; if not we can use the min_level config.

mwylde (changes requested):

crates/arroyo-server-common/src/lib.rs
zhuliquan: thank for @mwylde review, Sorry for later reply. I agree with Env var high priority, but I don't know how to implement min log level for each sub-crate.

zhuliquan: > I think we want the RUST_LOG env var to have the highest precedence, so that they can override what's the in the config file as needed.
> 
> With this change, it's not possible to use RUST_LOG at all because min_level is set in defaults.toml, and always will override the env var.
> 
> Additionally, the env var supports much more than just static levels (like "info" or "warn")—it's able to completely configure the logger on a per-crate basis. For example:
> 
> ```
> RUST_LOG=info,arroyo_worker=debug,arroyo_controller=debug,arroyo_state=debug
> ```
> 
> I think an approach would be to:
> 
> * First check if the env var is set, and if so use that with `from_env_lossy`; if not we can use the min_level config.

parse level like this? first fetch ENV var then fetch min_level of logging config
```rust
let level_filter = std::env::var(EnvFilter::DEFAULT_ENV).unwrap_or(
        config()
            .logging
            .min_level
            .clone()
            .unwrap_or("info".to_string()),
    );
```

PR #789, opened 2024-11-21
support AWS_MSK_IAM authentication

emef:
adds an AWS_MSK_IAM authentication type to the kafka config profile. submitting PR as a draft for early feedback while I do some testing on our deployment

proof that connection tester works (removed bootstrap servers from screenshot):
image

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -566,6 +566,8 @@ dependencies = [
  "async-nats",
  "async-trait",
  "aws-config",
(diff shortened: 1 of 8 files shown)


mwylde (changes requested):
Thanks for the contribution, looks great! Just needs a clean `cargo clippy` run to merge. It looks like a test needs to be updated with the API changes.

mwylde (approved):
Thanks!

PR #791, opened 2024-11-22
add kafka connection properties to profile

emef:
this is to support adding additional rdkafka config params when creating a kafka connection profile

### DIFF
--- a/crates/arroyo-connectors/src/kafka/mod.rs
+++ b/crates/arroyo-connectors/src/kafka/mod.rs
@@ -105,7 +105,7 @@ impl KafkaConnector {
             authentication: auth,
             bootstrap_servers: BootstrapServers(pull_opt("bootstrap_servers", options)?),
             schema_registry_enum: schema_registry,
-            connection_properties: vec![],
+            connection_properties: HashMap::new(),
         })
     }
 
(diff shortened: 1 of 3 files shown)


crates/arroyo-connectors/src/kafka/profile.json
```diff
@@ -106,6 +106,16 @@
                     "sensitive": ["apiSecret"]
                 }
             ]
+        },
+        "connectionProperties": {
```
mwylde, line 110: This would be a bit better as  as map instead of a vec so that users don't need to know the k/v syntax. This is how we do property overrides on the table:

https://github.com/ArroyoSystems/arroyo/blob/a9ac8f70e6468d9dadf4d7dfd0d7fab01c14f6f6/crates/arroyo-connectors/src/kafka/table.json#L83-L90

crates/arroyo-connectors/src/kafka/mod.rs
```diff
@@ -926,6 +927,14 @@ pub fn client_configs(
...
     };
 
+    for prop in connection.connection_properties.iter() {
+        if let Some((k, v)) = prop.split_once('=') {
+            client_configs.insert(k.to_string(), v.to_string());
+        } else {
+            bail!("invalid connection property: {}", prop);
+        }
+    }
+
```
mwylde, line 937: With this change we'll be able to set client configs from both the connection and the table. I think it would be good to warn when iterating through the table configs (on 939) if both define the same property.

crates/arroyo-connectors/src/kafka/profile.json
mattforbes-gr: 👍 agreed map is better

mwylde (approved):
Looks great!

Issue #700, opened 2024-07-31
May be can merge same sink node in a logical graph?


zhuliquan:
This is a awesome project, I notice that arroyo merge source node (two query sql with same source) in logical graph. Why not merge the sink nodes (if the sink is the same) as well, which will save the number of file handles or Kafka connections.

zhuliquan:
for example:
```sql
CREATE TABLE process_details (
  timestamp TIMESTAMP,
  uuid BIGINT,
  asset_oid BIGINT,
  process_id1 INT,
  process_id2 INT,
  process_id3 INT
) WITH (
  connector = 'single_file',
  path = '$input_dir/test.json',
  format = 'json',
  type = 'source'
);

CREATE TABLE alerts (
  timestamp TIMESTAMP,
  uuid BIGINT,
  asset_oid BIGINT,
  process_id1 INT,
  process_id2 INT,
  process_id3 INT,
  rule_id TEXT,
) WITH (
  connector = 'single_file',
  path = '$output_path',
  format = 'json',
  type = 'sink'
);
INSERT INTO alerts SELECT *, 'v1' as rule_id FROM process_details 
WHERE partition_record((asset_oid = 1)) AND ((process_id1 + process_id2 = 2) OR (process_id3 = 4));
INSERT INTO alerts SELECT *, 'v2' as rule_id FROM process_details 
WHERE partition_record((asset_oid = 1)) AND ((process_id1 + process_id2 = 3) OR (process_id3 = 3));
```
we can get two logical plan
```shell
# Logical Plan 1
Projection: process_details.timestamp AS timestamp, process_details.uuid AS uuid, process_details.asset_oid AS asset_oid, process_details.process_id1 AS process_id1, process_details.process_id2 AS process_id2, process_details.process_id3 AS process_id3, rule_id AS rule_id, process_details._timestamp
  Projection: process_details.timestamp, process_details.uuid, process_details.asset_oid, process_details.process_id1, process_details.process_id2, process_details.process_id3, Utf8("v1") AS rule_id, process_details._timestamp
    Filter: partition_record(process_details.asset_oid = Int64(1)) AND (process_details.process_id1 + process_details.process_id2 = Int32(2) OR process_details.process_id3 = Int32(4))
      WatermarkNode(process_details): process_details.timestamp, process_details.uuid, process_details.asset_oid, process_details.process_id1, process_details.process_id2, process_details.process_id3, process_details._timestamp
[...]

PR #794, opened 2024-11-26
Feature merge sink

zhuliquan:
fix issue #700 merge sinks with same table, it's helpful to save file handles or kafka connectors. Besides, it may fix bug two pipeline write data to same file.
For example
```sql
CREATE TABLE cars (
        timestamp TIMESTAMP,
        driver_id BIGINT,
        event_type TEXT,
        location TEXT
) WITH (
        connector = 'single_file',
        path = 'cars.json',
        format = 'json',
        type = 'source'
);

CREATE TABLE cars_output (
        timestamp TIMESTAMP,
        driver_id BIGINT,
        event_type TEXT,
        location TEXT
) WITH (
        connector = 'single_file',
        path = 'cars_output.json',
        format = 'json',
        type = 'sink'
);
INSERT INTO cars_output SELECT * FROM cars WHERE driver_id = 100 AND event_type = 'pickup';
INSERT INTO cars_output SELECT * FROM cars WHERE driver_id = 101 AND event_type = 'dropoff';
```
if we don't merge sink node of two query, results of query1 will be overwrited by results of query2. Because there two sink node infer two file handle for same file (`cars_output.json`).

### DIFF
--- a/crates/arroyo-planner/src/builder.rs
+++ b/crates/arroyo-planner/src/builder.rs
@@ -213,6 +213,7 @@ pub(crate) enum NamedNode {
     Source(TableReference),
     Watermark(TableReference),
     RemoteTable(TableReference),
+    Sink(TableReference),
 }
 
 struct ArroyoExtensionPlanner {}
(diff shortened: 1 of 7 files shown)


mwylde (approved):
Apologies for the long review time. Thanks for the change!

PR #798, opened 2024-12-01
Add RabbitMQ Stream source connector

ecarrara:
This PR adds RabbitMQ Stream source connector.


crates/arroyo-connectors/src/rabbitmq_stream/mod.rs
```diff
@@ -0,0 +1,276 @@
...
+
+import_types!(schema = "src/rabbitmq_stream/profile.json");
+import_types!(schema = "src/rabbitmq_stream/table.json");
+
+impl Connector for RabbitmqStreamConnector {
+    type ProfileT = RabbitmqStreamConfig;
+    type TableT = RabbitmqStreamTable;
+
+    fn name(&self) -> &'static str {
+        "rabbitmq_stream"
```
mwylde, line 29: I think just `rabbitmq` would be a better name, as it aligns with our other connectors that identify the system rather than the particular way they interact. So for example, we have a single `redis` connector that can read from various kinds of redis datastructures.

crates/arroyo-connectors/src/rabbitmq_stream/mod.rs
```diff
@@ -0,0 +1,276 @@
...
+                        trust_certificates: options
+                            .remove("tls_config.trust_certificates")
+                            .map(|trust_certificates| trust_certificates == "true"),
+                        root_certificates_path: options.remove("tls_config.root_certificates_path"),
+                        client_certificates_path: options
+                            .remove("tls_config.client_certificates_path"),
+                        client_keys_path: options.remove("tls_config.client_keys_path"),
+                    });
+
+                let load_balancer_mode = options.remove("json.unstructured").map(|t| t == "true");
```
mwylde, line 111: I'm not quite sure what's going on here—`json.unstructured` controls our deserialization logic. Was this meant to be another config?

crates/arroyo-connectors/src/rabbitmq_stream/source.rs
```diff
@@ -0,0 +1,178 @@
...
+                control_message = ctx.control_rx.recv() => {
+                    info!("control_message {:?}", control_message);
+                    match control_message {
+                        Some(ControlMessage::Checkpoint(c)) => {
+                            debug!("starting checkpointing {}", ctx.task_info.task_index);
+
+                            let s = ctx.table_manager.get_global_keyed_state("s")
+                               .await
+                               .expect("should be able to get rabbitmq stream state");
+                            s.insert(self.stream.clone(), offset).await;
```
mwylde, line 148: On line 77 we're trying to read the state as a RabbitmqStreamState, but here we're inserting a u64. That breaks when we try to checkpoint the state with 

```
2024-12-05T00:05:42.869074Z ERROR arroyo_server_common: panicked at crates/arroyo-connectors/src/rabbitmq_stream/source.rs:147:33:
should be able to get rabbitmq stream state: Failed to downcast table s to key type alloc::string::String and value type u64 panic.file="crates/arroyo-connectors/src/rabbitmq_stream/source.rs" panic.line=147 panic.column=33
```

mwylde (changes requested):
A few comments but overall this looks great! Thanks for the contribution.

crates/arroyo-connectors/src/rabbitmq_stream/mod.rs
ecarrara: I will also rename the `RabbitmqStreamConnector` struct to `RabbitmqConnector`. Ok?

crates/arroyo-connectors/src/rabbitmq_stream/mod.rs
ecarrara: Yes, it should be `load_balancer_mode`! Thank you for identifying this during the review process. This option forces the client to connect to the load balancer instead of connecting directly to the node.

https://www.rabbitmq.com/blog/2021/07/23/connecting-to-streams#with-a-load-balancer

mwylde (approved):
Looks great!

Issue #844, opened 2025-03-10
Kubernetes Tolerations and NodeSelector for Helm install


Morphexe:
Can we add tolerations and nodeSelector to a Arroyo Worker via config when installing in Kubernetes via helm?

mwylde:
This would be a great contribution for anyone familiar with helm!

mwylde:
This was addressed by #850, thanks @RatulDawar!

PR #850, opened 2025-03-22
Add nodeSelector and tolerations support for worker pods in configmap

RatulDawar:
Added nodeSelector and tolerations for arroyo workers. 
https://github.com/ArroyoSystems/arroyo/issues/844

### DIFF
--- a/k8s/arroyo/values.yaml
+++ b/k8s/arroyo/values.yaml
@@ -30,6 +30,21 @@ worker:
       memory: 490Mi
       cpu: 900m
   slots: 16
+  # ref: https://kubernetes.io/docs/user-guide/node-selection/
+  nodeSelector: {}
+    # kubernetes.io/role: worker
+    # node-type: arroyo-worker
+
+  # Tolerations for worker pods assignment
+  # ref: https://kubernetes.io/docs/concepts/configuration/taint-and-toleration/
+  #
+  tolerations: {}
+    # - key: "dedicated"
+    #   operator: "Equal"
+    #   value: "arroyo"
+    #   effect: "NoSchedule"
+    # - key: "CriticalAddonsOnly"
+    #   operator: "Exists"
 
 postgresql:
   # set to true to deploy a postgres instance in-cluster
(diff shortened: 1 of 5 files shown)


k8s/arroyo/values.yaml
```diff
@@ -30,6 +30,21 @@ worker:
       memory: 490Mi
       cpu: 900m
   slots: 16
+  # ref: https://kubernetes.io/docs/user-guide/node-selection/
+  nodeSelector: {}
+    # kubernetes.io/role: worker
```
mwylde, line 35: Should these commented lines be left in?

mwylde (changes requested):
Thanks for the contribution!

This is a great start. There are a few more changes needed to make this work for the workers.

The actual worker pods are created by the Arroyo kubernetes scheduler (https://github.com/ArroyoSystems/arroyo/blob/master/crates/arroyo-controller/src/schedulers/kubernetes/mod.rs). 

This will need to be configured with the node selectors and tolerations (https://github.com/ArroyoSystems/arroyo/blob/1a6c5cc95af500dbd5e7e7e5a5cc1d5ed3354a2f/crates/arroyo-controller/src/schedulers/kubernetes/mod.rs#L134).

This in turn should be threaded through via the configuration system. You can see an example of this for annotations here: https://github.com/ArroyoSystems/arroyo/blob/1a6c5cc95af500dbd5e7e7e5a5cc1d5ed3354a2f/crates/arroyo-rpc/src/config.rs#L540.

k8s/arroyo/values.yaml
RatulDawar: Have taken reference from all the standard helm charts, like starrocks etc. But yes not necessary to leave it here.

RatulDawar: Missed the scheduler thing, fixing that right now

RatulDawar: Made the requried changes and tested it locally on minikube configs getting applied as expected

mwylde (approved):
Thanks, this is great!

PR #851, opened 2025-03-23
Support NATS JWT auth

nlap:
Adds JWT/NKey auth for NATS. This auth type is required to use Arroyo with the Synadia NATS cloud service.

Thanks for this project - looking forward to your feedback!


mwylde (approved):
Thanks for the contribution!

PR #852, opened 2025-03-23
Fix NATS consumer conflict

nlap:
When multiple pipelines connect to the same NATS stream, they create consumers with identical names, resulting in "consumer deleted" error:

`ERROR arroyo_server_common: panicked at crates/arroyo-connectors/src/nats/source/mod.rs:69:17:
NATS message error: consumer deleted panic.file="crates/arroyo-connectors/src/nats/source/mod.rs" panic.line=69 panic.column=17
INFO arroyo_connectors::nats::source: >> Existing consumer deleted. Recreating consumer with new `start_sequence`.`

Adding the job ID to the consumer name prevents this conflict

### DIFF
--- a/crates/arroyo-connectors/src/nats/source/mod.rs
+++ b/crates/arroyo-connectors/src/nats/source/mod.rs
@@ -192,7 +192,7 @@ impl NatsSourceFunc {
         };
 
         let consumer_name = format!(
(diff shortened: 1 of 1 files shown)


mwylde (approved):
Thanks!

PR #885, opened 2025-05-09
Implement Display for Sensitve type

pancernik:
This adds masking to the Postgres password printed in logs in [main.rs](https://github.com/ArroyoSystems/arroyo/blob/master/crates/arroyo/src/main.rs#L370).


mwylde (approved):
Nice!

PR #886, opened 2025-05-13
Add native support for r2

mwylde:
This PR adds native support for [Cloudflare R2](https://www.cloudflare.com/developer-platform/products/r2/) object storage for checkpoints and the filesystem connector.

Previously, this was supported by using an S3 custom endpoint (as R2 is S3-compatible), but that would cause some unwanted AWS-specific configuration logic to be run. Now R2 can be used directly, in one of four forms:

* `r2://bucket/path/to/files` (this requires the account id to be specified as an option or environment variable `CLOUDFLARE_ACCOUNT_I`
* `r2://{account_id}@bucket/path/to/files`
* `https://{account_id}.r2.cloudflarestorage.com/mybucket/my/key/path`
* `https://{account_id}.eu.r2.cloudflarestorage.com/mybucket/my/key/path` (specifying a jurisdiction)


cmackenzie1 (approved):
LGTM!

Not sure it matters for this to support the virtual bucket host names `https://BUCKET.ACCOUNT_ID.r2.cloudflarestorage.com/path/to/object`?

mwylde:
Thanks for pointing that out — I didn't realize R2 supported virtual bucket host names. We do support that for s3 so probably makes sense to do it for r2 as well

Issue #818, opened 2025-01-02
Build a Connector for Iggy


tbar4:
Iggy is the persistent message streaming platform written in Rust, supporting QUIC, TCP and HTTP transport protocols, capable of processing millions of messages per second. Considering Iggy is written in Rust, it would be interesting if there was a connector that was designed to take iggy messages and process them in arroyo. 

You could stream iggy directly into DataFusion for processing (similar to Arroyo's architecture) but it would be nice to just build your producer in iggy and have Arroyo be the consumer for stream processing. 

https://iggy.rs

PR #897, opened 2025-06-07
Connector/apache iggy

RatulDawar:
Adding connector for [Apache Iggy](https://github.com/apache/iggy). 
We tried to keep the connector similar to apache kafka. 
Currently supporting http transfer protocol and unstructed json data type. 
Thanks to @Vinamra7  for helping contribute the source connector.
This refers the issue https://github.com/ArroyoSystems/arroyo/issues/818

### DIFF
--- a/crates/arroyo-connectors/src/iggy/sink/mod.rs
+++ b/crates/arroyo-connectors/src/iggy/sink/mod.rs
@@ -122,14 +122,13 @@ impl ArrowOperator for IggySinkFunc {
 
         let mut messages = Vec::new();
         for (i, v) in values.into_iter().enumerate() {
-            let _timestamp = timestamps
-                .and_then(|ts| {
-                    if ts.is_null(i) {
-                        None
-                    } else {
-                        Some(ts.value(i) / 1_000_000)
-                    }
-                });
+            let _timestamp = timestamps.and_then(|ts| {
+                if ts.is_null(i) {
+                    None
+                } else {
+                    Some(ts.value(i) / 1_000_000)
+                }
+            });
 
             let message = Message {
                 id: 0,


webui/src/routes/connections/DefineSchema.tsx
```diff
@@ -38,14 +45,26 @@ const SchemaFormatEditor = ({
...
+
+  const baseSchemaTypeOptions: SchemaTypeOption[] = [
     { name: 'Infer schema', value: 'inferred' },
     { name: capitalize(format) + ' schema', value: 'schema' },
   ];
 
-  if (format == 'json') {
-    schemaTypeOptions.push({ name: 'Unstructured JSON', value: 'unstructured' });
-  }
+  const schemaTypeOptions = useMemo(() => {
```
mwylde, line 54: Can you explain the reason for this change? In general, all formats should be supported for any connector, as the format system is responsible for deserializing the bytes after they've been read from the source. We also want to avoid introducing special cases for particular connectors in the UI.

crates/arroyo-connectors/src/iggy/iggy.svg
```diff
@@ -0,0 +1,56 @@
+<?xml version="1.0" encoding="utf-8"?>
```
mwylde, line 1: If possible, this should be converted to black and white to match with the other connector images. (If that's not too much trouble, I can just up a monochrome version).

crates/arroyo-connectors/src/iggy/profile.json
```diff
@@ -0,0 +1,63 @@
...
+                    "additionalProperties": false
+                },
+                {
+                    "type": "object",
+                    "title": "Username/Password",
+                    "properties": {
+                        "username": {
+                            "title": "Username",
+                            "type": "string",
+                            "description": "Username for authentication"
```
mwylde, line 38: It's useful for username to also be a var-str, even though it's not generally sensitive, as that allows users to manage usernames and passwords together.

crates/arroyo-connectors/src/iggy/table.json
```diff
@@ -0,0 +1,86 @@
...
+                        "offset": {
+                            "type": "string",
+                            "description": "The offset to start reading from",
+                            "enum": [
+                                "latest",
+                                "earliest"
+                            ]
+                        },
+                        "consumer_id": {
+                            "type": "integer",
```
mwylde, line 32: I think in general users shouldn't need to set a consumer name, since in arroyo we manage the horizontal scaling using low-level apis rather than high-level consumer group APIs.

It also looks like Iggy supports both numeric and textual consumers; I think many users would prefer to use names so perhaps it makes sense to support both types here by taking a string?

crates/arroyo-connectors/src/iggy/table.json
```diff
@@ -0,0 +1,86 @@
...
+                            "title": "Consumer ID",
+                            "description": "The ID of the consumer to use for polling messages"
+                        },
+                        "partition_id": {
```
mwylde, line 36: I'm not sure we want to have the user choose a particular partition to read from; generally you want to read from all partitions. At the very least I think we need an option to read from all partitions.

crates/arroyo-connectors/src/iggy/mod.rs
```diff
@@ -0,0 +1,624 @@
...
+    type TableT = IggyTable;
+
+    fn name(&self) -> &'static str {
+        "iggy"
+    }
+
+    fn metadata(&self) -> arroyo_rpc::api_types::connections::Connector {
+        arroyo_rpc::api_types::connections::Connector {
+            id: "iggy".to_string(),
+            name: "Apache Iggy".to_string(),
```
mwylde, line 67: I'd prefer just "Iggy" to "Apache Iggy" as otherwise all of the apache projects would clump at the beginning of the list and be harder to find.

crates/arroyo-connectors/src/iggy/mod.rs
```diff
@@ -0,0 +1,624 @@
...
+            transport: transport_protocol,
+            authentication: authentication.clone(),
+        };
+
+        let stream = options.pull_str("stream")?;
+        let topic = options.pull_str("topic")?;
+
+        let table_type = if options.pull_bool("is_source")? {
+            let offset = match options.pull_str("offset")?.as_str() {
+                "earliest" => "earliest",
```
mwylde, line 162: Perhaps the conversion to SourceOffset should happen here?

crates/arroyo-connectors/src/iggy/mod.rs
```diff
@@ -0,0 +1,624 @@
...
+                    "latest" => SourceOffset::Latest,
+                    _ => bail!("Invalid offset value"),
+                },
+                consumer_id: consumer_id as i64,
+                partition_id: partition_id as i64,
+                auto_commit,
+            }
+        } else {
+            let partitioning = match options.pull_str("partitioning")?.as_str() {
+                "partition_id" => "partition_id",
```
mwylde, line 183: Similarly here, we're first validating the strings and then on line 190 converting to enums, we can just do that once

crates/arroyo-connectors/src/iggy/mod.rs
```diff
@@ -0,0 +1,624 @@
...
+
+    fn from_config(
+        &self,
+        id: Option<i64>,
+        name: &str,
+        config: Self::ProfileT,
+        table: Self::TableT,
+        schema: Option<&ConnectionSchema>,
+    ) -> Result<ArroyoConnection> {
+        let schema = schema.cloned().unwrap_or_else(|| ConnectionSchema {
```
mwylde, line 217: This needs to use the actual passed in schema. See examples from the kafka connector.

crates/arroyo-connectors/src/iggy/mod.rs
```diff
@@ -0,0 +1,624 @@
...
+
+pub struct IggyTester {
+    pub connection: IggyConfig,
+}
+
+impl IggyTester {
+    async fn connect(&self) -> Result<IggyClient, String> {
+        info!("Testing connection to Iggy at {}", self.connection.endpoint);
+
+        let client = IggyClient::default();
```
mwylde, line 358: It looks like we're not actually using the endpoint, so it will only work with the default

crates/arroyo-connectors/src/iggy/mod.rs
```diff
@@ -0,0 +1,624 @@
...
+            .connect()
+            .await
+            .map_err(|e| format!("Failed to connect to Iggy server: {:?}", e))?;
+
+        match &self.connection.authentication {
+            IggyConfigAuthentication::None {} => {
+                info!("No authentication required");
+            }
+            IggyConfigAuthentication::UsernamePassword { username, password } => {
+                let password_str = password.sub_env_vars().map_err(|e| {
```
mwylde, line 370: sub_env_vars already produces a useful error message, that it can just be propagated directly with ?

crates/arroyo-connectors/src/iggy/source/mod.rs
```diff
@@ -0,0 +1,314 @@
...
+    pub client: Option<IggyClient>,
+    pub messages_per_second: NonZeroU32,
+    pub metadata_fields: Vec<MetadataField>,
+}
+
+impl IggySourceFunc {
+    async fn init_client(&mut self) -> Result<()> {
+        info!("Creating Iggy client for {}", self.endpoint);
+
+        let client = IggyClient::default();
```
mwylde, line 56: We're not passing in the configured endpoint here

crates/arroyo-connectors/src/iggy/source/mod.rs
```diff
@@ -0,0 +1,314 @@
...
+                        let table = ctx
+                            .table_manager
+                            .get_global_keyed_state("iggy")
+                            .await
+                            .map_err(|err| {
+                                UserError::new("Failed to get global key value", err.to_string())
+                            })?;
+                        table.insert(self.partition_id, state).await;
+
+                        let client_temp = self.client.take();
```
mwylde, line 144: why is this here?

crates/arroyo-connectors/src/iggy/source/mod.rs
```diff
@@ -0,0 +1,314 @@
...
+        message: &PolledMessage,
+        collector: &mut SourceCollector,
+    ) -> Result<(), UserError> {
+        let timestamp =
+            { SystemTime::UNIX_EPOCH + Duration::from_millis(message.timestamp / 1000) };
+
+        let mut additional_fields = HashMap::new();
+        for field in &self.metadata_fields {
+            match field.key.as_str() {
+                "offset" => {
```
mwylde, line 241: This can be made a bit more succinct by pulling the insert call out of each match statement (since it's always the same)

crates/arroyo-connectors/src/iggy/sink/mod.rs
```diff
@@ -0,0 +1,160 @@
...
+            PartitioningStrategy::Balanced => Partitioning::balanced(),
+        };
+
+        let mut messages_to_send = messages;
+        match client
+            .send_messages(&stream_id, &topic_id, &partitioning, &mut messages_to_send)
+            .await
+        {
+            Ok(_) => {
+                info!("Sent {} messages to Iggy", messages_to_send.len());
```
mwylde, line 77: This should be downgraded to a debug or trace, otherwise will be very noisy

crates/arroyo-connectors/src/iggy/sink/mod.rs
```diff
@@ -0,0 +1,160 @@
...
+                length: v.len() as u32,
+                payload: Bytes::from(v),
+                headers: None,
+            };
+
+            messages.push(message);
+        }
+
+        if let Err(e) = self.send_messages(messages, ctx).await {
+            error!("Failed to send messages to Iggy: {:?}", e);
```
mwylde, line 144: This is currently dropping messages on failure, which is generally the wrong default for a stream processing system like arroyo. Instead, we should retry some number of times and then fail the pipeline by panicking.

mwylde (changes requested):
Apologies for the long delay on the review, and thanks for the contribution! This is a good start that needs a little more work before we'll be able to merge it.

In addition to the changes requested in the PR, it will also need state handling in the source. In order to provide at least-once semantics, arroyo needs to manage it own offset tracking as part of its state. Otherwise, if we're relying on auto-comitting, we might drop messages on restart. Look at how this is implemented in the Kafka connector.

PR #898, opened 2025-06-11
Better understand checkpoint timing

hhoughgg:
Screenshot 2025-06-11 at 12 59 57 PM

Aiming to better understand checkpoint timing by displaying the data as part of the entire checkpoint timeline. Looks like some of the timing information is missing or invalid and hope to address later.

Also made a change to request checkpoints from workers in parallel from the controller rather than waiting for a response from each worker before asking the next.


webui/src/components/CheckpointDetails.tsx
```diff
@@ -1,21 +1,48 @@
...
+  checkpoint: Checkpoint;
 }
 
-function formatDuration(micros: number): string {
+export function formatDuration(micros: number): string {
   let millis = micros / 1000;
   let secs = Math.floor(millis / 1000);
   if (millis < 1000) {
     return `${millis} ms`;
+  } else if (millis < 10000) {
```
cmackenzie1, line 34: If I am reading this right, <10s is formatted as milliseconds?

webui/src/components/CheckpointDetails.tsx
```diff
@@ -30,68 +57,228 @@ function formatDuration(micros: number): string {
...
-const spans = (subtasks: SubtaskCheckpointGroup[], spanType: CheckpointSpanType) => {
-  return (
-    <Stack>
-      {subtasks.map(s => (
-        <Text>{spanDuration(s, spanType)}</Text>
-      ))}
-    </Stack>
-  );
-};
+  useEffect(() => {
```
cmackenzie1, line 54: This useEffect seems heavy. Can the graph structure and the data be separated into two things?

mwylde (approved):

PR #899, opened 2025-06-12
Add TLS support for all services

mwylde:
This PR adds support for running all Arroyo services over TLS. This includes the HTTP APIs (REST API and admin endpoints), gRPC (controller, worker, compiler, and node), and the customer worker TCP shuffle networking.

TLS can be enabled globally or per-service, via the `tls` top-level config, or the various `{service}.tls` configs, which will override the top-level one for that service

```toml
[tls]
enabled = true
cert-file = '/Users/mwylde/tls/server.crt'
key-file = '/Users/mwylde/tls/server.key'

[worker.tls]
enabled = false
```

To run locally with TLS (using a private CA), you can follow these steps:

```
# 1. Generate CA private key and self-signed cert
openssl genrsa -out ca.key 4096               # CA private key
openssl req -x509 -new -key ca.key -sha256 -days 3650 -out ca.crt -subj "/CN=Test Root CA"  # CA cert

# 2. Generate localhost key and CSR
openssl genrsa -out server.key 4096           # server private key
openssl req -new -key server.key -out server.csr -subj "/CN=localhost"  # CSR for localhost

# 3. Sign server cert with CA
openssl x509 -req -in server.csr -CA ca.crt -CAkey ca.key -CAcreateserial -out server.crt -days 365 -sha256
```

Then configure arroyo with cert-file and key-file pointing to the server.crt and server.key files generated.

You may also need to point rustls to your CA (this seems to be necessary at least on macos), which you can do with `SSL_CERT_FILE=~/tls/ca.crt`

In a clustered setting, this will currently only really work with servers that have hostnames, as it is operationally challenging to generate certs for ip addresses. We will need to figure out a path forward for that. There are is no support for authentication yet; that will be handled in a follow-on PR.

### DIFF
--- a/crates/arroyo-api/src/lib.rs
+++ b/crates/arroyo-api/src/lib.rs
@@ -119,21 +119,24 @@ pub(crate) fn to_micros(dt: OffsetDateTime) -> u64 {
 
 pub async fn compiler_service() -> Result<CompilerGrpcClient<Channel>, ErrorResp> {
     // TODO: cache this
-    CompilerGrpcClient::connect(config().compiler_endpoint().to_string())
+    let config = config();
+    let endpoint = config.compiler_endpoint();
+
+    let channel = arroyo_rpc::connect_grpc(endpoint, &config.compiler.tls)
         .await
         .map_err(|e| {
             error!("Failed to connect to compiler service: {}", e);
             service_unavailable("compiler-service")
-        })
+        })?;
+
+    Ok(CompilerGrpcClient::new(channel))
 }
 
 pub async fn start_server(database: DatabaseSource, guard: ShutdownGuard) -> anyhow::Result<u16> {
     let config = config();
     let addr = SocketAddr::new(config.api.bind_address, config.api.http_port);
-    let listener = TcpListener::bind(addr).await?;
-    let local_addr = listener.local_addr()?;
 
-    let app = rest::create_rest_app(database, &config.controller_endpoint()).layer(
+    let app = rest::create_rest_app(database).layer(
         CompressionLayer::new().zstd(true).compress_when(
             DefaultPredicate::new()
                 // compression doesn't work for server-sent events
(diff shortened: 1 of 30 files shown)


crates/arroyo-controller/src/schedulers/mod.rs
```diff
@@ -329,6 +345,11 @@ impl NodeScheduler {
         }
     }
 
+    async fn client(node: &NodeStatus) -> anyhow::Result<NodeGrpcClient<Channel>> {
+        let channel = connect_grpc(format!("http://{}", node.addr), &config().node.tls).await?;
```
cmackenzie1, line 349: Should this be `https://` for TLS enabled services?

mwylde: We do the substitution in connect_grpc so that all of the callers don't have to include the "is this tls" logic

crates/arroyo-api/src/lib.rs
```diff
@@ -142,12 +145,28 @@ pub async fn start_server(database: DatabaseSource, guard: ShutdownGuard) -> any
         ),
     );
 
-    info!("Starting API server on {:?}", local_addr);
-    guard.into_spawn_task(wrap_start("api", local_addr, async {
-        axum::serve(listener, app.into_make_service()).await
-    }));
+    if let Some(tls_config) = config.get_tls_config(&config.api.tls) {
```
oliy, line 148: This pattern is repeated in all the services that startup TLS.  Maybe it should be part of `wrap_start`, with optional TLS configuration passed in?

crates/arroyo-controller/src/schedulers/mod.rs
oliy: Maybe the `http://` part should just be part of the `connect_grpc` responsibility then?  (i.e. instead of the URL, you pass the address, path (defaulted to ""), and tls config?).

crates/arroyo-node/src/lib.rs
```diff
@@ -250,26 +265,36 @@ pub async fn start_server(guard: ShutdownGuard) -> NodeId {
...
+            .max_frame_size(Some((1 << 24) - 1)) // 16MB
+            .add_service(NodeGrpcServer::new(server))
+            .serve(bind_addr)
+    } else {
+        info!(
+            "Starting node server on {} with {} slots",
+            bind_addr, config.node.task_slots
+        );
+
+        arroyo_server_common::grpc_server()
```
oliy, line 287: same theme as other comments.  maybe could make `arroyo_server_comment::grpc_server` accept an optional TLS config as a builder option before `serve` is called?

crates/arroyo-controller/src/schedulers/mod.rs
mwylde: That could be a good option, but would be a breaking configuration change (or some version of this code at each callsite) to transition to. The other callers of this function are passing in user-configured endpoints, which all include the scheme.

crates/arroyo-api/src/lib.rs
mwylde: That would definitely be nice, but isn't possible to make generic across multiple services due to the type-centric design of tower (at least, not possible without using a macro).

crates/arroyo-node/src/lib.rs
mwylde: Yeah, for the same reason — Rust lacks higher-level generics which makes it impossible to abstract some of this stuff (without macros).

PR #901, opened 2025-06-17
Add support for configuring MQTTs max_packet_size

janekhaertter:
When using the MQTT sink for records containing a lot of data, Arroyo errors with
```
ERROR arroyo_connectors::mqtt::sink: Failed to poll mqtt eventloop: MqttState(OutgoingPacketTooLarge { pkt_size: 25546, max: 10240 })
```

Currently, it is impossible to increase the max value. This PR adds the `maxPacketSize` configuration option to the MQTT sink that allows the user to set the `max_packet_size` on the underlying rumqttc client, resolving the issue.

### DIFF
--- a/crates/arroyo-connectors/src/mqtt/mod.rs
+++ b/crates/arroyo-connectors/src/mqtt/mod.rs
@@ -74,12 +74,18 @@ impl MqttConnector {
             None
         };
 
+        let max_packet_size = options
+            .pull_opt_str("maxPacketSize")?
+            .map(|s| s.parse::<u32>())
+            .transpose()?;
+
         Ok(MqttConfig {
             url,
             username,
             password,
             tls,
             client_prefix: options.pull_opt_str("client_prefix")?,
+            max_packet_size,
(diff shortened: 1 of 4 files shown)


crates/arroyo-connectors/src/mqtt/mod.rs
```diff
@@ -74,12 +74,18 @@ impl MqttConnector {
             None
         };
 
+        let max_packet_size = options
+            .pull_opt_str("maxPacketSize")?
```
mwylde: nit: for SQL options, we use underscores (so `max_packet_size`). Also, since this is a number you can pull it directly as one with `pull_opt_u64` which will allow users to not need to quote the value, and will handle errors nicely

mwylde (changes requested):
One small comment but overall looks great!

janekhaertter:
Thanks for the quick response!

mwylde (approved):
This is great! Thanks!

PR #908, opened 2025-06-27
mTLS and key-based authentication

mwylde:
This PR adds mTLS authentication for all services, and the ability to set a static Bearer token to product HTTP APIs (the API service and admin service).

mTLS is enabled by setting the global or service-specific `tls.mtls_ca_file` configuration to point to a CA certificate file.

### DIFF
--- a/crates/arroyo-server-common/src/tls.rs
+++ b/crates/arroyo-server-common/src/tls.rs
@@ -1,25 +1,82 @@
-use anyhow::{anyhow, Result};
-use arroyo_rpc::config::TlsConfig;
+use anyhow::{anyhow, bail, Result};
+use arroyo_rpc::config::{config, ApiAuthMode, TlsConfig};
 use arroyo_rpc::native_cert_store;
 use axum_server::tls_rustls::RustlsConfig;
+use rustls::pki_types::pem::PemObject;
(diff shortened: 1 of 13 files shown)


crates/arroyo-server-common/src/tls.rs
```diff
@@ -1,25 +1,82 @@
...
+use rustls::pki_types::{CertificateDer, PrivateKeyDer};
+use rustls::server::WebPkiClientVerifier;
 use rustls::{ClientConfig, RootCertStore, ServerConfig};
 use std::sync::Arc;
-use tonic::transport::{Identity, ServerTlsConfig};
+use tonic::transport::{Certificate, Identity, ServerTlsConfig};
 
 /// Create TLS configuration for Axum HTTP/HTTPS servers
-pub async fn create_http_tls_config(tls_config: &TlsConfig) -> Result<RustlsConfig> {
+pub async fn create_http_tls_config(
```
oliy, line 13: Maybe this should be renamed?  Not really just the tls config anymore.  More of auth config and configuring TLS is one of the auth strategies.

mwylde: This method is still just handling TLS config (possibly mTLS). HTTP auth is handled elsewhere, when creating the server.

PR #912, opened 2025-07-08
Fix infinite retries in MQTT sink

janekhaertter:
Currently, there is no limit on the connection attempts of an MQTT sink due to a sign error. This PR fixes the issue.


mwylde:
Thanks for the fix!

mwylde (approved):

PR #914, opened 2025-07-11
Don't schedule on workers with old run ids

mwylde:
This PR adds run_id (an incrementing integer which identifies a particular run of a job) to the worker -> controller register RPC, and updates the scheduling state to ignore workers with the wrong run id.

This fixes a race condition case in scheduling in the case that a worker container crashes and is restarted. In certain distributed schedulers, this can lead to a sequence:

1. The controller detects the failure of the worker and moves to recovering
2. In Recovering, the controller tears down the existing cluster
3. It moves to Scheduling, and creates a new set of workers
4. The restarted worker comes up, and registers itself
5. The controller tries to schedule tasks on the worker
6. The cluster scheduler finishes tearing down the worker node
7. Scheduling times out

By conditioning scheduling on the worker having the expected run_id, we avoid the mistaken schedule in step 5.

This PR also changes the name of "node_id" (which identifies which server the worker is running on) to "machine_id" to not clash with our _other_ meaning of node_id, which is the dataflow graph vertex that a task is associated with, and changes the type of a string to more usefully identify servers across various platforms (like physical server names or kubernetes pods).

### DIFF
--- a/crates/arroyo-rpc/proto/rpc.proto
+++ b/crates/arroyo-rpc/proto/rpc.proto
@@ -6,14 +6,20 @@ import "api.proto";
 
 // Controller
 
+message WorkerInfo {
+  uint64 worker_id = 1;
+  string job_id = 2;
+  uint64 run_id = 3;
+  string machine_id = 4;
+}
+
+
 message WorkerResources {
   uint64 slots = 1;
 }
 
 message RegisterWorkerReq {
-  uint64 worker_id = 1;
-  uint64 node_id = 2;
-  string job_id = 3;
+  WorkerInfo worker_info = 1;
   string rpc_address = 4;
   string data_address = 5;
   WorkerResources resources = 6;
@@ -109,7 +115,7 @@ message TaskStartedResp {
 }
 
 message RegisterNodeReq {
-  uint64 node_id = 1;
+  string machine_id = 1;
   uint64 task_slots = 2;
   string addr = 3;
 }
@@ -118,7 +124,7 @@ message RegisterNodeResp {
 }
 
 message HeartbeatNodeReq {
-  uint64 node_id = 1;
+  string machine_id = 1;
   uint64 time = 2;
 }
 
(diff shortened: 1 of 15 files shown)


crates/arroyo-rpc/proto/rpc.proto
```diff
@@ -139,10 +145,8 @@ message SinkDataResp {
 }
 
 message WorkerFinishedReq {
-  uint64 node_id = 1;
-  uint64 worker_id = 2;
+  WorkerInfo worker_info = 1;
   uint64 slots = 3;
```
cmackenzie1, line 149: Does the gap for `2` matter at all here? Does it need to be `reserved` or slots renumbered?

crates/arroyo-types/src/lib.rs
```diff
@@ -27,8 +28,14 @@ pub const HASH_SEEDS: [u64; 4] = [
 #[derive(Debug, Hash, Eq, PartialEq, Copy, Clone)]
 pub struct WorkerId(pub u64);
 
-#[derive(Debug, Hash, Eq, PartialEq, Copy, Clone)]
-pub struct NodeId(pub u64);
+#[derive(Debug, Hash, Eq, PartialEq, Clone)]
+pub struct MachineId(pub Arc<String>);
```
cmackenzie1, line 32: `Arc<String>` feels funny here

crates/arroyo-rpc/proto/rpc.proto
mwylde: It doesn't particularly matter, although it could also be renumbered. We don't maintain rpc-compatibility for minor version updates (so workers for 0.15 won't be compatible with a controller for 0.14, and users need to upgrade the controller first).

crates/arroyo-types/src/lib.rs
mwylde: Makes it cheaper to pass these around without paying to copy the string, although we end up copying a bunch anyways to put this in protobufs.

PR #922, opened 2025-07-30
Add support for time-based generated filenames

cmackenzie1:
This commit adds two new time-based filename generation strategies: UUIDv7 and ULID.

These can be enabled by choosing them when defining the connection using `'filename.strategy' = 'uuidv7'` or `'filename.strategy' = 'ulid'`

It should be noted that these times represent ONLY the time the filename was generated and SHOULD NOT be used to infer the contents of each file.


mwylde (approved):
Looks great!

PR #923, opened 2025-08-01
add azure storage support

askvinni:
## Summary

Keeping in mind that I'm not anywhere near proficient in rust, I'm starting to look into arroyo as a possible streaming engine, but noticed it lacks azure support. This PR attempts to fix that.

Should be fairly self explanatory as the `object_store` crate includes an azure builder. Since Azure has a ton of different authentication methods, I'm opting to only offer it through env vars parsed automatically by the `object_store` builder (see below)

```
* AZURE_STORAGE_ACCOUNT_NAME: storage account name
* AZURE_STORAGE_ACCOUNT_KEY: storage account master key
* AZURE_STORAGE_ACCESS_KEY: alias for AZURE_STORAGE_ACCOUNT_KEY
* AZURE_STORAGE_CLIENT_ID -> client id for service principal authorization
* AZURE_STORAGE_CLIENT_SECRET -> client secret for service principal authorization
* AZURE_STORAGE_TENANT_ID -> tenant id used in oauth flows
```

## How I tested these changes

Unit tests, compiled image and ran locally against an azure account, tested both writing to azure with the filesystem sink, as well as the backend checkpointing.
```
docker run -p 5115:5115 \
  -e ARROYO__CHECKPOINT_URL="https://my-cool-storageaccount.blob.core.windows.net/my-cool-bucket/arroyo-test" \
  -e AZURE_STORAGE_ACCOUNT_NAME= \
  -e AZURE_STORAGE_CLIENT_ID= \
  -e AZURE_STORAGE_CLIENT_SECRET= \
  -e AZURE_STORAGE_TENANT_ID= \
  arroyo-full
```

### To note

The delta sink is untested; I expect it would work based on the underlying filesystem implementation, but can't be sure as I don't have a delta setup. The docs on `storage_provider.get_backing_store()` confused me a little, as typically azure would contain both the container and the storage account, but the `object_store` crate also has a few examples that include a URI following the format I passed.

### DIFF
--- a/crates/arroyo-connectors/src/filesystem/sink/delta.rs
+++ b/crates/arroyo-connectors/src/filesystem/sink/delta.rs
@@ -76,6 +76,14 @@ pub(crate) async fn load_or_create_table(
                 storage_provider.qualify_path(empty_path)
             ),
         ),
+        BackendConfig::Azure(azure) => (
+            storage_provider.get_backing_store(),
+            format!(
+                "abfs://{}/{}",
+                azure.container,
+                storage_provider.qualify_path(empty_path)
+            ),
+        ),
         BackendConfig::Local(_) => (storage_provider.get_backing_store(), "/".to_string()),
     };
 


mwylde (approved):
Sorry for the slow review, this looks good!

PR #929, opened 2025-08-20
Iceberg sink

mwylde:
This PR adds an Iceberg sink, which is capable of writing data into Iceberg V2 tables. It looks like this:

```sql
create table nexmark with (
    connector = 'nexmark',
    event_rate = 100
);

create table sink with (
    connector = 'iceberg',
    'catalog.type' = 'rest',
    'catalog.rest.url' = 'https://catalog.cloudflarestorage.com/599a34f21899e3773eb72ed48b6c41b1/iceberg-testing',
    'catalog.rest.token' = '{{CC_TOKEN}}',
    'catalog.warehouse' = '599a34f21899e3773eb72ed48b6c41b1_iceberg-testing',
    type = 'sink',
    format = 'parquet',
    table_name = 'arroyo_test',
    'rolling_policy.interval' = '30s'
);

insert into sink
select bid.auction, bid.bidder, bid.price, bid.channel, bid.url,  bid.extra from nexmark
where bid is not null;
```

An Iceberg sink is a special kind of FilesystemSink, which in addition to writing parquet files efficiently to object storage will also compute and write Iceberg metadata on commit. This is performed as part of Arroyo's consistent checkpointing system and 2pc protocol to ensure exactly-once semantics, when used with exactly-once sources (like kafka).

Currently there are some limitations:
* Parquet-only
* Only Rest catalogs are supported
* No partitioning support
* We don't yet compact metadata
* Creating Iceberg tables is not yet supported in the web UI

With this, we have also standardized some of the options for the filesystem and delta sinks, which is a breaking change for users of those connectors.

### DIFF
--- a/crates/arroyo-connectors/src/filesystem/iceberg.rs
+++ b/crates/arroyo-connectors/src/filesystem/iceberg.rs
@@ -74,13 +74,13 @@ impl Connector for IcebergConnector {
     ) -> anyhow::Result<Connection> {
         let schema = schema
             .map(|s| s.to_owned())
-            .ok_or_else(|| anyhow!("no schema defined for DeltaLake connection"))?;
+            .ok_or_else(|| anyhow!("no schema defined for Iceberg connection"))?;
 
         let format = schema
             .format
             .as_ref()
             .map(|t| t.to_owned())
-            .ok_or_else(|| anyhow!("'format' must be set for DeltaLake connection"))?;
+            .ok_or_else(|| anyhow!("'format' must be set for Iceberg connection"))?;
 
         let (description, connection_type, partition_fields) = match &table {
             IcebergTable::Sink(IcebergSink {
--- a/crates/arroyo-connectors/src/filesystem/mod.rs
+++ b/crates/arroyo-connectors/src/filesystem/mod.rs
@@ -58,8 +58,8 @@ pub fn make_sink(
             backend_config.is_local()
         }
         TableFormat::Iceberg { .. } => {
-            // for iceberg, there's no way for us to know the path (and whether it's local or not
-            // until we connect to the catalog, which we can't do until in a non-async context
+            // for iceberg, there's no way for us to know the path (and whether it's local or not)
+            // until we connect to the catalog, which we can't do in a non-async context
             // so for now we'll just support object storage
             // TODO: support local paths for iceberg
             false


crates/arroyo-connectors/src/filesystem/iceberg.rs
```diff
@@ -0,0 +1,162 @@
...
+        &self,
+        id: Option<i64>,
+        name: &str,
+        config: Self::ProfileT,
+        table: Self::TableT,
+        schema: Option<&ConnectionSchema>,
+    ) -> anyhow::Result<Connection> {
+        let schema = schema
+            .map(|s| s.to_owned())
+            .ok_or_else(|| anyhow!("no schema defined for DeltaLake connection"))?;
```
cmackenzie1: Iceberg?

crates/arroyo-connectors/src/filesystem/mod.rs
```diff
@@ -1,37 +1,108 @@
...
+    table_format: TableFormat,
+) -> Result<ConstructedOperator> {
+    let is_local = match table_format {
+        TableFormat::None | TableFormat::Delta => {
+            let backend_config = BackendConfig::parse_url(&sink.path, true)?;
+            backend_config.is_local()
+        }
+        TableFormat::Iceberg { .. } => {
+            // for iceberg, there's no way for us to know the path (and whether it's local or not
+            // until we connect to the catalog, which we can't do until in a non-async context
```
cmackenzie1:
```suggestion
            // until we connect to the catalog, which we can't do until in an async context
```

crates/arroyo-connectors/src/filesystem/sink/iceberg/mod.rs
```diff
@@ -0,0 +1,359 @@
...
+impl IcebergTable {
+    pub fn new(catalog: &IcebergCatalog, sink: &IcebergSink) -> anyhow::Result<Self> {
+        match catalog {
+            IcebergCatalog::Rest(rest) => {
+                let mut props = HashMap::new();
+                if let Some(token) = &rest.token {
+                    props.insert("token".to_string(), token.sub_env_vars()?);
+                }
+
+                println!("props {:?}", props);
```
cmackenzie1: This would dump the token if present?

crates/arroyo-connectors/src/filesystem/mod.rs
mwylde: should be "which we can't do in a non-async context" (until shouldn't be there)

crates/arroyo-connectors/src/filesystem/sink/iceberg/mod.rs
mwylde: removed!

devtyagi-c:
@mwylde hey, any idea when this is getting released?

PR #970, opened 2025-12-08
Structured error handling (part 1)

mwylde:
This PR is the start of a big migration for the Arroyo dataflow, from panic-based error handling to proper structured errors.

Currently, if something goes wrong in an operator it is expected to panic (possibly after reporting an error informationally to the controller). This has the effect of closing its queues. When an upstream subtask tries to send new data to it or  when a downstream subtask tries to read from it, that subtask will also panic. This has the effect of propagating the panic through the entire dataflow. 

These panics are detected by the engine, which reports them to the controller, which eventually shuts down the job and recovers it.

This actually works pretty well in practice and is very robust. However it has a few limitations:

* There is no automated way to distinguish different kinds of errors, particularly user-caused errors vs system-caused errors, or within the space of user-caused errors, between data-provoked errors (like a failed CAST) vs infrastructural errors
* The chain of panics creates a lot of noise in the logs, and it can be hard to find the root cause in all of the downstream panics
* In the future we may want to implement finer-grained recovery which is not possible if the whole system has shutdown automatically (inspired by [Flink's task-local recovery](https://cwiki.apache.org/confluence/display/FLINK/FLIP-1%3A+Fine+Grained+Recovery+from+Task+Failures))

So we're moving to structured error handling. Now each operator is expected to return an error rather than panicking, and errors will be handled centrally by the operator infrastructure and worker engine.

Since this is such a larger change, I'm splitting it up across several PRs. This first PR includes

* The new arroyo Error type, DataflowError
* Changes to the Operator and Source traits to return that error from every fallible method
* Partial updates to operators and connectors to return errors instead of panicking

Operator migrations have been partially completed, and will be continued in later PRs. There's also a fallback in place for anyhow, which will currently get promoted into a DataflowError. Eventually that will be removed and all uses of anyhow in operators will be replaced with a proper error.


crates/arroyo-rpc/src/errors.rs
```diff
@@ -0,0 +1,168 @@
...
+    #[error("bincode decode error: {0}")]
+    BincodeDecodeError(#[from] bincode::error::DecodeError),
+    #[error("bincode encode error: {0}")]
+    BincodeEncodeError(#[from] bincode::error::EncodeError),
+    #[error("protobuf decode error: {0}")]
+    ProtoDecodeError(#[from] prost::DecodeError),
+}
+
+#[derive(Debug, Clone)]
+pub struct UserError {
```
oliy, line 138: Are all user facing errors supposed to be instances of this implementation?  Seems like it should be an trait, with this as a default/generic implementation for simple ones?

oliy: Oh.  Nevermind.  Looks like most "user" errors are created using the ConnectorError macro, since they typically come from the connector implementation.  Other error types would just be custom enum from one of the other DataFlowError hierarchy.  Are these UserError errors getting phased out?  Might be good to have a extended explanation and examples beyond the basic ones for the connector_error macro.

mwylde: Yes, UserErrors are being removed, but it made the diff smaller in the short term to keep it around. This isn't a new struct, it's just moved so that it could be referenced by the new stuff in arroyo-rpc.

oliy (approved):

PR #980, opened 2025-12-20
Structured error handling (part 2)

mwylde:
This is the follow-on to #970. That PR introduced error types and started migrating operators. This PR builds on the earlier work to actually propagate failures through the system via the new error types and updating all of the operator infrastructure to work with errors.

The failure handling process now looks like this:
* An operator returns an Err(DataflowError) from one of its methods 
* This bubbles up to the runner code in the engine, which sends a TaskFailed RPC to the controller
* The controller decides, based on the retry hint, whether to attempt to retry the job (moving to Recovering) or fail it (a new Failing state, which attempts to cleanly shutdown the pipeline)
* In either case, The job is attempted to shut down by sending Stop RPCs to the workers, which will attempt to gracefully shutdown all tasks at which point the entire shuts down

In addition, the database schemas and API have been updated to return error domain information along with the cause of the error.

### DIFF
--- a/crates/arroyo-operator/src/operator.rs
+++ b/crates/arroyo-operator/src/operator.rs
@@ -161,7 +161,7 @@ impl OperatorNode {
                     &source_context.task_info,
                 );
 
-                s.operator.on_start(&mut source_context).await.unwrap();
+                s.operator.on_start(&mut source_context).await?;
 
                 ready.wait().await;
                 info!(
--- a/crates/arroyo-worker/src/arrow/async_udf.rs
+++ b/crates/arroyo-worker/src/arrow/async_udf.rs
@@ -1,7 +1,7 @@
 use anyhow::anyhow;
 use arrow::row::{OwnedRow, RowConverter, SortField};
 use arrow_array::{make_array, Array, RecordBatch, UInt64Array};
-use arrow_schema::{Field, Schema};
+use arrow_schema::{ArrowError, Field, Schema};
 use arroyo_datastream::logical::DylibUdfConfig;
 use arroyo_operator::context::{Collector, OperatorContext};
 use arroyo_operator::operator::{
@@ -265,10 +265,7 @@ impl ArrowOperator for AsyncUdfOperator {
                 .into_iter()
                 .map(|t| t.to_data())
                 .collect();
-            self.udf
-                .send(*id, args)
-                .await
-                .expect("failed to send data to async UDF runtime");
+            self.udf.send(*id, args).await?;
         }
 
         self.next_id = self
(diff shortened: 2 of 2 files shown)


crates/arroyo-operator/src/operator.rs
```diff
@@ -251,28 +247,50 @@ impl OperatorNode {
...
+                    "Task finished"
+                );
+
+                control_tx
+                    .send(ControlResp::TaskFinished {
+                        node_id: chain_info.node_id,
+                        task_index: chain_info.task_index as usize,
+                    })
+                    .await
+                    .expect("control response unwrap");
```
garvit-gupta, line 275: Could these expect statements be problematic in the future? Should we handle/improve these now while we are addressing error handling?

crates/arroyo-controller/src/states/mod.rs
```diff
@@ -384,10 +398,61 @@ impl JobContext<'_> {
...
+                source: anyhow!("task failed: {}", event.reason),
+                message: event.reason,
+                domain: event.error_domain,
+            },
+            errors::RetryHint::WithBackoff => StateError::RetryableError {
+                state,
+                source: anyhow!("task failed: {}", event.reason),
+                message: event.reason,
+                domain: event.error_domain,
+                retries: 20,
```
oliy, line 452: Maybe separate out into labelled constant?

crates/arroyo-rpc/src/errors.rs
```diff
@@ -68,20 +71,189 @@ pub enum DataflowError {
...
+    fn from(value: DataflowError) -> Self {
+        let message = value.to_string();
+        let (domain, retry_hint, details) = match value {
+            DataflowError::ConnectorError {
+                domain,
+                retry,
+                source,
+                ..
+            } => (domain, retry, source.as_ref().map(|s| format!("{s:?}"))),
+            DataflowError::ArrowError(_) => (ErrorDomain::Internal, RetryHint::NoRetry, None),
```
garvit-gupta, line 223: Would they always be internal errors? Any chance we need something similar to `classify_datafusion_error` here?

crates/arroyo-controller/src/job_controller/mod.rs
```diff
@@ -568,20 +563,17 @@ impl RunningJobModel {
...
             if let TaskState::Failed(reason) = &status.state {
-                error!(
-                    message = "task failed",
-                    job_id = *self.job_id,
-                    operator_id,
-                    subtask,
-                    reason,
-                );
-                return true;
+                return Some(reason.clone());
```
garvit-gupta, line 572: We seem to be returning early here. Is there a reason for this approach, rather than aggregating errors from all tasks (if there are multiple errors)?

crates/arroyo-operator/src/operator.rs
mwylde: These are still expected to panic; the error handling system relies on being able to report errors via these queues and so if sending to the queue fails (because the task on the other end has failed), there's nothing for us to do but panic at that point.

crates/arroyo-rpc/src/errors.rs
mwylde: We don't really have enough context to classify them at this point, but agreed we will likely be able to do better in the future.

crates/arroyo-controller/src/job_controller/mod.rs
mwylde: This code is handling the failure (driving the state machine), so it's just concerned with the fatal error that causes us to shut down the pipeline. Other errors will be reported in logs though.

Issue #976, opened 2025-12-16
Filesystem sink with partitioning panics in 0.15 (regression from 0.14)


szemyd:
Works in 0.14, fails in 0.15
`Panic at parquet.rs:272 with "No such file or directory"`

Simple sink without partitioning works fine

Adding `'partitioning.time_pattern'` or `'partitioning.fields'` causes the panic

macOS local filesystem path

Example:
```CREATE TABLE sink (
  window_start TIMESTAMP,
  total DOUBLE,
  exchange_count BIGINT,
  symbol TEXT
) WITH (
  connector = 'filesystem',
  type = 'sink',
  path = 'file:///<path_to_existing_directory>',
  format = 'parquet',
  'partitioning.time_pattern' = 'year=%Y/month=%m/day=%d/hour=%H'
);

INSERT INTO oi_sink
SELECT 
    window.start as window_start,
    total_oi,
    exchange_count,
    symbol
FROM (
    SELECT 
        TUMBLE(INTERVAL '1 minute') as window,
        value->0->>'symbol' as symbol,
        SUM(CAST(value->0->>'interest' AS DOUBLE)) as total,
        COUNT(*) as exchange_count
    FROM existingnatsconnection
    GROUP BY 
        TUMBLE(INTERVAL '1 minute'),
        value->0->>'symbol'
);```

**Comment by szemyd** [comment]
I've tried:
- Not including `  'partitioning.time_pattern' = 'year=%Y/month=%m/day=%d/hour=%H'`
- switching to deltalake
- downgrading 0.14 <- this works

**Comment by szemyd** [comment]
More of the error message:

```
2025-12-17T12:43:45.814475Z ERROR arroyo_controller::job_controller: task failed job_id="job_mekzoaqw4r" operator_id=6 subtask=0 reason="task 36 panicked with message \"
called `Result::unwrap()` on an `Err` value: Os { code: 2, kind: NotFound, message: \\\"No such file or directory\\\" }\""
```

GZKeeper:
@mwylde Would you be willing to publish a patch release, e.g. `v0.15.1`, that includes this fix? It would make it much easier for us to test and deploy without building a custom image from main.

PR #981, opened 2025-12-22
Fixed local file partitioning file not found issue #976

benamartin43:
Added code to create parent directories when partitioning is used.  I found that the code to automatically do this seems to have been removed when the iceberg partitioning was added (b0c83ea27f692b28ed25370f53818991effcb072).  I have based my code on the previous code.  Please let me know if I need to change the PR in any way.

I was able to reproduce #976 and verify the fix using the following query with the nexmark data source.
```sql
CREATE TABLE sink (
  window_start TIMESTAMP,
  total DOUBLE,
  bid_count BIGINT,
  name TEXT
) WITH (
  connector = 'filesystem',
  type = 'sink',
  path = 'file:////tmp/',
  format = 'parquet',
  'partitioning.time_pattern' = 'year=%Y/month=%m/day=%d/hour=%H/minute=%M'
);

INSERT INTO sink
SELECT 
    window.start as window_start,
    total,
    bid_count,
    name
FROM (
    SELECT 
        TUMBLE(INTERVAL '1 minute') as window,
        person.name as name,
        SUM(CAST(bid.price AS DOUBLE)) as total,
        COUNT(*) as bid_count
    FROM nexmark
    GROUP BY 
        TUMBLE(INTERVAL '1 minute'),
        person.name
);

```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -370,7 +370,7 @@ dependencies = [
 [[package]]
 name = "arrow-json"
 version = "55.2.0"
-source = "git+https://github.com/ArroyoSystems/arrow-rs?branch=55.2.0%2Fjson#0ae49da04b5c0ceb20cbd8e506fb3ceab3ebc219"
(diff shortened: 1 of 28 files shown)


mwylde (approved):
Apologies for the slow review and thanks for the fix!

PR #988, opened 2026-01-08
Retry getting state from remote store

oliy:


crates/arroyo/src/run.rs
```diff
@@ -246,12 +246,16 @@ impl MaybeLocalDb {
...
-                .get_if_present(REMOTE_STATE_KEY)
-                .await
-                .unwrap_or_else(|e| {
-                    panic!("Failed to fetch database from configured state directory '{s}': {e}");
-                });
+            let db = retry!(
+                provider.get_if_present(REMOTE_STATE_KEY).await,
+                10,
+                Duration::from_millis(100),
+                Duration::from_secs(2),
```
mwylde: I think it's worth making this a bit bigger -- as is we'll wait <20s which might not be enough time for the object store to recover from a transient issue

oliy: Maybe up to 5 minutes?

mwylde (approved):

PR #990, opened 2026-01-13
Fix tutorial link in README.md

wshirey:
The existing tutorial link is no longer available. I'm not sure what the original tutorial was, but I was interested in the "First pipeline" tutorial so that's what I updated the link to point to.


mwylde (approved):
Thanks!

PR #992, opened 2026-01-14
feat: add ignore_state option to pipeline restart

garvit-gupta:
Enables recovery from corrupted operator state by restarting pipelines without loading checkpoints.

When pipelines fail due to corrupted state, restarting normally restores the corrupted state and fails again. This adds ignore_state flag to restart API that skips checkpoint loading (restore_epoch=None), allowing recovery with fresh state.

Implementation uses ignore_state_before_epoch threshold: API queries MAX(checkpoint epoch) and sets threshold = max + 1. Controller only restores checkpoints with epoch >= threshold. This ensures reliable recovery across retries and respects config/status separation (controller reads config, never mutates it).

### DIFF
--- a/crates/arroyo-api/sqlite_migrations/V5__ignore_state_before_epoch.sql
+++ b/crates/arroyo-api/sqlite_migrations/V5__ignore_state_before_epoch.sql
@@ -1,0 +1,2 @@
+ALTER TABLE job_configs
+ADD COLUMN ignore_state_before_epoch integer;
--- a/crates/arroyo-api/src/pipelines.rs
+++ b/crates/arroyo-api/src/pipelines.rs
@@ -777,11 +777,24 @@ pub async fn restart_pipeline(
         RestartMode::safe
     };
 
+    // If user wants to ignore state, query max checkpoint epoch and compute threshold
+    let ignore_before_epoch = if req.ignore_state.unwrap_or(false) {
+        api_queries::fetch_max_checkpoint_epoch(&db, &job_id, &auth_data.organization_id)
+            .await?
+            .into_iter()
+            .next()
+            .and_then(|r| r.max_epoch)
+            .map(|max_epoch| max_epoch + 1)
+    } else {
+        None
+    };
+
     let res = api_queries::execute_restart_job(
         &db,
         &OffsetDateTime::now_utc(),
         &auth_data.user_id,
         &mode,
+        &ignore_before_epoch,
         &job_id,
         &auth_data.organization_id,
     )
--- a/crates/arroyo-controller/src/states/scheduling.rs
+++ b/crates/arroyo-controller/src/states/scheduling.rs
@@ -332,29 +332,64 @@ impl State for Scheduling {
             needs_commits: bool,
         }
 
-        let checkpoint_info = controller_queries::fetch_last_successful_checkpoint(
-            &ctx.db.client().await.unwrap(),
-            &*ctx.config.id,
-        )
-        .await
-        .unwrap()
-        .into_iter()
-        .next()
-        .map(|r| {
-            info!(
-                message = "restoring checkpoint",
-                job_id = *ctx.config.id,
-                epoch = r.epoch,
-                min_epoch = r.min_epoch
-            );
+        let checkpoint_info = {
+            let client = match ctx.db.client().await {
+                Ok(c) => c,
+                Err(e) => {
+                    return Err(ctx.retryable(self, "failed to get db client", e.into(), 10));
+                }
+            };
 
-            CheckpointInfo {
-                epoch: r.epoch as u32,
-                min_epoch: r.min_epoch as u32,
-                id: r.pub_id,
-                needs_commits: r.needs_commits,
-            }
-        });
+            let checkpoint_result = match controller_queries::fetch_last_successful_checkpoint(
+                &client,
+                &*ctx.config.id,
+            )
+            .await
+            {
+                Ok(r) => r,
+                Err(e) => {
+                    return Err(ctx.retryable(
+                        self,
+                        "failed to fetch last checkpoint",
+                        e.into(),
+                        10,
+                    ));
+                }
+            };
+
+            checkpoint_result.into_iter().next().and_then(|r| {
+                // Filter checkpoint based on ignore_state_before_epoch threshold
+                let should_restore = ctx
+                    .config
+                    .ignore_state_before_epoch
+                    .map(|threshold| r.epoch >= threshold)
+                    .unwrap_or(true); // If no threshold, restore any checkpoint
+
+                if should_restore {
+                    info!(
+                        message = "restoring checkpoint",
+                        job_id = *ctx.config.id,
+                        epoch = r.epoch,
+                        min_epoch = r.min_epoch
+                    );
+
+                    Some(CheckpointInfo {
+                        epoch: r.epoch as u32,
+                        min_epoch: r.min_epoch as u32,
+                        id: r.pub_id,
+                        needs_commits: r.needs_commits,
+                    })
+                } else {
+                    info!(
+                        message = "skipping checkpoint due to ignore_state_before_epoch threshold",
+                        job_id = *ctx.config.id,
+                        checkpoint_epoch = r.epoch,
+                        threshold = ctx.config.ignore_state_before_epoch.unwrap()
+                    );
+                    None
+                }
+            })
+        };
 
         {
             // mark in-progress checkpoints as failed
@@ -362,13 +397,33 @@ impl State for Scheduling {
                 .as_ref()
                 .map(|checkpoint_info| checkpoint_info.epoch)
                 .unwrap_or(0);
-            controller_queries::execute_mark_failed(
-                &ctx.db.client().await.unwrap(),
+
+            let client = match ctx.db.client().await {
+                Ok(c) => c,
+                Err(e) => {
+                    return Err(ctx.retryable(
+                        self,
+                        "failed to get db client for mark_failed",
+                        e.into(),
+                        10,
+                    ));
+                }
+            };
+
+            if let Err(e) = controller_queries::execute_mark_failed(
+                &client,
                 &*ctx.config.id,
                 &(last_epoch as i32 + 1),
             )
             .await
-            .unwrap();
+            {
+                return Err(ctx.retryable(
+                    self,
+                    "failed to mark in-progress checkpoints as failed",
+                    e.into(),
+                    10,
+                ));
+            }
         }
 
         let mut committing_state = None;
--- a/webui/src/routes/pipelines/PipelineDetails.tsx
+++ b/webui/src/routes/pipelines/PipelineDetails.tsx
@@ -4,6 +4,12 @@ import { useParams } from 'react-router-dom';
 import {
   Alert,
   AlertDescription,
+  AlertDialog,
+  AlertDialogBody,
+  AlertDialogContent,
+  AlertDialogFooter,
+  AlertDialogHeader,
+  AlertDialogOverlay,
   AlertIcon,
   AlertTitle,
   Badge,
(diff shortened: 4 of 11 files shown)


mwylde (changes requested):
Great first pass at this. One issue: I believe that as currently implemented, this will cause the pipeline to persistently ignore state, i.e., if the pipeline restarts on its own it will restart without state.

What we want is to just ignore state once—when the user initiates the restart.

crates/arroyo-controller/src/states/scheduling.rs
```diff
@@ -332,29 +332,46 @@ impl State for Scheduling {
...
-                needs_commits: r.needs_commits,
-            }
-        });
+            // Reset ignore_state immediately so it only applies once
+            controller_queries::execute_reset_ignore_state(
+                &ctx.db.client().await.unwrap(),
+                &*ctx.config.id,
+            )
+            .await
+            .unwrap();
```
mwylde: We shouldn't be unwrapping in the controller as that causes the entire state machine to panic. Instead return a state error (in this case, retryable) so that it can recover. I see that the existing code also has an unwrap which could be improved as well.

crates/arroyo-controller/src/states/scheduling.rs
```diff
@@ -332,29 +332,46 @@ impl State for Scheduling {
...
 
-            CheckpointInfo {
-                epoch: r.epoch as u32,
-                min_epoch: r.min_epoch as u32,
-                id: r.pub_id,
-                needs_commits: r.needs_commits,
-            }
-        });
+            // Reset ignore_state immediately so it only applies once
+            controller_queries::execute_reset_ignore_state(
```
mwylde: This approach will generally work but won't be completely reliable. In particular, if we fail later in scheduling we'll restart the scheduling process but the flag will have already been cleared, so we'll end up recovering with state.

webui/src/routes/pipelines/PipelineDetails.tsx
```diff
@@ -300,6 +308,57 @@ export function PipelineDetails() {
         {inner}
       </Grid>
       {configModal}
+      <AlertDialog
```
mwylde, line 332: I think instead of adding the confirmation dialog to all restarts, we instead want to give the option of restarting without state and confirming that. For example, you can look at the pattern we use for previewing, where hovering over the preview button pops up an option dialog:

Image

Separate nit: alert dialogs should be vertically centered.

webui/src/routes/pipelines/PipelineDetails.tsx
```diff
@@ -300,6 +308,57 @@ export function PipelineDetails() {
...
+              Restart Pipeline
+            </AlertDialogHeader>
+
+            <AlertDialogBody>
+              Are you sure you want to restart this pipeline?
+              <FormControl mt={4}>
+                <Checkbox isChecked={ignoreState} onChange={e => setIgnoreState(e.target.checked)}>
+                  Restart without state
+                </Checkbox>
+                <FormHelperText fontSize="xs" color="orange.500">
```
mwylde: I wouldn't make this xs — it looks a bit funky. Generally default to using the standard styles for UI elements.

webui/src/routes/pipelines/PipelineDetails.tsx
```diff
@@ -300,6 +308,57 @@ export function PipelineDetails() {
...
+                  This will permanently discard all checkpointed state. Use with caution.
```
mwylde: Technically it doesn't permanently discard the state, which will still be in object storage. After making the above change such that the confirmation is purely about restarting without state, I'd make the warning something like

"Restarting without state could lead to data loss, duplication, and incorrect results. Are you sure you want to continue?"

mwylde (changes requested):

crates/arroyo-controller/src/states/scheduling.rs
garvit-gupta: Added retryable error emission to the unwraps in `fetch_last_successful_checkpoint`, `execute_mark_failed` and `execute_reset_ignore_state` and their corresponding db client fetches.

crates/arroyo-controller/src/states/scheduling.rs
garvit-gupta: Good catch. I moved `execute_reset_ignore_state` to the end of the scheduling process. That way it should only reset the flag when we are ready to transition over from the scheduling state to running.

webui/src/routes/pipelines/PipelineDetails.tsx
garvit-gupta: Using the PopOver pattern now.

Screenshot 2026-01-20 at 10 58 52

webui/src/routes/pipelines/PipelineDetails.tsx
garvit-gupta: Screenshot 2026-01-20 at 10 59 07

Fixed

crates/arroyo-controller/src/states/scheduling.rs
```diff
@@ -332,43 +332,87 @@ impl State for Scheduling {
...
-                job_id = *ctx.config.id,
-                epoch = r.epoch,
-                min_epoch = r.min_epoch
+                message = "ignore_state flag set, starting without checkpoint",
+                job_id = *ctx.config.id
             );
+            None
+        } else {
+            let client = match ctx.db.client().await {
+                Ok(c) => c,
```
mwylde, line 337: Instead of using a match block for all of these, you can use map_err and ? to return the error. This also allows you to define the error mapping logic once in a function instead of having to repeat it each time.

crates/arroyo-controller/src/states/scheduling.rs
```diff
@@ -654,6 +698,29 @@ impl State for Scheduling {
...
+                        self,
+                        "failed to get db client for reset",
+                        e.into(),
+                        10,
+                    ));
+                }
+            };
+
+            if let Err(e) =
+                controller_queries::execute_reset_ignore_state(&client, &*ctx.config.id).await
```
mwylde: Unfortunately, this still will be unreliable because we might still fail before making our first checkpoint, in which case we will recover from the previous checkpoint.

Beyond that particular issue, in the database, we have a philosophical split between two types of data: job configs—which are set by the user and specify a desired config, and job statuses—which are set by the system and reflect the current state of the job. (This distinction is derived from the controller pattern in control planes). As a general rule: the controller reads config, and writes state.

With the current design, we're violating that split, because the controller is now overwriting the user's config. 

So we want to find a way to transmute this idea—restart without state on *this* particular restart—as config, without needing to clear out the config.

One idea: we have a current checkpoint epoch that we are attempting not to restore from. So what if we instead store the next epoch in something like "ignore_state_before_epoch"—then the logic becomes stateless; we look up the newest epoch and only restore from it if it's after that value.

webui/src/routes/pipelines/PipelineDetails.tsx
```diff
@@ -253,13 +270,29 @@ export function PipelineDetails() {
...
     if (job.state == 'Failed') {
       actionButton = (
-        <Button
-          onClick={async () => {
-            await restartPipeline();
-          }}
-        >
-          Restart
-        </Button>
+        <Popover trigger="hover" placement="bottom-start">
```
mwylde, line 273: I was imagine something more literally like the preview UI:

* Clicking the restart button restarts with state (in other words, operates exactly like it does currently)
* Hovering opens a popover that lets you restart without state

webui/src/routes/pipelines/PipelineDetails.tsx
```diff
@@ -253,13 +270,29 @@ export function PipelineDetails() {
...
+          <PopoverTrigger>
+            <Button>Restart</Button>
+          </PopoverTrigger>
+          <PopoverContent width="auto">
+            <PopoverArrow />
+            <PopoverBody p={2}>
+              <Stack spacing={1}>
```
mwylde: This should have more spacing—see how other things are spaced in the UI

webui/src/routes/pipelines/PipelineDetails.tsx
```diff
@@ -253,13 +270,29 @@ export function PipelineDetails() {
...
+                <Button
+                  size="sm"
+                  onClick={async () => {
+                    await restartPipeline(false);
+                  }}
+                >
+                  Restart
+                </Button>
+                <Button size="sm" colorScheme="orange" onClick={onRestartWithoutStateModalOpen}>
```
mwylde: I'm not a fan of the orange — we don't use it anywhere else so it looks out of place here

webui/src/routes/pipelines/PipelineDetails.tsx
```diff
@@ -300,6 +333,43 @@ export function PipelineDetails() {
...
+                results. Are you sure you want to continue?
+              </Text>
+            </AlertDialogBody>
+
+            <AlertDialogFooter>
+              <Button ref={cancelRef} onClick={onRestartWithoutStateModalClose}>
+                Cancel
+              </Button>
+              <Button
+                colorScheme="orange"
```
mwylde: same here with the orange

crates/arroyo-controller/src/states/scheduling.rs
garvit-gupta: That is the pattern I had attempted to use when I approached this task. But the `.map_err(|e| ctx.retryable(self, ...))` pattern moves self into the closure. Since self is moved, it leads to compilation errors because we need self at the end to transition to the next state (`Transition::next(*self, Running {})`).

We can potentially create a fresh `Box::new(Scheduling {})` for each error instead of reusing the original self. Scheduling is just an empty marker anyway, and this lets us stay consistent with the .map_err() style. We could also derive the `Clone` trait for Scheduling, which may be fine until Scheduling remains an empty struct.

webui/src/routes/pipelines/PipelineDetails.tsx
garvit-gupta: Makes sense. Updated.

webui/src/routes/pipelines/PipelineDetails.tsx
garvit-gupta: Made it red to make it appear like a warning. Is that better?

crates/arroyo-controller/src/states/scheduling.rs
garvit-gupta: Thank you for the suggestion! Updated the ignore logic to make it resilient to checkpoint failures, and to maintain the config-status split in the controller design.

crates/arroyo-api/sqlite_migrations/V5__ignore_state_before_epoch.sql
mwylde, line 1: This file should be V5__ignore_state_before_epoch.sql

crates/arroyo-controller/src/states/scheduling.rs
mwylde: ah, that's a good point. might be worth writing a macro at some point for this as the the amount of error handling code is now dwarfing the actual code, but this is fine for now

crates/arroyo-api/src/pipelines.rs
```diff
@@ -777,11 +777,29 @@ pub async fn restart_pipeline(
...
 
+    // If user wants to ignore state, query max checkpoint epoch and compute threshold
+    let ignore_before_epoch = if req.ignore_state.unwrap_or(false) {
+        let max_epoch =
+            api_queries::fetch_max_checkpoint_epoch(&db, &job_id, &auth_data.organization_id)
+                .await?
+                .into_iter()
+                .next()
+                .and_then(|r| r.max_epoch)
+                .unwrap_or(0); // If no checkpoints exist, start from epoch 0
```
mwylde: if no checkpoints exist, we can use None — there's no checkpoint to restore from anyways so there's no difference between restore from checkpoint or not

garvit-gupta: Sure.

PR #997, opened 2026-01-31
Improve single-file uploads by not sending contents over the commit rpcs

mwylde:
In #989, alongside the v2 filesystem sink, we added a new feature: the ability to write single files to object storage, instead of multiparts. Why would we want this? Well, sometimes you just need to write a MB or two, and it's wasteful to do three API calls (initialize multipart, upload part, close multipart) when one could do.

But although single file uploads _sound_ simpler, there's actually some complexity here. In particular, we rely on multipart uploads as a convenient intermediate state store between checkpoints and commits. In other words, during checkpointing phase, we can upload all of the parts (clearing them from our local state), then to commit them we just need to send some metadata since all of the actual data has already been uploaded.

Why does this matter? We wanted to provide transactional/exactly-once guarantees for raw file uploads, without requiring a table format like delta or iceberg. But without some other consistency mechanism, the only "commit" operation is the actual file write. That means we need to defer it until the commit phase, because if we were to write it during checkpointing and then crashed, we would recover and might end up with duplicate or incorrect data.

But with single files, we no longer have the ability to effectively "stage" our data in object storage. The initial implementation worked around this by sending the in-progress single files as part of the commit messages. This works, but introduces a new problem: the commit messages are sent to the controller over gRPC and then broadcasted to all of the workers, again over gRPC. But sending actual file contents risked exceeding the max gRPC message size and was in any case quite wasteful.

This PR addresses that issue. We take two different strategies depending on whether we're using a table format or not:

1. If no table format, we store the in-progress single files in memory at checkpoint time and write them to state. In commit, each subtask is responsible for writing its own files by reading them from memory. This avoids writing them before the commit phase and making them available before we are sure they are the correct files to write.
2. If iceberg/delta, we write the files during checkpoint, then commit them to the catalog during the commit phase. We can't write the files in commit, because each subtask only has access to its own files, but we need some coordination to not commit to the catalog until all files have been written. (Note this does risk orphaning files on recovery, but this should be relatively rare and can be cleaned up by standard iceberg maintenance procedures).

Note that this is a breaking change for v2 state, hopefully the last one.

### DIFF
--- a/crates/arroyo-connectors/src/filesystem/sink/v2/mod.rs
+++ b/crates/arroyo-connectors/src/filesystem/sink/v2/mod.rs
@@ -438,7 +438,7 @@ impl<BBW: BatchBufferingWriter + Send + 'static> ArrowOperator for FileSystemSin
         if ctx.task_info.task_index == 0 {
             for s in state.into_values() {
                 for f in s.open_files {
-                    debug!(path = f.path, buffered_size = f.data.len(),
+                    debug!(path = f.path, buffered_size = f.data.0.len(),
                         state = ?f.state, "recovering and finishing open file");
 
                     let mut open_file = OpenFile::from_checkpoint(
--- a/crates/arroyo-connectors/src/filesystem/sink/v2/open_file.rs
+++ b/crates/arroyo-connectors/src/filesystem/sink/v2/open_file.rs
@@ -11,8 +11,8 @@ use crate::filesystem::sink::{
     BatchBufferingWriter, FsEventLogger, MultiPartWriterStats, split_into_parts,
 };
 use arrow::record_batch::RecordBatch;
-use arroyo_rpc::connector_err;
 use arroyo_rpc::errors::DataflowResult;
+use arroyo_rpc::{SerializableBytes, connector_err};
 use arroyo_storage::StorageProvider;
 use bytes::Bytes;
 use futures::FutureExt;
@@ -41,7 +41,7 @@ impl PendingSingleFile {
         InProgressFile {
             path: self.path.to_string(),
             total_size: self.bytes.len(),
-            data: self.bytes.to_vec(),
+            data: SerializableBytes(self.bytes.clone()),
             metadata: self.iceberg_metadata.clone(),
             state: InProgressFileState::SingleFileToFinish,
         }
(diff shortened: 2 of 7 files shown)


crates/arroyo-connectors/src/filesystem/sink/v2/mod.rs
```diff
@@ -629,14 +637,27 @@ impl<BBW: BatchBufferingWriter + Send + 'static> ArrowOperator for FileSystemSin
...
+                CommitPreparation::LocalCommit(pending) => {
+                    open_files.push(pending.as_checkpoint());
+                    self.active.pending_local_commits.push(pending);
+                }
+            }
+        }
+
+        // Wait for PerOperator single file uploads to complete
+        while let Some(result) = upload_futures.next().await {
+            files_to_commit.push(result?);
```
QnJ1c2kNCg, line 660: How good is our error traceability today? If `result` was an `Err` would it be easy to trace back?

crates/arroyo-connectors/src/filesystem/sink/v2/open_file.rs
```diff
@@ -21,6 +25,59 @@ use std::sync::Arc;
...
+    logger: FsEventLogger,
+}
+
+impl PendingSingleFile {
+    /// Returns checkpoint data for this pending file
+    pub fn as_checkpoint(&self) -> InProgressFile {
+        InProgressFile {
+            path: self.path.to_string(),
+            total_size: self.bytes.len(),
+            data: self.bytes.to_vec(),
```
QnJ1c2kNCg: Are we concerned about the performance impact of copying the data here?

mwylde: Good call out. We were copying because this struct needs to be serialized/deserialized with bincode, and Bytes (our reference-counted bytearray type) doesn't implement bincode traits, and due to the orphan rule we can't implement them ourselves. 

However, that's solvable by wrapping Bytes in a newtype and implementing the traits on that. I've gone ahead and taken that approach which removes this copy (although we still have to copy on deserialization due to how bincode works).

crates/arroyo-connectors/src/filesystem/sink/v2/mod.rs
mwylde: The errors are indicating where they come from, which I think is more useful than where they are expressed. We have a bunch of places where these futures might be polled, but the error come from the future itself and is not a property of where that future happened to have been polled.

QnJ1c2kNCg: Make sense!

crates/arroyo-connectors/src/filesystem/sink/v2/open_file.rs
QnJ1c2kNCg: Nice!

PR #1003, opened 2026-02-04
Preserve epoch continuity when restarting without state

garvit-gupta:
Context
When using "restart without state", Arroyo sets `ignore_state_before_epoch = max_epoch + 1`. The pipeline then starts fresh creating checkpoints 1, 2, 3... But since these are all below the threshold, every subsequent start also ignores state until checkpoints eventually exceed the threshold.

Fix
Use `ignore_state_before_epoch - 1` as the default starting epoch when no checkpoint is restored. New checkpoints then start at/above the threshold and can be restored on subsequent starts.

### DIFF
--- a/crates/arroyo-controller/src/states/scheduling.rs
+++ b/crates/arroyo-controller/src/states/scheduling.rs
@@ -697,15 +697,34 @@ impl State for Scheduling {
             .await
             .insert(ctx.config.id.clone(), metrics.clone());
 
+        // Use ignore_state_before_epoch as default so new checkpoints exceed the threshold
+        let default_epoch = ctx
+            .config
+            .ignore_state_before_epoch
+            .filter(|&t| t > 0)
+            .map(|t| {
+                let epoch = (t - 1) as u32;
+                info!(
+                    message = "starting from ignore_state_before_epoch threshold",
+                    job_id = *ctx.config.id,
+                    default_epoch = epoch,
+                );
+                epoch
+            })
+            .unwrap_or(0);
+
         let mut controller = JobController::new(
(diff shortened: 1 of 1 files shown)


crates/arroyo-controller/src/states/scheduling.rs
```diff
@@ -697,15 +697,34 @@ impl State for Scheduling {
             .await
             .insert(ctx.config.id.clone(), metrics.clone());
 
+        // Use ignore_state_before_epoch as default so new checkpoints exceed the threshold
+        let default_epoch = ctx
+            .config
+            .ignore_state_before_epoch
+            .filter(|&t| t > 0)
+            .map(|t| {
+                let epoch = (t - 1) as u32;
```
QnJ1c2kNCg, line 706: I'm not sure I understand the `- 1`. Wouldn't the `- 1` potentially overwrite an old checkpoint?

garvit-gupta: No, it won't overwrite. `t` here is `max_epoch + 1` (set by the API handler). So `t - 1 = max_epoch`, which becomes the initial epoch in the JobController. Since start_checkpoint does `self.epoch += 1` before creating a checkpoint, the first new checkpoint lands at exactly at `max_epoch + 1`, not at max_epoch. The old checkpoint is untouched.

QnJ1c2kNCg: Got it! Thanks for explaining!

QnJ1c2kNCg (approved):

QnJ1c2kNCg (approved):

PR #1007, opened 2026-02-09
Add support for Gzip compression for the JSON format on Filesystem connector (V1 and V2)

QnJ1c2kNCg:
### Description
This patch adds support for compression when using the JSON format with the Filsystem connector (V1 and V2). The compression algorithm can be specified using the `json.compression` configuration knob.

At the moment, only `gzip` or `uncompressed` are supported; the default is `uncompressed`.

Note that local filesystem is currently **not** supported.

### Context

I'd like to share some context on the implementation choices.

**Streaming compression versus multi-member Gzip**
Intuitively, we would use a streaming compressor here. As we add data to the `JsonWriter` (which is a `BatchBufferingWriter`), we would compress and keep only one buffer in memory, containing the compressed data. The issue with this approach is that Gzip is stateful and we need to store a _trailer_ at the end of the compressed buffer. This means that in order to play well with the checkpointing logic, we would either have to store the state of the compressor in the checkpoint, or apply the footer at each checkpoint.

This leads us to the approach used in this PR: generate a multi-member Gzip. We now keep two buffers in memory, one for compressed data and one with uncompressed data. The writer will start compressing when it accumulates 5MB of uncompressed data or when it is asked to prepare a checkpoint. The data being checkpointed is compressed with the footer, thus doesn't require any particular handling upon recovery. This means that the files produced will be multi-member Gzip, those are common and easy to extract with utilities like `gunzip`.

I'm still not convinced that this is the right approach, but I think the discussion will be more productive with a PR.

**MPU minimum size**
A note is that an MPU part has a [minimum size of 5MB](https://docs.aws.amazon.com/AmazonS3/latest/userguide/qfacts.html).

### Testing
Added some unit tests.

Tested manually locally with various permutations of: uncompressed, compressed, Filesystem V1 and V2, large amount of data (to ensure that we were actually uploading MPU parts before the file got rolled). My manual tests were using and Impulse source and MinIO as a sink. For all tests, I validated that the produced `json.gz` was extractable.

### Next steps
* Align on the approach in this PR
* (If we want) Add support for the local filesystem (it uses a different code path)
* Add more compression options
* Update docs

### Other notes
* The code in this PR uses `panic`s... which I'm not a fan of, but I didn't want to refactor the entire trait as part of this PR. If we agree that the panics aren't ideal, then I propose to refactor the trait in its separate PR.
* I would love us to add **more tests**. I'm still not the most familiar with our various test types, but I think we could leverage some type of in-memory integration tests to validate this change and make sure we don't regress.


crates/arroyo-connectors/src/filesystem/sink/json.rs
```diff
@@ -7,14 +7,42 @@ use crate::filesystem::config;
...
+use flate2::Compression as GzipCompression;
+use flate2::write::GzEncoder;
 use std::time::Duration;
 use std::{fs::File, io::Write, time::Instant};
+use tracing::debug;
+
+/// Size of uncompressed chunks to compress at once.
+/// Larger chunks = better compression ratio, but more memory usage.
+/// 5MB provides good compression while keeping memory reasonable.
+const COMPRESSION_CHUNK_SIZE: usize = 5 * 1024 * 1024; // 5MB
```
QnJ1c2kNCg: RFC on this. Both the actual number, and if this should a configuration knob we can dynamically play with.

crates/arroyo-connectors/src/filesystem/sink/json.rs
```diff
@@ -56,23 +109,62 @@ impl BatchBufferingWriter for JsonWriter {
...
-        0
+        self.uncompressed_buffer.len()
     }
 
     fn buffered_bytes(&self) -> usize {
-        self.current_buffer.len()
+        // NOTE: Currently only retuns the compressed size, which is un under-estimate of the
+        // total buffered data. The uncompressed data that have up to [`COMPRESSION_CHUNK_SIZE`]
+        // bytes that are waiting to be compressed.
+        self.compressed_buffer.len()
```
garvit-gupta: I can see how the new impl of `unflushed_bytes` and `buffered_bytes` aligns with the expectations of the `BatchBufferingWriter`. But could there be any unintended/unsafe consequences of the change in this behavior?

For example, I had noticed that in `open_file.rs`, we compute the sum of unflushed bytes and the buffered bytes as a parameter to make the flush decision.

crates/arroyo-rpc/src/formats.rs
```diff
@@ -131,6 +158,10 @@ impl JsonFormat {
                 }
             });
 
+        let compression: JsonCompression = opts
```
garvit-gupta: Should we also have a check for the connector type, if we want only the filesystem connectors to support JSON compression?

crates/arroyo-connectors/src/filesystem/sink/json.rs
```diff
@@ -166,3 +260,217 @@ impl LocalWriter for JsonLocalWriter {
...
+    use arrow::array::{RecordBatch, StringArray};
+    use arrow::datatypes::{DataType, Field, Schema};
+    use arroyo_rpc::formats::JsonFormat;
+    use arroyo_types::TaskInfo;
+    use bytes::Bytes;
+    use flate2::read::{GzDecoder, MultiGzDecoder};
+    use std::io::Read;
+    use std::sync::Arc;
+
+    #[test]
```
garvit-gupta: Nice tests!

crates/arroyo-connectors/src/filesystem/sink/json.rs
QnJ1c2kNCg: Very nice catch, indeed I think there is an issue with the way `open_file.rs` uses these APIs.

Specifically, this snippet could be incorrect because the `unflushed_bytes` could greatly reduce once compressed and become smaller than the minimum MPU size.
```rust
writer.unflushed_bytes() + writer.buffered_bytes()
                        >= self.minimum_multipart_size
```

***
Overall, I feel like this implementation is a bit clunky with the rest of the existing traits. I propose to wait for @mwylde to review and comment on the approach.

crates/arroyo-rpc/src/formats.rs
QnJ1c2kNCg: AFAICT, there is already some type of validation (some me manually testing); but I will add an explicit check.

crates/arroyo-connectors/src/filesystem/sink/json.rs
```diff
@@ -26,23 +54,48 @@ impl BatchBufferingWriter for JsonWriter {
...
+            compression,
             event_logger,
         }
     }
-    fn suffix() -> String {
-        "json".to_string()
+
+    fn suffix_for_format(format: &Format) -> &str {
+        if let Format::Json(json) = format {
+            match json.compression {
```
mwylde: minor nit: it would be a bit cleaner to promote the match to the top-level rather than having both an if-let and match, like

```rust
        match format {
            Format::Json(JsonFormat { compression: Some(JsonCompression::Gzip), .. }) => "json.gz",
            Format::Json(_) => "json",
            _ => panic!("")
        }
```

crates/arroyo-connectors/src/filesystem/sink/json.rs
```diff
@@ -26,23 +54,48 @@ impl BatchBufferingWriter for JsonWriter {
...
+                Some(JsonCompression::Gzip) => "json.gz",
+                None => "json",
+            }
+        } else {
+            panic!("JsonLocalWriter configured with non-json format {format:?}");
```
mwylde: Should be JsonWriter

docs/architecture-diagrams.md
```diff
@@ -0,0 +1,826 @@
+# Arroyo Architecture Diagrams
```
mwylde: Did you intend to commit this file? There's a lot in here, and would take a good amount of time to review for accuracy.

AGENTS.md
```diff
@@ -0,0 +1,357 @@
+# Arroyo - Distributed Stream Processing Engine
```
mwylde: Let's leave this out of this PR, we can discuss including an agents file separately

crates/arroyo-rpc/src/formats.rs
```diff
@@ -62,6 +62,25 @@ impl TryFrom<&str> for DecimalEncoding {
     }
 }
 
+#[derive(Serialize, Deserialize, Copy, Clone, Debug, PartialEq, Eq, Hash, PartialOrd, ToSchema)]
+#[serde(rename_all = "snake_case")]
+pub enum JsonCompression {
```
mwylde, line 69: I think it's a bit cleaner to match the parquet behavior: instead of Option<JsonCompression>, have a JsonCompression::Uncompressed which avoids us having to handle two different levels of types and preserves options to mean unset

crates/arroyo-connectors/src/filesystem/sink/json.rs
```diff
@@ -56,23 +109,62 @@ impl BatchBufferingWriter for JsonWriter {
...
+    /// and appends them to compressed_buffer. Leaves remainder in uncompressed_buffer.
+    fn flush_uncompressed_to_compressed(&mut self) {
+        while self.uncompressed_buffer.len() >= COMPRESSION_CHUNK_SIZE {
+            let chunk = self
+                .uncompressed_buffer
+                .split_to(COMPRESSION_CHUNK_SIZE)
+                .freeze();
+            let compressed = compress_chunk(chunk, self.compression)
+                .expect("Failed to compress JSON chunk - this should never happen with valid data");
+            self.compressed_buffer.extend_from_slice(&compressed);
```
mwylde: Would it be more efficient to write directly into the buffer instead of writing to a new vec and then copying into the buffer?

crates/arroyo-connectors/src/filesystem/sink/json.rs
```diff
@@ -26,23 +54,48 @@ impl BatchBufferingWriter for JsonWriter {
...
             size += k.len() + 1;
-            self.current_buffer.extend(k);
-            self.current_buffer.extend(b"\n");
+            self.uncompressed_buffer.extend(k);
+            self.uncompressed_buffer.extend(b"\n");
         }
+
+        // Compress uncompressed buffer in chunks
+        self.flush_uncompressed_to_compressed();
+        debug!(
```
mwylde: I would make this trace — add_batch_data is called with high frequency

crates/arroyo-connectors/src/filesystem/sink/json.rs
mwylde: As Garvit said, we use this to determine whether or not to flush, so there's no correctness issue here (we just might take longer to flush than currently).

The behavior introduced here is the same as how it works with parquet; we buffer up some amount of data in the parquet writer which is not included in buffered_bytes until the parquet row group is flushed (on its own schedule).

However, as Garvit points out, the implementation of unflushed_bytes isn't really correct; we want an estimate of the compressed size, not the uncompressed size. The ParquetWriter gives us that estimate directly (via in_progress_size). Is there a way we can get a similar estimate for gzip?

mwylde (commented):
This is overall a reasonable approach. There are some inefficiencies (we're introducing several new copies, even for the no compression case), and the multi-member approach might cause compatibility issues for certain readers.

I think the ideal would be to do streaming compression while storing the internal state of the compressor in the checkpoint (similar to what we do for Parquet). This would however likely involve a fork of flate2 (like we had to do for arrow-parquet) which probably isn't worth it.

docs/architecture-diagrams.md
QnJ1c2kNCg: Whoops, I committed this file (and the `AGENTS.md` file) by accident when address the first round of comments.

The goal was for me to get an overview of the system quickly, acknowledging the presence of innaccuracies.

AGENTS.md
QnJ1c2kNCg: I committed this by accident, will remove in the next revision.

crates/arroyo-rpc/src/formats.rs
QnJ1c2kNCg: I personally preferred the `Option` it seemed more semantically accurate, but I don't have a strong opinion here; I'll update to the `JsonCompression::Uncompressed` approach.

crates/arroyo-connectors/src/filesystem/sink/json.rs
QnJ1c2kNCg: Good point, I will update.

crates/arroyo-connectors/src/filesystem/sink/json.rs
QnJ1c2kNCg: > Is there a way we can get a similar estimate for gzip?

I don't think there is a way provided by the library. We could store our own compression ratio, starting with a conservative number and then updating it based on the previous compressed chunk.

I don't think we can use a static compression ratio as there could be a high variance based on user data.

crates/arroyo-connectors/src/filesystem/sink/parquet.rs
```diff
@@ -52,12 +52,12 @@ fn writer_properties_from_format(format: &ParquetFormat) -> (WriterProperties, u
 /// A buffer with interior mutability shared by the [`ArrowWriter`] and
 /// [`AsyncArrowWriter`]. From Arrow. This lets us write data from the buffer to S3.
 #[derive(Clone)]
-struct SharedBuffer {
+pub(crate) struct SharedBuffer {
```
QnJ1c2kNCg, line 55: I'm using the same struct for the JSON implementation, happy to move or make more generalized if needed.

crates/arroyo-connectors/src/filesystem/sink/json.rs
```diff
@@ -60,19 +112,55 @@ impl BatchBufferingWriter for JsonWriter {
...
 
     fn close(&mut self) -> (Bytes, Option<IcebergFileMetadata>) {
-        (self.current_buffer.split().freeze(), None)
+        // Finish and close encoder if in compressed mode
+        if let Some(encoder) = self.encoder.take() {
+            encoder
+                .finish()
+                .expect("Failed to finish gzip encoder at close");
+            trace!("Finished final gzip member at close");
+        }
```
QnJ1c2kNCg: Here, I don't create a new encoder because we expect the caller to not add more data after calling `close()` but it might be a bit brittle, curious to hear what other folks think.

I wonder if we should make this function take `self` (instead of a reference) to prevent callers from using the writer after `close()` was called.

crates/arroyo-connectors/src/filesystem/sink/json.rs
```diff
@@ -26,23 +36,65 @@ impl BatchBufferingWriter for JsonWriter {
...
             size += k.len() + 1;
-            self.current_buffer.extend(k);
-            self.current_buffer.extend(b"\n");
+
+            match &mut self.encoder {
+                Some(encoder) => {
+                    // Compressed mode: write to streaming encoder (compresses immediately)
+                    encoder
+                        .write_all(&k)
+                        .expect("Failed to write to gzip encoder");
```
mwylde, line 125: Can these writes fail? If not, please add a comment explaining why.

crates/arroyo-connectors/src/filesystem/sink/json.rs
```diff
@@ -26,23 +36,65 @@ impl BatchBufferingWriter for JsonWriter {
...
+        } else {
+            panic!("JsonWriter configured with non-json format {format:?}");
+        };
+
+        let buffer = SharedBuffer::new(0);
+
+        let encoder = match compression {
+            JsonCompression::Uncompressed => None,
+            JsonCompression::Gzip => {
+                Some(GzEncoder::new(buffer.clone(), GzipCompression::default()))
```
mwylde: We're sharing the SharedBuffer between current_buffer and (potentially) within the GzEncoder, but really this is exclusive: either we need to manage the buffer directly or via the GzEncoder. I think this could lead to errors as users need to carefully maintain that invariant.

Rust gives us tools to prevent this sort of thing at the type level—I think it would be worth wrapping the buffer into an enum like this:

```rust
enum JsonBuffer {
    Uncompressed(BytesMut),
    Gzipped(GzEncoder<bytes::buf::Writer<BytesMut>>),
}
```

This avoids the need for locking and makes the logic for interacting with these cases simpler and safer

crates/arroyo-connectors/src/filesystem/sink/json.rs
mwylde: It's a good call out and probably would have been a better design. I think it would be a bit painful to retrofit this into filesystem v1, but isn't an issue for v2. Maybe worth doing once we remove v1.

mwylde (commented):
A few little comments but overall looking good!

crates/arroyo-connectors/src/filesystem/sink/json.rs
QnJ1c2kNCg: They cannot fail. I will add a block comment:
```
// Compressed mode: write to streaming encoder (compresses immediately).
// These writes are infallible: the inner SharedBuffer wraps a BytesMut
// whose Writer<BytesMut>::write() always returns `Ok`.
// Ref: https://docs.rs/crate/bytes/1.11.1/source/src/buf/writer.rs#78-83
```

crates/arroyo-connectors/src/filesystem/sink/json.rs
QnJ1c2kNCg: I like this! However, it requires a slight dance when `get_trailing_bytes_for_checkpoint()` is called. This is because we want to `finish()` the encoder, but have it still use the same underlying buffer. (and that `finish()` takes `self`).

It would look something like this:

```rust
/// The bytes stored in the checkpoint are bytes that we are ready to uploaded as-is.
/// For compressed mode, we finish the current gzip member to ensure all data is complete.
fn get_trailing_bytes_for_checkpoint(&mut self) -> (Vec<u8>, Option<IcebergFileMetadata>) {
    // Finish current gzip member and create a new encoder for the next member.
    let bytes = match &mut self.buffer {
        JsonBuffer::Uncompressed(bytes_mut) => bytes_mut.to_vec(),
        // In the Gzip case, we do a dance where we consume the encoder by calling`finish()`,
        // and then create a new `GzEncoder` using the same underlying buffer.
        JsonBuffer::Gzipped(encoder_slot) => {
            let old_encoder = std::mem::replace(
                encoder_slot,
                GzEncoder::new(BytesMut::new().writer(), GzipCompression::default()),
            );

            // finish() flushes remaining compressed data and the gzip footer into the buffer.
            let inner = old_encoder
                .finish()
                .expect("Failed to finish gzip encoder at checkpoint");

            let bytes = inner.get_ref().to_vec();

            // Reuse the buffer (retaining the data) for the new gzip member.
            *encoder_slot = GzEncoder::new(inner, GzipCompression::default());

            bytes
        }
    };

    (bytes, None)
}
```

I think your proposal is still a cleaner approach. What are your thoughts?

QnJ1c2kNCg: Discussed offline, I will go with this approach + an `Option`.

mwylde (approved):
Looks great!

PR #1008, opened 2026-02-10
Build `webui/` as part of `cargo build`

QnJ1c2kNCg:
**This change may be controversial, since we are coupling the Rust & JS toolchains. I personally don't have a strong opinion here.**

The `arroyo-api` crate requires our `webui/` to be built in order to compile. This patch adds a step in the `build.rs` of `arroyo-api` to check if `webui/` needs to be built. If so, it calls the `pnpm` commands needed to build the `webui/`.

### Testing
Before the change, `cargo build` would fail at the `arroyo-api` crate (since I didn't build the `webui/` component).
```
error: #[derive(RustEmbed)] folder '/Users/broy/github/arroyo-all/bug-hunting/crates/arroyo-api/../../webui/dist' does not exist. cwd: '/Users/broy/github/arroyo-all/bug-hunting'
  --> crates/arroyo-api/src/rest.rs:40:1
   |
40 | / #[folder = "../../webui/dist"]
41 | | struct Assets;
   | |______________^

error[E0599]: no function or associated item named `get` found for struct `Assets` in the current scope
  --> crates/arroyo-api/src/rest.rs:76:19
   |
41 | struct Assets;
   | ------------- function or associated item `get` not found for this struct
...
76 |     match Assets::get(path) {
   |                   ^^^ function or associated item not found in `Assets`
   |
   = help: items from traits can only be used if the trait is implemented and in scope
   = note: the following traits define an item `get`, perhaps you need to implement one of them:
```

After the change, `cargo build` works out of the box.

### DIFF
--- a/crates/arroyo-api/build.rs
+++ b/crates/arroyo-api/build.rs
@@ -1,10 +1,11 @@
 use cornucopia::{CodegenSettings, Error};
 use postgres::{Client, NoTls};
-use std::path::Path;
+use std::path::PathBuf;
 use std::process::Command;
 
 fn ensure_webui_built() {
-    let webui_dir = Path::new("../../webui");
+    let manifest_dir = PathBuf::from(std::env::var("CARGO_MANIFEST_DIR").unwrap());
+    let webui_dir = manifest_dir.join("../../webui");
     let dist_index = webui_dir.join("dist/index.html");
 
     // Only re-run this build script when dist/ changes (or is created).
@@ -12,7 +13,10 @@ fn ensure_webui_built() {
     // build toolchain. Developers who modify the frontend should run
     // `pnpm build` in webui/ themselves; this block only bootstraps
     // an initial build so that `cargo build` works out of the box.
-    println!("cargo:rerun-if-changed=../../webui/dist");
+    println!(
+        "cargo:rerun-if-changed={}",
+        webui_dir.join("dist").display()
+    );
 
     if dist_index.exists() {
         return;
@@ -35,7 +39,7 @@ fn ensure_webui_built() {
     // Install dependencies (frozen so the lockfile is never mutated by a Rust build)
     let install = Command::new("pnpm")
         .args(["install", "--frozen-lockfile"])
-        .current_dir(webui_dir)
+        .current_dir(&webui_dir)
         .status()
         .expect("failed to spawn `pnpm install`");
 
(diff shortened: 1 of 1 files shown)


crates/arroyo-api/build.rs
```diff
@@ -1,7 +1,73 @@
 use cornucopia::{CodegenSettings, Error};
 use postgres::{Client, NoTls};
+use std::path::Path;
+use std::process::Command;
+
+fn ensure_webui_built() {
+    let webui_dir = Path::new("../../webui");
```
garvit-gupta: Should we use `CARGO_MANIFEST_DIR` instead of a relative dir path?

crates/arroyo-api/build.rs
```diff
@@ -1,7 +1,73 @@
 use cornucopia::{CodegenSettings, Error};
 use postgres::{Client, NoTls};
+use std::path::Path;
+use std::process::Command;
+
+fn ensure_webui_built() {
```
garvit-gupta, line 6: Should we gate this under a cargo feature to allow for an opt-out behavior? In case a developer using these does not have pnpm or some other dependency involved?

crates/arroyo-api/build.rs
QnJ1c2kNCg: Good idea, let me update!

crates/arroyo-api/build.rs
QnJ1c2kNCg: Humm, I'm not sure. Today, there is already a dependency on having `webui/` built in order for a full `cargo build` to succeed. This patch just automates the building of the web components.

Folks are always able to build other crates without having to build the `arroyo-api` / `webui/` crate.


***
That being said, I get that this can be controversial and if you prefer we don't couple the build systems, I'm happy discarding this PR.

garvit-gupta: I think this is a good change to have. The systems are already coupled. Arroyo can't compile without webui today and this just automates what every dev has to figure out manually. I've hit this myself.

Re: the opt-out idea: I was thinking more about new users and CI/Docker scenarios where you might want fewer implicit dependencies, but if I think about it, both already build or would need to build the webui as a prior step, so this change doesn't really add dependencies or complexity. Happy to defer to you and the others on the next steps on this.

PR #1009, opened 2026-02-10
Fix various nitpicking/typos

QnJ1c2kNCg:
These were identified by Claude; but they all look like real typos/bugs to me.

### DIFF
--- a/crates/arroyo-api/src/connection_tables.rs
+++ b/crates/arroyo-api/src/connection_tables.rs
@@ -694,7 +694,7 @@ async fn expand_json_schema(
                         schema_response.0.schema_type
                     )));
                 }
-                confluent_schema_id.replace(schema_response.0.version);
+                confluent_schema_id.replace(schema_response.0.id);
 
                 schema.definition = Some(SchemaDefinition::JsonSchema {
                     schema: schema_response.0.schema,
--- a/crates/arroyo-controller/src/lib.rs
+++ b/crates/arroyo-controller/src/lib.rs
@@ -25,8 +25,6 @@ use arroyo_server_common::shutdown::ShutdownGuard;
 use arroyo_server_common::wrap_start;
 use arroyo_types::{MachineId, WorkerId, from_micros};
 use cornucopia_async::DatabaseSource;
-use lazy_static::lazy_static;
-use prometheus::{Gauge, register_gauge};
 use states::{Created, State, StateMachine};
 use std::collections::{HashMap, HashSet};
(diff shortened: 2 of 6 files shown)


crates/arroyo-controller/src/lib.rs
```diff
@@ -60,24 +58,6 @@ use types::public::{RestartMode, StopMode};
 
 pub const CHECKPOINTS_TO_KEEP: u32 = 5;
 
-lazy_static! {
```
garvit-gupta, line 63: Have these been unused after a recent change? Or were these never in use? If the former, maybe we should understand if these will be needed again in the future.

crates/arroyo-planner/src/tables.rs
```diff
@@ -408,7 +408,7 @@ impl ConnectorTable {
         table.idle_time = options
             .pull_opt_i64("idle_micros")?
             .or_else(|| DEFAULT_IDLE_TIME.map(|t| t.as_micros() as i64))
-            .filter(|t| *t <= 0)
+            .filter(|t| *t > 0)
```
garvit-gupta, line 411: Unrelated to this change, but`table.idle_time` is passed, but is never really consumed. Not sure if that is intentional.

crates/arroyo-controller/src/lib.rs
QnJ1c2kNCg: Git blame says those were added 3 years ago, and I couldn't find any reference in the codebase.

crates/arroyo-api/src/connection_tables.rs
```diff
@@ -694,7 +694,7 @@ async fn expand_json_schema(
                         schema_response.0.schema_type
                     )));
                 }
-                confluent_schema_id.replace(schema_response.0.version);
```
mwylde, line 697: This change seems incorrect — was it tested?

crates/arroyo-planner/src/tables.rs
mwylde: Your correct, it's meant to be passed into the watermark generator but it is not happening currently (we always set it to None).

crates/arroyo-api/src/connection_tables.rs
QnJ1c2kNCg: You're right, let me raise a PR for it.

QnJ1c2kNCg: Partial revert: #1016

PR #1010, opened 2026-02-10
Redact raw values from deserialization error logging

garvit-gupta:
Context
When logging user errors from JSON deserialization validation, the ValidationError details string included `actual_value` — the raw input data that failed validation. This could potentially contain sensitive details that ends up being reported to analytics.

Changes
Replace raw values with safe metadata (expected type, actual type, string length) at the log_event! boundary. Added unit tests to validate the expected error output.

### DIFF
--- a/crates/arroyo-formats/src/de.rs
+++ b/crates/arroyo-formats/src/de.rs
@@ -3,7 +3,7 @@ use crate::proto::schema::get_pool;
 use crate::{proto, should_flush};
 use arrow::array::{Int32Builder, Int64Builder};
 use arrow::compute::kernels;
-use arrow::json::reader::FailureKind;
+use arrow::json::reader::{FailureKind, JsonType, ValidationError};
 use arrow_array::builder::{
     ArrayBuilder, BinaryBuilder, GenericByteBuilder, StringBuilder, TimestampNanosecondBuilder,
     UInt64Builder, make_builder,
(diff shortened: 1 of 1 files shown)


crates/arroyo-formats/src/de.rs
```diff
@@ -126,6 +126,45 @@ fn failure_kind_to_str(kind: FailureKind) -> &'static str {
...
+                        format!("got {actual_type} (length: {})", val.len())
+                    }
+                    _ => format!("got {actual_type}"),
+                },
+                (Some(actual_type), None) => format!("got {actual_type}"),
+                _ => "got incompatible type".to_string(),
+            };
+
+            let suffix = if matches!(verr.failure_kind, FailureKind::ParseFailure) {
+                " — parse failure"
```
QnJ1c2kNCg: Why an em dash? :P

crates/arroyo-formats/src/de.rs
```diff
@@ -1045,4 +1084,161 @@ mod tests {
...
+            (
+                ValidationError {
+                    row_index: 0,
+                    field_path: "created_at".to_string(),
+                    failure_kind: FailureKind::ParseFailure,
+                    expected_type: Arc::new(DataType::Utf8),
+                    actual_type: Some(JsonType::String),
+                    actual_value: Some("\"not-a-date\"".to_string()),
+                },
+                "field 'created_at': expected Utf8, got string (length: 10) \u{2014} parse failure",
```
QnJ1c2kNCg: Why use the `\u{2014}` instead of the em dash directly?

Should we use regular dash? in case a client reads the log without UTF-8 support.

crates/arroyo-formats/src/de.rs
garvit-gupta: Oops. Replaced with a regular one.

PR #1019, opened 2026-02-20
Add output format and table format labels to filesystem sink trace events

garvit-gupta:
### Context
Filesystem sink trace events currently don't include the output format or table format. Inclusion of these events would enrich analytic capabilities to describe sink behavior.

### Changes
- Add `Format::name()` and `TableFormat::name()` returning `&'static str` to prevent string allocations
- Add `output_format` and `table_format` fields to `FsEventLogger`, emitted as labels in `filesystem_write` trace events

### DIFF
--- a/crates/arroyo-connectors/src/filesystem/mod.rs
+++ b/crates/arroyo-connectors/src/filesystem/mod.rs
@@ -40,6 +40,14 @@ pub enum TableFormat {
 }
 
 impl TableFormat {
+    pub fn name(&self) -> &'static str {
+        match self {
+            TableFormat::None => "none",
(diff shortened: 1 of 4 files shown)


rfcs/structured-error-aggregation.md
```diff
@@ -0,0 +1,89 @@
+# RFC: Structured Deserialization Error Aggregation
```
QnJ1c2kNCg: The two RFCs seems like they were used to help the agent, were they committed by mistake?

garvit-gupta: Fixed.

PR #1021, opened 2026-02-26
Local json compression + optimization for empty encoders

QnJ1c2kNCg:
This PR does two things:

1. [Add support for Gzip compression for the JsonLocalWriter](https://github.com/ArroyoSystems/arroyo/commit/62e6dfd93373388a0f92ea9bb49d7fa9563bb2a8)

Very similar to https://github.com/ArroyoSystems/arroyo/pull/1007, but for the Local file backend.

2. [Avoid finishing the encoder if no data was written](https://github.com/ArroyoSystems/arroyo/commit/b40ab7f0e6b004ab4a3506020002811ee13a3d93)

I noticed that the Gzip encoder would produce bytes (header + footer) on every checkpoint even when no user traffic. This patch tracks if we wrote user data to the underlying buffer, marking the buffer as `dirty`, and only finishing the encoder if this flag is set.

The `dirty` flag starts as `false` and is set to `true` when `write_all()` is called. The flag resets to `false` when a new encoder is created.

***
Tested locally a variety of scenarios:
* sink v1, v2 with minio
* sink v2 with local FS
* with and without compression
* with and without data (to test that we don't finish the encoder)
* crashes and recovery

### DIFF
--- a/crates/arroyo-connectors/src/filesystem/sink/json.rs
+++ b/crates/arroyo-connectors/src/filesystem/sink/json.rs
@@ -21,13 +21,15 @@ use std::{fs::File, io::Write, time::Instant};
 ///
 /// Generic over `W: Write` so it can wrap both in-memory buffers (`Writer<BytesMut>`)
 /// and file handles (`File`).
-///
-/// The `Gzipped` variant uses an `Option` so that the encoder can be temporarily
-/// taken out (via `.take()`) for operations that consume it (like `finish()`).
-/// In addition, `None` is used to indicate that the writer has been closed.
 enum JsonBuffer<W: Write> {
     Uncompressed(W),
-    Gzipped(Option<GzEncoder<W>>),
+    /// The `Gzipped` variant uses an `Option` so that the encoder can be temporarily
+    /// taken out (via `.take()`) for operations that consume it (like `finish()`).
+    /// In addition, `None` is used to indicate that the writer has been closed.
+    Gzipped {
+        encoder: Option<GzEncoder<W>>,
+        dirty: bool,
+    },
 }
 
 impl<W: Write> JsonBuffer<W> {
@@ -35,17 +37,14 @@ impl<W: Write> JsonBuffer<W> {
     fn write_all(&mut self, data: &[u8]) -> std::io::Result<()> {
         match self {
             JsonBuffer::Uncompressed(w) => w.write_all(data),
-            JsonBuffer::Gzipped(Some(encoder)) => encoder.write_all(data),
-            JsonBuffer::Gzipped(None) => panic!("write_all called after close()"),
-        }
-    }
-
-    /// Flushes the encoder or the inner writer.
-    fn flush(&mut self) -> std::io::Result<()> {
-        match self {
-            JsonBuffer::Uncompressed(w) => w.flush(),
-            JsonBuffer::Gzipped(Some(encoder)) => encoder.flush(),
-            JsonBuffer::Gzipped(None) => panic!("flush called after close()"),
+            JsonBuffer::Gzipped {
+                encoder: Some(enc),
+                dirty,
+            } => {
+                *dirty = true;
+                enc.write_all(data)
+            }
+            JsonBuffer::Gzipped { encoder: None, .. } => panic!("write_all called after close()"),
         }
     }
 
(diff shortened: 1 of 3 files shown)


cmackenzie1 (approved):

crates/arroyo-connectors/src/filesystem/sink/json.rs
```diff
@@ -507,4 +607,396 @@ mod tests {
...
+            partitioning: Default::default(),
+            multipart: Default::default(),
+            version: Default::default(),
+        };
+
+        JsonLocalWriter::new(tmp_path, final_path, &config, format, arroyo_schema)
+    }
+
+    #[test]
+    fn test_local_writer_suffix_for_format() {
```
mwylde: I think some of these tests can be removed — this one just duplicates the straightforward code. Tests should be validating meaningful functionality.

crates/arroyo-connectors/src/filesystem/sink/json.rs
```diff
@@ -259,29 +314,74 @@ impl LocalWriter for JsonLocalWriter {
...
         Ok(size)
     }
 
     fn close(&mut self) -> anyhow::Result<super::local::FilePreCommit> {
-        LocalWriter::sync(self)?;
+        match &self.buffer {
+            JsonBuffer::Gzipped { .. } => {
+                // Always finalize local gzip on close so file bytes cannot change later
+                // from encoder drop-time writes after we return.
+                let mut file = self.buffer.finish_encoder()?;
```
mwylde: Should this check the dirty flag? I'm not sure I understand the comment

crates/arroyo-connectors/src/filesystem/sink/json.rs
```diff
@@ -154,52 +178,63 @@ impl BatchBufferingWriter for JsonWriter {
...
-                    .expect("Failed to finish gzip encoder at checkpoint");
-
-                let bytes = inner.get_ref().to_vec();
-
-                // Reuse the buffer (retaining the data) for the new gzip member.
-                self.buffer
-                    .restore_encoder(GzEncoder::new(inner, GzipCompression::default()));
+            JsonBuffer::Uncompressed(inner) => inner.get_ref().to_vec(),
+            JsonBuffer::Gzipped { dirty, .. } => {
+                if *dirty {
```
mwylde: This logic gets repeated a bunch of times. Can we abstract it out?

crates/arroyo-connectors/src/filesystem/sink/json.rs
```diff
@@ -259,29 +314,74 @@ impl LocalWriter for JsonLocalWriter {
...
+                // Only flush the encoder if it contains data.
+                if *dirty {
+                    enc.flush()?;
+                }
+            }
+            JsonBuffer::Gzipped { encoder: None, .. } => panic!("sync called after close()"),
+            JsonBuffer::Uncompressed(w) => w.flush()?,
+        }
+        let size = self.buffer.writer_ref().metadata()?.len() as usize;
+        if let Some(stats) = &mut self.stats {
```
mwylde: this changed from an unwrap to an if let — but do we expect that stats can be None?

QnJ1c2kNCg: `stats` is only `None` when the struct is instantiated, it becomes `Some` after the first `write_batch` call. So in practice we wouldn't panic on the unwrap today.

I generally try to avoid using `.unwrap()` whenever possible. IMO, having unwraps can make the code harder to parse and harder to reason about "when can this panic?".

In this case, if you think that `stats` being `None` is a logic bug worth panicking I'm happy to revert.

crates/arroyo-connectors/src/filesystem/sink/json.rs
QnJ1c2kNCg: Agreed.

crates/arroyo-connectors/src/filesystem/sink/json.rs
QnJ1c2kNCg: Yep, I had a hard time finding the right abstraction but I think I got something slightly better in the next commit.

crates/arroyo-connectors/src/filesystem/sink/json.rs
QnJ1c2kNCg: I refactored a few things in the next commit, should be clearer.

crates/arroyo-connectors/src/filesystem/sink/json.rs
mwylde: Not a huge deal, but yes, if we expect stats should always be Some and it's a bug for it to be None (meaning we're incorrectly not collecting stats, which is a big deal for us) we should panic if it violates the expectation.

mwylde (approved):
Looks great!

PR #1024, opened 2026-03-05
Output full Iceberg catalog error

mwylde:


crates/arroyo-connectors/src/filesystem/sink/iceberg/mod.rs
```diff
@@ -125,7 +127,7 @@ fn map_iceberg_error(error: iceberg::Error) -> DataflowError {
             let msg = if let Some(catalog_msg) = extract_catalog_message(&display) {
                 format!("Iceberg catalog error: {catalog_msg}")
             } else {
-                format!("Iceberg catalog error: {}", error.message())
+                format!("Iceberg catalog error: {}", error.to_string())
```
QnJ1c2kNCg: nit: I think the [`Display`](https://doc.rust-lang.org/std/fmt/trait.Display.html) trait would automatically call the `.to_string()` function.

PR #1027, opened 2026-03-11
Address non-idempotent recovery for initiated multiparts

mwylde:
This PR fixes a bug in the filesystem v2 sink which could cause panics on recovery due to an attempt to perform one or more final put part calls on an already completed multipart. The fix is to add logic to the final put parts that checks if the file has already been completed with the correct size, which is the same logic we already had for complete multipart.

In more detail:

The error occurs when we have a a multipart upload in the MultipartStarted phase with buffered data that is being finalized in a commit. At this point, we need to write the final bytes of the file as one or more final parts, and then finalize it. If we fail after the multipart has been finalized but before the commit is acknowledged, on the next recovery we will re-attempt to upload final parts, but this will fail because the multipart is already finalized and cannot be modified. However, we have already written the correct data, just not recorded that we have, so the correct thing to do is, well, nothing. The fix accomplishes that by checking that the file does in fact exist and has the correct size.

This change has been tested by injecting failures into the commit phase at the right point to trigger the issue and validating that we recover correctly.

### DIFF
--- a/crates/arroyo-connectors/src/filesystem/sink/v2/open_file.rs
+++ b/crates/arroyo-connectors/src/filesystem/sink/v2/open_file.rs
@@ -4,8 +4,9 @@ use crate::filesystem::sink::map_storage_error;
 use crate::filesystem::sink::two_phase_committer::CommitStrategy;
 use crate::filesystem::sink::v2::SinkConfig;
 use crate::filesystem::sink::v2::uploads::{
-    FsResponseData, UploadFuture, create_multipart_finalize_future, create_multipart_init_future,
-    create_part_upload_future, create_single_file_upload_future,
+    FsResponseData, UploadFuture, create_last_parts_upload_future,
+    create_multipart_finalize_future, create_multipart_init_future, create_part_upload_future,
+    create_single_file_upload_future,
 };
 use crate::filesystem::sink::{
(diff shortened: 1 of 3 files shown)


crates/arroyo-connectors/src/filesystem/sink/v2/open_file.rs
```diff
@@ -740,41 +765,30 @@ impl<BBW: BatchBufferingWriter + 'static> OpenFile<BBW> {
...
-                multipart_id.clone(),
-                parts.len(),
-                bytes,
-                self.logger.clone(),
-            )]
+            vec![bytes]
         };
 
-        self.stats.parts_written += futures.len();
+        self.stats.parts_written += part_bytes.len();
```
cmackenzie1, line 772: I am a little confused here, before it was tracking number of parts but now its length of bytes?

cmackenzie1: Or maybe the diff is making it harder for me to read

mwylde: The structure of the code has changed a bit. Previously we were building up a list of futures, but now there's always one future, which wraps the one or more final uploads with error handling to address the bug. 

The confusion here is that part_bytes is not a Bytes, but a `Vec<Bytes>`, one per part, so this is capturing the same value as before (the number of parts).

PR #1035, opened 2026-03-30
Introduce worker job controller mode

mwylde:
This PR is part one of the work to introduce the job controller, a worker-hosted component that manages all aspects of ongoing job running such that the data plane no longer has a dependency on the control plane.

The implementation was designed such that the existing job controller approach (running in the controller) should continue to work without significant risk from these changes; in other words, the new worker job controller is meant to be purely additive, such that it should be safe to merge these changes while they are still WIP.

There is a lot of functionality that is not yet implemented, including:

* Metrics
* Preview outputs (still go directly to the controller)
* Checkpoint metadata in the UI
* Tracking of last checkpoint for restoration
* Fencing at commit for split-brain

However, the basics should work at this point.

### DIFF
--- a/crates/arroyo-api/src/metrics.rs
+++ b/crates/arroyo-api/src/metrics.rs
@@ -8,7 +8,7 @@ use arroyo_rpc::api_types::OperatorMetricGroupCollection;
 use arroyo_rpc::api_types::metrics::OperatorMetricGroup;
 use arroyo_rpc::config::config;
 use arroyo_rpc::grpc::rpc::JobMetricsReq;
-use arroyo_rpc::grpc::rpc::controller_grpc_client::ControllerGrpcClient;
+use arroyo_rpc::grpc::rpc::job_controller_grpc_client::JobControllerGrpcClient;
 use tonic::Code;
 use tonic::codec::CompressionEncoding;
 use tracing::error;
@@ -53,7 +53,7 @@ pub async fn get_operator_metric_groups(
         service_unavailable("controller-service")
     })?;
 
-    let mut controller = ControllerGrpcClient::new(channel)
+    let mut controller = JobControllerGrpcClient::new(channel)
         .accept_compressed(CompressionEncoding::Zstd)
         .send_compressed(CompressionEncoding::Zstd);
 
--- a/crates/arroyo-controller/src/job_controller/leader_manager.rs
+++ b/crates/arroyo-controller/src/job_controller/leader_manager.rs
@@ -66,6 +66,14 @@ impl LeaderManager {
             );
         }
 
+        if response.run_id != self.run_id {
+            bail!(
+                "leader returned job status for wrong run: expected {}, got {}",
+                self.run_id,
+                response.run_id
+            );
+        }
+
         let status = response
             .job_status
             .ok_or_else(|| anyhow!("leader returned empty job status"))?;
--- a/crates/arroyo-controller/src/job_controller/leader_manager.rs
+++ b/crates/arroyo-controller/src/job_controller/leader_manager.rs
@@ -1,0 +1,205 @@
+use crate::JobMessage;
+use crate::states::leader_stopping::{LeaderStopBehavior, LeaderStopping};
+use crate::states::recovering::Recovering;
+use crate::states::{JobContext, State, StateError, Transition, TransitionTo, fatal};
+use crate::types::public::StopMode as SqlStopMode;
+use anyhow::{anyhow, bail};
+use arroyo_rpc::config::config;
+use arroyo_rpc::grpc::rpc;
+use arroyo_rpc::grpc::rpc::job_status_grpc_client::JobStatusGrpcClient;
+use arroyo_rpc::grpc::rpc::{JobFailure, JobState, JobStatusReq, JobStopMode, StopJobReq};
+use arroyo_rpc::{job_status_client, retry};
+use arroyo_types::JobId;
+use std::time::{Duration, Instant};
+use tonic::transport::Channel;
+use tracing::{info, warn};
+
+pub struct LeaderManager {
+    leader_client: JobStatusGrpcClient<Channel>,
+    pub job_id: JobId,
+    pub run_id: u64,
+    pub last_heartbeat: Instant,
+}
+
+impl LeaderManager {
+    pub async fn connect(job_id: JobId, run_id: u64, addr: String) -> anyhow::Result<Self> {
+        let leader_client = retry!(
+            job_status_client("controller", &config().worker.tls, addr.clone()).await,
+            5,
+            Duration::from_millis(100),
+            Duration::from_secs(2),
+            |e| warn!(job_id = *job_id.0, message = "failed to connect to worker leader", error = ?e)
+        )?;
+
+        Ok(Self {
+            job_id,
+            run_id,
+            leader_client,
+            last_heartbeat: Instant::now(),
+        })
+    }
+
+    pub async fn poll_leader_status(&mut self) -> anyhow::Result<rpc::JobStatus> {
+        let response = retry!(
+            self.leader_client.get_job_status(JobStatusReq {
+                  job_id: self.job_id.to_string(),
+                }).await,
+                5,
+                Duration::from_millis(100),
+                Duration::from_secs(2),
+                |e| warn!(job_id = *self.job_id.0, message = "failed to poll for job status", error = ?e)
+        )?.into_inner();
+
+        if response.job_id != *self.job_id.0 {
+            bail!(
+                "leader returned job status for wrong job: expected {}, got {}",
+                self.job_id,
+                response.job_id
+            );
+        }
+
+        if response.run_id != self.run_id {
+            bail!(
+                "leader returned job status for wrong run: expected {}, got {}",
+                self.run_id,
+                response.run_id
+            );
+        }
+
+        let status = response
+            .job_status
+            .ok_or_else(|| anyhow!("leader returned empty job status"))?;
+
+        self.last_heartbeat = Instant::now();
+
+        Ok(status)
+    }
+
+    pub async fn stop_leader(&mut self, stop_mode: JobStopMode) -> anyhow::Result<()> {
+        info!(
+            message = "sending stop request to leader",
+            job_id = *self.job_id.0,
+            stop_mode = ?stop_mode,
+        );
+
+        self.leader_client
+            .stop_job(StopJobReq {
+                stop_mode: stop_mode as i32,
+            })
+            .await?;
+
+        Ok(())
+    }
+
+    pub async fn wait_for_state(&mut self, expected: rpc::JobState) -> anyhow::Result<()> {
+        let mut timer = tokio::time::interval(Duration::from_millis(200));
+        loop {
+            let status = self.poll_leader_status().await?;
+
+            let state = JobState::try_from(status.job_state)
+                .map_err(|e| anyhow!("received invalid job state from leader: {e}"))?;
+
+            if state == expected {
+                return Ok(());
+            }
+
+            match state {
+                JobState::JobUnknown => bail!("received unknown job status"),
+                JobState::JobInitializing
+                | JobState::JobRunning
+                | JobState::JobStopping
+                | JobState::JobFinishing
+                | JobState::JobFailing => {
+                    // non-terminal states, continue waiting
+                }
+                JobState::JobStopped | JobState::JobFinished | JobState::JobFailed => {
+                    bail!(
+                        "reached unexpected terminal state {:?} while waiting for {:?}",
+                        state,
+                        expected
+                    );
+                }
+            }
+
+            timer.tick().await;
+        }
+    }
+}
+
+pub async fn handle_leader_stopping<'a, S, T>(
+    state: S,
+    ctx: &mut JobContext<'a>,
+    expected_state: rpc::JobState,
+    next: T,
+    timeout: Option<Duration>,
+) -> Result<Transition, StateError>
+where
+    S: State,
+    T: State,
+    S: TransitionTo<T>,
+    S: TransitionTo<LeaderStopping>,
+    S: TransitionTo<Recovering>,
+{
+    let started = Instant::now();
+
+    loop {
+        let timeout = timeout
+            .map(|t| (started + t).saturating_duration_since(Instant::now()))
+            .unwrap_or(Duration::MAX);
+
+        tokio::select! {
+            msg = ctx.rx.recv() => {
+                match msg {
+                    Some(JobMessage::ConfigUpdate(c)) => {
+                        let next = match c.stop_mode {
+                            SqlStopMode::force => {
+                                Some(LeaderStopBehavior::StopWorkers)
+                            }
+                            SqlStopMode::immediate => {
+                                Some(LeaderStopBehavior::StopJob(JobStopMode::JobStopImmediate))
+                            }
+                            SqlStopMode::graceful => {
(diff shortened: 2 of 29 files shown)


crates/arroyo-api/src/metrics.rs
```diff
@@ -53,7 +53,7 @@ pub async fn get_operator_metric_groups(
         service_unavailable("controller-service")
     })?;
 
-    let mut controller = ControllerGrpcClient::new(channel)
+    let mut controller = JobControllerGrpcClient::new(channel)
```
QnJ1c2kNCg, line 56: nit: I wonder if we should avoid `JobController` to reduce confusion with the Arroyo Controller. `JobManager`?

crates/arroyo-controller/src/job_controller/leader_manager.rs
```diff
@@ -0,0 +1,205 @@
...
+
+        Ok(Self {
+            job_id,
+            run_id,
+            leader_client,
+            last_heartbeat: Instant::now(),
+        })
+    }
+
+    pub async fn poll_leader_status(&mut self) -> anyhow::Result<rpc::JobStatus> {
```
QnJ1c2kNCg, line 42: Should we also validate `worker_id`? Ideally `worker_id` would be validated on every RPC.

Here I worry about the case where a job_id/run_id pair is loaded on two workers.

crates/arroyo-controller/src/job_controller/leader_manager.rs
```diff
@@ -0,0 +1,205 @@
...
+        let status = response
+            .job_status
+            .ok_or_else(|| anyhow!("leader returned empty job status"))?;
+
+        self.last_heartbeat = Instant::now();
+
+        Ok(status)
+    }
+
+    pub async fn stop_leader(&mut self, stop_mode: JobStopMode) -> anyhow::Result<()> {
```
QnJ1c2kNCg, line 86: Should a generation of leader have a unique ID associated with it?

This could be useful to make sure we stop the correct leadership, and maybe avoid ABA type problems.

crates/arroyo-controller/src/job_controller/leader_manager.rs
```diff
@@ -0,0 +1,205 @@
...
+
+            let state = JobState::try_from(status.job_state)
+                .map_err(|e| anyhow!("received invalid job state from leader: {e}"))?;
+
+            if state == expected {
+                return Ok(());
+            }
+
+            match state {
+                JobState::JobUnknown => bail!("received unknown job status"),
```
QnJ1c2kNCg, line 115: What are use cases for `JobUnknown`?

crates/arroyo-controller/src/job_controller/leader_manager.rs
```diff
@@ -0,0 +1,205 @@
...
+                            return Ok(Transition::next(state, LeaderStopping {
+                                stop_behavior
+                            }));
+                        }
+                    }
+                    Some(msg) => {
+                        warn!(job_id = *ctx.config.id, ?msg, "unexpected job message in leader leader mode");
+                    }
+                    None => {
+                        panic!("job queue shut down");
```
QnJ1c2kNCg, line 188: nit: Should we use this opportunity to either reduce our usage of panics or comment all new panics with why this shouldn't happen and/or why we want to panic here.

crates/arroyo-controller/src/states/leader_running.rs
```diff
@@ -0,0 +1,207 @@
...
+use anyhow::anyhow;
+use arroyo_rpc::config::config;
+use arroyo_rpc::grpc::rpc;
+use arroyo_rpc::grpc::rpc::{ErrorDomain, JobFailure, RetryHint};
+use arroyo_rpc::log_event;
+use std::time::{Duration, Instant};
+use tokio::time::MissedTickBehavior;
+use tracing::{error, warn};
+
+const LEADER_POLL_INTERVAL: Duration = Duration::from_millis(500);
```
QnJ1c2kNCg: nit: should this be a config?

crates/arroyo-controller/src/states/mod.rs
```diff
@@ -365,22 +490,67 @@ macro_rules! stop_if_desired_non_running {
...
+            StopMode::force => {
+                return Ok(Transition::next(
+                    *$self,
+                    LeaderStopping {
+                        stop_behavior: LeaderStopBehavior::StopWorkers,
+                    },
+                ));
+            }
+            StopMode::checkpoint => {
+                return Ok(Transition::next(*$self, LeaderCheckpointStopping {}));
```
QnJ1c2kNCg, line 510: (With the goal of potentially simplifying things), is there a reason we need both `LeaderStopping` and `LeaderCheckpointStopping`? Instead of having a `LeaderStopBehavior::Checkpoint`.

crates/arroyo-controller/src/states/mod.rs
```diff
@@ -460,6 +630,76 @@ impl JobContext<'_> {
...
+        &self,
+        state: T,
+        failure: arroyo_rpc::grpc::rpc::JobFailure,
+    ) -> Result<Transition, StateError> {
+        let error_domain = errors::ErrorDomain::from(failure.error_domain());
+        let retry_hint = errors::RetryHint::from(failure.retry_hint());
+        let operator_id = failure
+            .operator_id
+            .clone()
+            .unwrap_or_else(|| "<unknown>".to_string());
```
QnJ1c2kNCg, line 644: Can we make the `operator_id` non optional?

crates/arroyo-controller/src/states/recovering.rs
```diff
@@ -49,44 +51,122 @@ impl Recovering {
...
+                // somehow we got here before scheduling set the job controller / leader manager
+            }
+        };
+
+        // then tear down the workers
+        retry!(
+            Self::tear_down_workers(ctx).await,
+            10,
+            Duration::from_millis(200),
+            Duration::from_secs(10),
```
QnJ1c2kNCg, line 165: Is 10 seconds realistic for tearing down the workers?

crates/arroyo-rpc/src/config.rs
```diff
@@ -269,6 +272,14 @@ pub struct Config {
     pub disable_telemetry: bool,
 }
 
+#[derive(Debug, Deserialize, Serialize, Clone, Default)]
+#[serde(rename_all = "kebab-case", deny_unknown_fields)]
+pub enum JobControllerMode {
+    #[default]
+    Controller,
+    Worker,
```
QnJ1c2kNCg, line 280: Should we embed the `LeaderManager` in `Worker`? `JobControllerMode::Worker(LeaderManager)`?

crates/arroyo-rpc/proto/rpc.proto
```diff
@@ -50,11 +51,12 @@ enum TaskCheckpointEventType {
   FINISHED_COMMIT = 4;
 }
 
+
 message TaskCheckpointEventReq {
   WorkerContext worker_context = 1;
   uint64 time = 2;
   string operator_id = 8;
-  uint32 subtask_idx = 5;
+  uint32 subtask_index = 5;
```
QnJ1c2kNCg: This was recently changed to `subtask_idx` as part of #1033

crates/arroyo-rpc/proto/rpc.proto
```diff
@@ -225,20 +228,81 @@ service ControllerGrpc {
...
+
+message CheckpointStatus {
+  uint32 epoch = 1;
+  uint64 started_at = 2;
+  optional uint64 finished_at = 3;
+  uint64 size_bytes = 4;
+}
+
+enum JobState {
+  JOB_UNKNOWN = 0;
```
QnJ1c2kNCg, line 262: Is this used? Can we do without?

crates/arroyo-controller/src/lib.rs
```diff
@@ -432,15 +328,153 @@ impl ControllerGrpc for ControllerServer {
...
+            },
+        )
+        .await?;
+
+        Ok(Response::new(WorkerInitializationCompleteResp {}))
+    }
+}
+
+#[tonic::async_trait]
+impl JobControllerGrpc for ControllerServer {
```
QnJ1c2kNCg, line 359: Will we be able to remove this impl block once we fully migrate to having the job controller in worker?

crates/arroyo-sql-testing/src/smoke_tests.rs
```diff
@@ -147,31 +150,49 @@ async fn checkpoint(ctx: &mut SmokeTestContext<'_>, epoch: u32) {
...
+                    worker_context: Some(WorkerContext {
+                        machine_id: "test".to_string(),
+                        worker_id: 1,
+                        job_id: (*ctx.job_id).clone(),
+                        run_id: 0,
+                    }),
                     time: to_micros(c.time),
                     operator_id: c.operator_id,
-                    subtask_idx: c.subtask_idx,
+                    subtask_index: c.subtask_idx,
```
QnJ1c2kNCg: nit: `subtask_idx`

crates/arroyo-worker/src/job_controller/model.rs
```diff
@@ -0,0 +1,586 @@
...
+
+        Ok(())
+    }
+
+    pub async fn finish_committing(
+        &self,
+        checkpoint_id: &str,
+        store: &dyn CheckpointMetadataStore,
+    ) -> anyhow::Result<()> {
+        info!("finishing committing");
```
QnJ1c2kNCg: nit: should we log `checkpoint_id` (or other Ids)

crates/arroyo-worker/src/job_controller/model.rs
```diff
@@ -0,0 +1,586 @@
...
+        match msg {
+            RunningMessage::TaskCheckpointEvent(c) => {
+                if let Some(checkpoint_state) = &mut self.checkpoint_state {
+                    if c.epoch != self.epoch {
+                        warn!(
+                            message = "Received checkpoint event for wrong epoch",
+                            epoch = c.epoch,
+                            expected = self.epoch,
+                            job_id = *self.job_id,
+                        );
```
QnJ1c2kNCg, line 140: When would this happen?

crates/arroyo-worker/src/job_controller/model.rs
```diff
@@ -0,0 +1,586 @@
...
+                                } else {
+                                    warn!("unexpected checkpoint event type {:?}", c.event_type())
+                                }
+                            }
+                        };
+                        self.update_db(store).await?;
+                    }
+                } else {
+                    debug!(
+                        message = "Received checkpoint event but not checkpointing",
```
QnJ1c2kNCg, line 161: When would this happen?

crates/arroyo-worker/src/job_controller/model.rs
```diff
@@ -0,0 +1,586 @@
...
+                        subtask_idx = key.1,
+                        reason = event.reason,
+                    );
+                }
+            }
+            RunningMessage::WorkerHeartbeat { worker_id, time } => {
+                if let Some(worker) = self.workers.get_mut(&worker_id) {
+                    worker.last_heartbeat = time;
+                } else {
+                    warn!(
```
QnJ1c2kNCg, line 241: This should never happen right? It would mean that workers from different clusters are talking with eachother?

QnJ1c2kNCg (approved):

crates/arroyo-api/src/metrics.rs
mwylde: Agreed that "controller" has become a bit overloaded. We might want to rename that service at some point. Controller here is in the kubernetes sense, and they're often nested like we have. So there's a PipelineController that creates a Job, which is orchestrated by a JobController.

crates/arroyo-controller/src/job_controller/leader_manager.rs
mwylde: > where a job_id/run_id pair is loaded on two workers

That shouldn't be able to happen (you could have multiple start up, but only one should get the initialization request from the controller). If we do actually have two initialized leaders with the same generation, that would break our commit protocol.

But additional validations wouldn't hurt.

crates/arroyo-controller/src/job_controller/leader_manager.rs
mwylde: See https://protobuf.dev/programming-guides/proto3/#enum-default

crates/arroyo-controller/src/job_controller/leader_manager.rs
mwylde: There aren't really generations of *leaders* — generations are a cluster level concept. However, we should be validating worker ids in all RPCs. I'll do that as a future PR, covering the existing RPCs as well.

crates/arroyo-controller/src/job_controller/leader_manager.rs
mwylde: These are panics because it means that the infrastructure for this job is completely broken somehow (likely because the other side of the queue already panicked). There's nothing we can do from here to recover, so the pattern is to panic ourselves and then the controller will restart the state machine for this pipeline.

crates/arroyo-controller/src/states/mod.rs
mwylde: They could be combined, but the behavior is significantly different so I think it's clearer to have them as different states. For Stopping, we just need the cluster to stop, so we go through a sequence of progressively more forceful approaches. For CheckpointStopping, we need the cluster to stop successfully with a final checkpoint, and if that fails we need to go into recovering.

crates/arroyo-controller/src/states/mod.rs
mwylde: there isn't always an operator id for an error, as some errors are at higher levels than a specific operator

crates/arroyo-controller/src/states/recovering.rs
mwylde: 10 seconds here is the maximum backoff time for our retry with exponential backoff, not a timeout

crates/arroyo-rpc/proto/rpc.proto
mwylde: good catch, must have been a rebase issue as I was splitting up the commits

crates/arroyo-rpc/proto/rpc.proto
mwylde: Discussed above

crates/arroyo-rpc/src/config.rs
mwylde: These are the config definitions, they need to be deserializable from text (json/yaml/toml).

crates/arroyo-controller/src/lib.rs
mwylde: Yes

crates/arroyo-worker/src/job_controller/model.rs
mwylde: This is all existing code (model.rs is basically unchanged, just moved) and was written for the controller version. I think this would look something like a worker resending an old checkpoint complete message.

crates/arroyo-worker/src/job_controller/model.rs
mwylde: See above

crates/arroyo-worker/src/job_controller/model.rs
mwylde: See above, model.rs is existing code from the controller job controller, where we want to be resilient to other workers sending us messages.

PR #1038, opened 2026-04-03
Populate snapshot summary stats in Iceberg commits

garvit-gupta:
#### Context
Iceberg snapshot summaries contain all-zero values for `added-*` and `total-*` stats due to a bug in iceberg-rust 0.6.0 

Fixed upstream in iceberg-rust 0.8.0:
https://github.com/apache/iceberg-rust/pull/1767

#### Changes
Populate correct snapshot summary stats when committing to Iceberg tables.

#### Notes
- `added-*` stats will reflect per-commit values.
- `total-*` stats will equal `added-*` per commit, not cumulative across snapshots. This is due to a separate limitation present in all iceberg-rust versions (0.6.0 through 0.9.0 and current main). These fields are optional per the Iceberg spec and are not used for query correctness.
- This workaround can be removed when iceberg-rust is upgraded to 0.8.0+.

### DIFF
--- a/crates/arroyo-connectors/src/filesystem/sink/iceberg/mod.rs
+++ b/crates/arroyo-connectors/src/filesystem/sink/iceberg/mod.rs
@@ -357,6 +357,15 @@ impl IcebergTable {
             })
             .try_collect()?;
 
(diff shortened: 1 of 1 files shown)


QnJ1c2kNCg (approved):
LGTM!

PR #1043, opened 2026-04-22
Retry transient 409s instead of treating them as user errors

garvit-gupta:
S3-compatible stores returning 409 with body `error code: 1018` should be treated as transient errors. `object_store` maps these to `AlreadyExists`, which is treated as non-retryable, causing permanent pipeline failures on a transient error.

Treat 409s carrying `error code: 1018` as transient in the retry handler and in the filesystem sink's error classification. Other 409s are unchanged, so legitimate conditional-write conflicts still surface as real errors (now and in the future).

### DIFF
--- a/crates/arroyo-connectors/src/filesystem/sink/mod.rs
+++ b/crates/arroyo-connectors/src/filesystem/sink/mod.rs
@@ -167,6 +167,10 @@ fn map_storage_error(storage_error: StorageError) -> DataflowError {
 fn map_object_store_error(obj_err: &object_store::Error) -> DataflowError {
     use object_store::Error;
     match obj_err {
+        // 409s with "error code: 1018" are spurious upstream errors; let the task recover.
+        Error::AlreadyExists { source, .. } if source.to_string().contains("error code: 1018") => {
+            connector_err!(External, WithBackoff, "{}", obj_err)
+        }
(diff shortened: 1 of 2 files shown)


crates/arroyo-storage/src/lib.rs
```diff
@@ -1255,4 +1261,99 @@ mod tests {
                 .unwrap()
         );
     }
+
+    #[test]
+    fn test_should_retry() {
```
mwylde: I'd remove this test, I don't think it provides any real value, just duplicating the code

crates/arroyo-connectors/src/filesystem/sink/mod.rs
```diff
@@ -167,6 +167,10 @@ fn map_storage_error(storage_error: StorageError) -> DataflowError {
 fn map_object_store_error(obj_err: &object_store::Error) -> DataflowError {
     use object_store::Error;
     match obj_err {
+        // 409s with "error code: 1018" are spurious upstream errors; let the task recover.
```
QnJ1c2kNCg, line 170: nit: Should we specify that this is an R2 behavior? (IIRC)

QnJ1c2kNCg (approved):

mwylde (approved):

PR #1045, opened 2026-04-22
Validate target worker_id on WorkerGrpc RPCs

garvit-gupta:
#### Changes
- Each worker-addressed gRPC request now carries a target worker identity. Receiving worker rejects mismatches with PERMISSION_DENIED. This prevents stale gRPC channels from delivering control messages to the wrong worker when network endpoints are recycled across worker lifetimes.
- Uses a tonic Interceptor, so coverage is uniform across all WorkerGrpc RPCs and cannot be bypassed by adding new endpoints.
- Client construction goes through a single factory. Type system enforces interceptor attachment wherever clients are stored.

#### Tests
- Unit tests for identity injection, verification, and round-trip behavior.

### DIFF
--- a/crates/arroyo-rpc/src/identity.rs
+++ b/crates/arroyo-rpc/src/identity.rs
@@ -1,0 +1,141 @@
+//! Identity validation for worker-targeted gRPC requests.
+//!
+//! Every request addressed to a worker carries `x-target-worker-id` in
+//! metadata; the receiving server rejects mismatches with `PermissionDenied`.
+//! This guards against stale channels delivering to the wrong worker when
+//! network endpoints are recycled across worker lifetimes. Worker IDs are
+//! 64-bit random values in production, so they are effectively globally
+//! unique across jobs, runs, and clusters.
+
+use tonic::metadata::{Ascii, MetadataValue};
+use tonic::service::Interceptor;
+use tonic::service::interceptor::InterceptedService;
+use tonic::transport::Channel;
+use tonic::{Request, Status};
+use tracing::warn;
+
+use crate::grpc::rpc::worker_grpc_client::WorkerGrpcClient;
+
+/// Target worker_id header, ASCII decimal.
+pub const WORKER_ID_HEADER: &str = "x-target-worker-id";
+
+/// Worker client with identity interceptor attached.
+pub type WorkerClient = WorkerGrpcClient<InterceptedService<Channel, InjectWorkerId>>;
+
+/// Construct a worker client that injects `target` as `x-target-worker-id`.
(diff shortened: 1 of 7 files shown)


mwylde:
I think this is a reasonable approach, although it has the downside of requiring us to be careful to add worker ids to all messages and checks to all handlers. An alternative approach would be to set the worker id in a header on the caller side and validate it in an interceptor, which would ensure that we cannot miss adding this to new RPCs.

garvit-gupta:
> I think this is a reasonable approach, although it has the downside of requiring us to be careful to add worker ids to all messages and checks to all handlers. An alternative approach would be to set the worker id in a header on the caller side and validate it in an interceptor, which would ensure that we cannot miss adding this to new RPCs.

That makes sense. I have now switched to an interceptor approach. Also expanded the scope to include `JobControllerGrpc` and `JobStatusGrpc`. Changes for them were added as a separate commit though and can be removed if deemed unncessary.

crates/arroyo-rpc/proto/rpc.proto
```diff
@@ -438,6 +438,9 @@ message StartExecutionReq {
   // (used in worker leader mode)
   bool wait_for_leader = 8;
   uint64 checkpoint_interval_micros = 9;
+  // worker_id of the leader when job_controller_addr points to a leader worker;
```
mwylde: Let's exclude worker -> leader communication for now

crates/arroyo-rpc/src/identity.rs
```diff
@@ -0,0 +1,141 @@
...
+
+/// Server interceptor: rejects requests whose `x-target-worker-id` header
+/// does not match the server's own worker_id.
+#[derive(Clone, Debug)]
+pub struct VerifyWorkerId(pub u64);
+
+impl Interceptor for VerifyWorkerId {
+    fn call(&mut self, req: Request<()>) -> Result<Request<()>, Status> {
+        // Missing header is fail-closed: controllers and workers within a run
+        // are always the same version, so there is no upgrade window.
```
mwylde: This comment feels like an artifact of the dev process rather than something that's useful long-term in the code. I would remove it.

crates/arroyo-rpc/src/identity.rs
```diff
@@ -0,0 +1,141 @@
...
+        let received: u64 = req
+            .metadata()
+            .get(WORKER_ID_HEADER)
+            .ok_or_else(|| Status::permission_denied("x-target-worker-id mismatch"))?
```
mwylde: I think this is the wrong default — if the header is omitted we should allow requests through

mwylde (changes requested):

mwylde (approved):

PR #1047, opened 2026-04-22
Fix _connectors_ link in README

QnJ1c2kNCg:


README.md
```diff
@@ -49,7 +49,7 @@ In short: Arroyo lets you ask complex questions of high-volume real-time data wi
 
 🕒 Time-oriented stream processing via the [Dataflow model](https://www.oreilly.com/radar/the-world-beyond-batch-streaming-101/)
 
-🔌 A wide variety of [connectors](https://doc.arroyo.dev/connectors), including Kafka and Iceberg
+🔌 A wide variety of [connectors](https://doc.arroyo.dev/connectors/overview/), including Kafka and Iceberg
```
mwylde, line 52: This is actually a bug on the docs site, not here — looks like this broke when we migrated the docs to astro. I'll look into fixing it there.

QnJ1c2kNCg: I see, my attempt at fixing it based on previous examples: https://github.com/ArroyoSystems/arroyo-docs/pull/121

PR #1051, opened 2026-05-01
Implement new checkpoint protocol

mwylde:
This PR implements a new object store-backed checkpoint metadata protocol (documented [here](https://docs.google.com/document/d/1r3gyseAQF5G1qKsXiQ_F38RrCYEWxRlDFgKcA17TtW0/edit?usp=sharing)), used only in worker-leader mode. Existing deployments using controller leader mode are unaffected by these changes.

This is a large PR with a high degree of complexity. However, most of the changes are limited to a new crate, arroyo-state-protocol, which is the actual protocol implementation. This is self-contained, and implemented as a series of state machines over the underlying protocol types and workflows over those state machines that interact with storage (abstracted over a trait to simplify testing).

The rest of the changes are integration with the protocol within the leader job controller, as well as some necessary infrastructure changes (for example, we now need to pass pipeline id around so that we can use it as part of the checkpoint protocol).

This work is largely complete. There are a few items left as todos:
* GC'ing old checkpoints
* Compaction
* Checkpoint metadata in the API
* Integration tests for leader mode (blocked on API support)
* In new parts of the RPCs we're moving to 64-bit ints for epochs, but I haven't changed existing protocols to avoid breaking things


crates/arroyo-state-protocol/src/workflow.rs
```diff
@@ -0,0 +1,926 @@
...
+    if parent_status(store, &paths, Some(request.checkpoint)).await?
+        == ParentCheckpointStatus::NotReadyCanonical
+    {
+        return Ok(CheckpointPublication::Failed(
+            ResolveFailure::ParentNotReadyCanonical,
+        ));
+    }
+
+    let mut updated_generation_manifest = request.generation_manifest.clone();
+    updated_generation_manifest.latest_checkpoint_ref = Some(request.checkpoint_ref.clone());
```
garvit-gupta, line 571: The generation manifest is being updated before the epoch record is claimed. From what I understand, the resolve logic handles issues in this window correctly, so this shouldn't be a correctness problem. 

But swapping the order could mean the manifest only ever points at claimed checkpoints, which may help eliminate the need for "candidate" semantics? What do you think?

crates/arroyo-worker/src/job_controller/committing_state.rs
```diff
@@ -1,68 +1 @@
-use std::collections::{HashMap, HashSet};
```
garvit-gupta, line 1: nit: Can this file be removed?

garvit-gupta (approved):

crates/arroyo-state-protocol/src/lib.rs
```diff
@@ -0,0 +1,1596 @@
...
+use crate::types::{CheckpointRef, Epoch, Generation};
+use arroyo_types::{JobId, PipelineId};
+
+/// Builds canonical object-store paths for one pipeline/job checkpoint namespace.
+///
+/// All paths are relative to the configured checkpoint storage URI. Callers
+/// should use this rather than formatting protocol paths by hand so that all
+/// readers and writers agree on object names.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct ProtocolPaths {
```
QnJ1c2kNCg, line 22: S3 has a max prefix length of 1024, should have checks here.

crates/arroyo-state-protocol/src/resolve.rs
```diff
@@ -0,0 +1,116 @@
...
+    ReadyCanonical,
+    NotReadyCanonical,
+}
+
+/// Pure decision returned when resolving a generation manifest candidate.
+///
+/// Storage-facing callers usually use `workflow::resolve_generation_manifest`;
+/// this enum is useful for tests and for code that has already read all inputs.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum ResolveDecision {
```
QnJ1c2kNCg, line 21: nit: I think Rust doc would be beneficial here (for most variant)

crates/arroyo-worker/src/job_controller/controller.rs
```diff
@@ -350,13 +597,14 @@ impl WorkerJobController {
...
-            self.cleanup_task = Some(self.start_cleanup(new_epoch));
-        }
+        // TODO: cleaning
+        // if !self.stopping
+        //     && let Some(new_epoch) = self.model.cleanup_needed()
+        //     && self.cleanup_task.is_none()
+        //     && self.model.checkpoint_state.is_none()
+        // {
+        //     self.cleanup_task = Some(self.start_cleanup(new_epoch));
+        // }
```
QnJ1c2kNCg, line 607: Probably should delete or update the TODO

crates/arroyo-worker/src/job_controller/model.rs
```diff
@@ -26,7 +35,9 @@ use tonic::Request;
 use tracing::{debug, error, info, warn};
 
 pub struct RunningJobModel {
```
QnJ1c2kNCg, line 37: That's a lot of fields! Do you think we can logically group them in nested struct? (for instance, group the checkpoint related fields together).

Maybe in a different PR?

crates/arroyo-worker/src/job_controller/model.rs
```diff
@@ -38,6 +49,13 @@ pub struct RunningJobModel {
     pub operator_parallelism: HashMap<u32, usize>,
     pub metric_update_task: Option<JoinHandle<()>>,
     pub last_updated_metrics: Instant,
+    pub protocol_paths: ProtocolPaths,
+    pub checkpoint_parent_ref: Option<CheckpointRef>,
+    pub generation_manifest: Option<GenerationManifest>,
+
+    pub finished_operators: Vec<OperatorCheckpointMetadata>,
+
+    pub worker_leader_mode: bool,
```
QnJ1c2kNCg, line 58: Maybe this becomes an enum, where we can nest the fields that are only used in worker-leader mode

QnJ1c2kNCg (approved):

crates/arroyo-state-protocol/src/workflow.rs
mwylde: Swapping the order doesn't allow us to remove candidate semantics. With the current ordering, we have a potential window where the manifest points at an unclaimed checkpoint, which we resolve on recovery by reading the epoch record. If we swap, then we potentially have a checkpoint that is claimed but does not belong to any generation manifest, which makes recovery much trickier.

crates/arroyo-worker/src/job_controller/controller.rs
mwylde: Left in as we need to implement this as a followup

crates/arroyo-worker/src/job_controller/model.rs
mwylde: I'm not sure I agree that having a lot of fields in a struct is strictly speaking an issue, unless they're frequently passed around as units

crates/arroyo-worker/src/job_controller/model.rs
mwylde: Medium term we'll remove all of the non-worker-leader code, so this was optimized to make that easier rather than building more powerful abstractions over leader mode / controller mode

QnJ1c2kNCg: Make sense

PR #1053, opened 2026-05-06
Add optional state_url and tags to pipelines

QnJ1c2kNCg:
Adds two new optional pipeline-level metadata columns:

  * state_url: per-pipeline override for the worker's checkpoint URL. NULL falls back to controller config().checkpoint_url.
  * tags: a JSONB map of opaque key/value strings, exposed as a PipelineTags newtype on the API surface.

Both are optional on PipelinePost and default to NULL, so existing clients are unaffected.

Refactors the controller's job-startup path to load this data once at boot via a new PipelineInfo struct (carrying pipeline_id, state_url, tags) wrapped in Arc and threaded through JobContext, replacing the prior raw pipeline_pub_id String. The scheduling state now reads state_url from PipelineInfo to populate the worker's ARROYO__CHECKPOINT_URL env var, and forwards tags to the scheduler via StartPipelineReq.pipeline_tags.

There is also quite a bit of plumbing throughout the codebase to pass this new `state_url` around.

This PR also adds a pipeline PUT API that let's the user specify the pipeline ID.

**Testing**
The changes were tested locally, by creating 3 pipelines that have different `state_url`, and making sure that the checkpoint information was going to the right place. I also tested recovery by stopping/starting pipelines.

Pipelines config:
```
❯ rg state_url /tmp/pipeline-*
/tmp/pipeline-a.json
2:  "name": "test_state_url_a",
6:  "state_url": "file:///tmp/arroyo-test/checkpoints-a",

/tmp/pipeline-b.json
2:  "name": "test_state_url_b",
6:  "state_url": "file:///tmp/arroyo-test/checkpoints-b",

/tmp/pipeline-default.json
2:  "name": "test_no_state_url",
```
The DB:
```
SELECT name, state_url, tags
FROM pipelines
WHERE name LIKE 'test_%'
ORDER BY created_at DESC LIMIT 3;"
       name        |               state_url               |        tags
-------------------+---------------------------------------+---------------------
 test_no_state_url |                                       | {"test": "default"}
 test_state_url_b  | file:///tmp/arroyo-test/checkpoints-b | {"test": "b"}
 test_state_url_a  | file:///tmp/arroyo-test/checkpoints-a | {"test": "a"}
```

Also made sure that the new PUT API handled ID conflicts properly:
```
❯ curl -sS -X PUT http://localhost:5115/api/v1/pipelines/test_state_url_b -H 'Content-Type: application/json' -d @/tmp/pipeline-b.json | jq .
{
  "error": "A record already exists with that name or id"
}
```

### DIFF
--- a/crates/arroyo-api/migrations/V28__add_pipeline_tags_and_state_url.sql
+++ b/crates/arroyo-api/migrations/V28__add_pipeline_tags_and_state_url.sql
@@ -1,6 +1,7 @@
--- Optional R2/S3 URL where this pipeline's checkpoints live. NULL falls
+-- Optional URL where this pipeline's checkpoints live. NULL falls
 -- back to the controller's global checkpoint_url config. Existing rows
 -- on upgrade automatically get NULL — no default needed.
+-- See [Storage](https://doc.arroyo.dev/deployment/#storage) for a complete reference.
 ALTER TABLE pipelines ADD COLUMN state_url TEXT;
--- Optional JSON object of the form: { "key" : "value", ... }
-ALTER TABLE pipelines ADD COLUMN tags JSONB;
+-- JSON object of the form: { "key" : "value", ... }. Defaults to an empty map.
+ALTER TABLE pipelines ADD COLUMN tags JSONB NOT NULL DEFAULT '{}'::jsonb;
--- a/crates/arroyo-api/migrations/V28__add_pipeline_tags_and_state_url.sql
+++ b/crates/arroyo-api/migrations/V28__add_pipeline_tags_and_state_url.sql
@@ -1,0 +1,6 @@
+-- Optional R2/S3 URL where this pipeline's checkpoints live. NULL falls
+-- back to the controller's global checkpoint_url config. Existing rows
+-- on upgrade automatically get NULL — no default needed.
+ALTER TABLE pipelines ADD COLUMN state_url TEXT;
+-- Optional JSON object of the form: { "key" : "value", ... }
+ALTER TABLE pipelines ADD COLUMN tags JSONB;
--- a/crates/arroyo-controller/src/lib.rs
+++ b/crates/arroyo-controller/src/lib.rs
@@ -5,6 +5,7 @@
 #![allow(clippy::needless_lifetimes)]
 
 use anyhow::Result;
+use arroyo_rpc::api_types::pipelines::PipelineTags;
 use arroyo_rpc::config::config;
 use arroyo_rpc::grpc::rpc;
 use arroyo_rpc::grpc::rpc::controller_grpc_server::{ControllerGrpc, ControllerGrpcServer};
@@ -25,7 +26,7 @@ use arroyo_rpc::worker_types::{RunningMessage, TaskFailedEvent};
 use arroyo_rpc::{config, errors};
 use arroyo_server_common::shutdown::ShutdownGuard;
 use arroyo_server_common::wrap_start;
-use arroyo_types::{MachineId, WorkerId, from_micros};
+use arroyo_types::{MachineId, PipelineId, WorkerId, from_micros};
 use cornucopia_async::DatabaseSource;
 use states::{Created, State, StateMachine};
 use std::collections::{HashMap, HashSet};
@@ -73,6 +74,14 @@ pub struct JobConfig {
     restart_nonce: i32,
     restart_mode: RestartMode,
     ignore_state_before_epoch: Option<i32>,
+}
+
+/// Per-pipeline data that doesn't change for the lifetime of a job.
+#[derive(Clone, Debug)]
+pub struct PipelineInfo {
+    pub pipeline_id: PipelineId,
+    pub state_url: Option<String>,
+    pub tags: Option<PipelineTags>,
 }
 
 #[derive(Clone, Debug)]
--- a/crates/arroyo-controller/src/lib.rs
+++ b/crates/arroyo-controller/src/lib.rs
@@ -5,7 +5,6 @@
 #![allow(clippy::needless_lifetimes)]
 
 use anyhow::Result;
-use arroyo_rpc::api_types::pipelines::PipelineTags;
 use arroyo_rpc::config::config;
 use arroyo_rpc::grpc::rpc;
 use arroyo_rpc::grpc::rpc::controller_grpc_server::{ControllerGrpc, ControllerGrpcServer};
@@ -81,7 +80,7 @@ pub struct JobConfig {
 pub struct PipelineInfo {
     pub pipeline_id: PipelineId,
     pub state_url: Option<String>,
-    pub tags: Option<PipelineTags>,
+    pub tags: HashMap<String, String>,
 }
 
 #[derive(Clone, Debug)]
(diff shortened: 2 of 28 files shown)


crates/arroyo-api/migrations/V28__add_pipeline_tags_and_state_url.sql
```diff
@@ -0,0 +1,6 @@
+-- Optional R2/S3 URL where this pipeline's checkpoints live. NULL falls
```
mwylde: These are actually broader than r2/s3 — we support a bunch of different URL schemes for various storage locations (https://doc.arroyo.dev/deployment/#storage)

crates/arroyo-state/Cargo.toml
```diff
@@ -29,5 +29,6 @@ prost = {workspace = true}
 prometheus = { workspace = true }
 lazy_static = "1.4.0"
 object_store = { workspace = true }
+moka = { version = "0.12", features = ["future"] }
```
mwylde: Moka is a pretty huge dependency. We already use mini-moka in arroyo-worker, I would see if that can work for our needs before adding moka

crates/arroyo-state/src/lib.rs
```diff
@@ -173,18 +181,105 @@ pub fn hash_key<K: Hash>(key: &K) -> u64 {
...
+
+    fn unique_local_url(suffix: &str) -> String {
+        let nanos = SystemTime::now()
+            .duration_since(UNIX_EPOCH)
+            .unwrap()
+            .as_nanos();
+        format!("file:///tmp/arroyo-state-test-{nanos}-{suffix}")
+    }
+
+    #[tokio::test]
```
mwylde: These tests feel pretty useless, they're basically just testing moka, not any particular logic in our code

crates/arroyo-state/src/lib.rs
```diff
@@ -173,18 +181,105 @@ pub fn hash_key<K: Hash>(key: &K) -> u64 {
...
+/// Workers use [`StorageProviderFor::Worker`], which reads `config().checkpoint_url`
+/// (set per-pipeline via the `ARROYO__CHECKPOINT_URL` env var at worker startup).
+///
+/// Controllers manage many pipelines that may have different storage URLs and
+/// pass [`StorageProviderFor::Controller`] with the pipeline's `state_url`.
+/// When `state_url` is `None`, falls back to `config().checkpoint_url`.
+pub async fn get_storage_provider(
+    role: &StorageProviderFor,
+) -> Result<Arc<StorageProvider>, StorageError> {
+    let storage_url: String = match role {
```
mwylde: we're cloning each URL twice here, but in general we don't need to clone it at all. Can we avoid that?

crates/arroyo-state/src/lib.rs
```diff
@@ -173,18 +181,105 @@ pub fn hash_key<K: Hash>(key: &K) -> u64 {
...
-    tokio::sync::OnceCell::const_new();
+// TODO: Tweak these numbers and maybe introduce a config knob
+const STORAGE_PROVIDER_CACHE_CAPACITY: u64 = 5_000;
+const STORAGE_PROVIDER_CACHE_TTI: Duration = Duration::from_secs(60 * 60);
 
-pub async fn get_storage_provider() -> Result<&'static Arc<StorageProvider>, StorageError> {
-    // TODO: this should be encoded in the config so that the controller doesn't need
-    // to be synchronized with the workers
+/// Per-URL cache of [`StorageProvider`] instances.
+fn storage_provider_cache() -> &'static moka::future::Cache<String, Arc<StorageProvider>> {
```
mwylde: I understand the thinking of having an expiring / bounded cache here, but I'm not sure it's the best approach. In the worker, there will only ever be a single storage provider. In the controller we may have one per pipeline, but we really don't ever want to expire those. I think in practice the number of storage locations will be pretty bounded, and we can get away with just a HashMap behind a mutex

crates/arroyo-rpc/src/api_types/pipelines.rs
```diff
@@ -26,6 +28,18 @@ pub struct PipelinePost {
     pub udfs: Option<Vec<Udf>>,
     pub parallelism: u64,
     pub checkpoint_interval_micros: Option<u64>,
+    pub pub_id: Option<String>,
+    pub state_url: Option<String>,
+    pub tags: Option<PipelineTags>,
+}
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Eq, ToSchema)]
+pub struct PipelineTags(HashMap<String, String>);
```
mwylde: I think newtypes make sense when it helps enforce correctness in types — for example, we have a bunch of newtypes for various ids (PipelineId, JobId, etc.) because we have functions that take in several ids and the newtype prevent us from passing in the wrong one.

I'm less clear on why we need one here, and why we can't just use `HashMap<String, String>`

crates/arroyo-controller/src/lib.rs
```diff
@@ -75,6 +76,14 @@ pub struct JobConfig {
     ignore_state_before_epoch: Option<i32>,
 }
 
+/// Per-pipeline data that doesn't change for the lifetime of a job.
+#[derive(Clone, Debug)]
+pub struct PipelineInfo {
+    pub pipeline_id: PipelineId,
+    pub state_url: Option<String>,
+    pub tags: Option<PipelineTags>,
```
mwylde: With `Option<PipelineTags>` we have two ways of representing "no tags": `None`, and `Some<{}>`. Those seem semantically identical to me, so does it make sense to remove the option?

crates/arroyo-rpc/src/api_types/pipelines.rs
```diff
@@ -26,6 +28,18 @@ pub struct PipelinePost {
     pub udfs: Option<Vec<Udf>>,
     pub parallelism: u64,
     pub checkpoint_interval_micros: Option<u64>,
+    pub pub_id: Option<String>,
```
mwylde: I think this should just be `id`, because this is our public API here, where pub_id is the id

Also, from an API perspective it's a bit weird to have a post with an id field. Generally you would implement this as a PUT on /pipelines/{id}

crates/arroyo-rpc/src/api_types/pipelines.rs
QnJ1c2kNCg: I have a bias for using newtypes almost  everywhere, I feel like it makes things cleaner.

But I don't have a strong opinion here, I'll switch to using `HashMap<String, String>`.

crates/arroyo-state/src/lib.rs
QnJ1c2kNCg: Yeah that's fair. Let's start with that.

crates/arroyo-controller/src/lib.rs
QnJ1c2kNCg: Removed the `Option`, defaulting to an empty `HashMap`.

QnJ1c2kNCg:
Thanks for the review @mwylde. I have addressed all comments.

mwylde (approved):
Look great!

PR #1055, opened 2026-05-11
Support state machine adoption of existing clusters in leader mode

mwylde:
This PR continues the work to enable pipeline clusters (the Arroyo data plane) to operate independently of the controller (control plane). Before we introduced the worker leader mode in #1035, the workers were completely dependent on the controller to operate, and if the controller became unavailable the workers would shutdown. And so the controller, on startup, could just assume that all existing clusters were shut down or could be shut down, and would start over with a new cluster.

However, now that clusters can survive controller downtime, this is no longer desirable behavior. Here, we allow (in leader mode) the controller to adopt a running cluster from a previous controller process back into Running mode, so that there is no interruption to the worker cluster.

In order to do so, we need to store a bit more information, in particular the cluster leader's WorkerId and address, so that we can reconnect to it. This is added a new JSON column in the job_status table.

Currently we will only attempt adoption if the job is in Running. In the future, we may extend this to other states as well to further reduce the disruption of controller restarts.

In addition to the core change, there are a few other improvements:
* We've added worker id verification to the controller -> job status RPCs
* We're now correctly clearing the worker connection cache once we move into recovering
* We've added an option on the process scheduler to allow workers to survive controller (i.e., their parent process) shutdown, to make testing these features locally easier

### DIFF
--- a/crates/arroyo-api/migrations/V29__add_job_state_context.sql
+++ b/crates/arroyo-api/migrations/V29__add_job_state_context.sql
@@ -1,0 +1,1 @@
+ALTER TABLE job_statuses ADD COLUMN state_context JSONB DEFAULT '{"version": 1}' NOT NULL;
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1023,6 +1023,7 @@ dependencies = [
  "bincode",
  "chrono",
  "serde",
+ "tracing",
 ]
 
 [[package]]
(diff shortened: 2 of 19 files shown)


crates/arroyo-api/migrations/V29__add_job_state_context.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE job_statuses ADD COLUMN state_context JSONB DEFAULT '{"version": 1}' NOT NULL;
```
QnJ1c2kNCg, line 1: Are we overloading the word _state_? In other context, state refers to checkpoint info? (ie. `state_url`).

Though I can't think of a better name

QnJ1c2kNCg (approved):

crates/arroyo-api/migrations/V29__add_job_state_context.sql
mwylde: Yeah, it's a bit overloaded (the state here is referring to state machine)

Issue #955, opened 2025-10-27
Postgres database: support TLS/SSL connection settings


1u0:
Arroyo currently doesn't support TLS connection (settings) to a Postgres used as database backend. 
The Postgres connection settings are limited to:
https://github.com/ArroyoSystems/arroyo/blob/613d19443af3c39cdccb1db91b15e21c3ef8a865/crates/arroyo/src/main.rs#L193-L198
https://github.com/ArroyoSystems/arroyo/blob/613d19443af3c39cdccb1db91b15e21c3ef8a865/crates/arroyo/src/main.rs#L331-L337

In some Postgres setups, it might be a requirement to use or the DB forces to use TLS/SSL only connections.
Would be nice to at least support configuring Postgres SSL mode setting.

**Note:** maybe it would be more convenient to allow configuring Postgres connection via URL format (`postgres://...`), with possibility to override some of the parameters separately (mainly to support setting `password` via environment variable).

mwylde:
This was closed in #1058

PR #1058, opened 2026-05-12
Introduce a Postgres TLS config

QnJ1c2kNCg:
Add the ability to have TLS connections to Postgres.

This was tested against a variety of PG configurations.

### DIFF
--- a/crates/arroyo/src/main.rs
+++ b/crates/arroyo/src/main.rs
@@ -17,9 +17,8 @@ use std::process::exit;
 use std::sync::Arc;
 use std::time::Duration;
 use std::{env, fs};
-use tokio::io::{AsyncRead, AsyncWrite};
 use tokio::time::timeout;
-use tokio_postgres::{Client, Connection, NoTls};
+use tokio_postgres::{Client, NoTls};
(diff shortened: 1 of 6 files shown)


crates/arroyo/src/main.rs
```diff
@@ -320,28 +326,39 @@ mod sqlite_migrations {
...
-            Ok(r) => {
-                return Ok(r);
+        let attempt = match &tls_connector {
+            Some(connector) => pg_cfg.connect(connector.clone()).await.map(spawn_conn_task),
+            None => pg_cfg.connect(NoTls).await.map(spawn_conn_task),
+        };
+
+        match attempt {
+            Ok(client) => {
+                // Some PgBouncer configuration strip the `options` startup
```
mwylde: This helps for running migrations, but it seems like all of the other connections to the database at runtime would have the same issue. Is it worth including a partial solution here?

QnJ1c2kNCg: Discussed offline.

Agreed to remove this PgBouncer specific workaround.

mwylde (approved):
Looks great! Also worth noting this addresses #955

PR #1059, opened 2026-05-13
Move common crates to the workspace's Cargo.toml

QnJ1c2kNCg:
All of the crates added to the workspace `Cargo.toml` were present in at least 2 arroyo crates.

I think moving to this model will make it easier to upgrade and ensure that our system is consistent between the various crates.


mwylde (approved):
Nice!

PR #1060, opened 2026-05-14
Introduce `env_vars` column to the job_configs table

QnJ1c2kNCg:
This can be used to pass environment variables to the worker at scheduling time.

**Testing**

Claude made a script that tested the APIs and validated the state of the DB.
```
❯ ./scripts/test_env_vars.sh

▶ Preflight
  ✓ cluster is reachable at http://localhost:5115
  ✓ job_configs.env_vars is JSONB
  latest applied migration: add_job_env_vars

▶ Test 1 — POST /api/v1/pipelines with env_vars
  ✓ POST succeeded — pipeline id = pl_W0d5fOIVQu
  ✓ POST response.env_vars matches input
  value: {"MY_SECRET":"redacted-1","ANOTHER_VAR":"redacted-2"}

▶ Test 2 — env_vars persisted in job_configs
  ✓ DB row env_vars matches input
  value: {"MY_SECRET": "redacted-1", "ANOTHER_VAR": "redacted-2"}

▶ Test 3 — GET /api/v1/pipelines/{id} returns env_vars
  ✓ GET response.env_vars matches input
  value: {"ANOTHER_VAR":"redacted-2","MY_SECRET":"redacted-1"}

▶ Test 4 — POST without env_vars defaults to {}
  ✓ POST response defaults to env_vars = {}
  ✓ DB row defaults to env_vars = {}

▶ Test 5 — controller warns on reserved-key collision
  ✓ POST with reserved key succeeded — pipeline id = pl_VKBoYgm9tU
  ✓ controller emitted collision warning
  2026-05-14T23:18:35.926968Z  WARN arroyo_controller::states::scheduling: job env_vars contained a reserved infrastructure key; user-supplied value has been overridden by the controller job_id="job_Ni0FX0gMk3" key="ARROYO__CHECKPOINT_URL"
  ✓ user-supplied value was NOT logged

─────────────────────────────────────────
All checks passed (11 ok, 0 failed)

Cleaning up test pipelines…
```

### DIFF
--- a/crates/arroyo-api/src/jobs.rs
+++ b/crates/arroyo-api/src/jobs.rs
@@ -1,5 +1,4 @@
 use crate::queries::api_queries::{DbCheckpoint, DbLogMessage, DbPipelineJob};
-use anyhow::anyhow;
 use arroyo_rpc::api_types::checkpoints::{
     Checkpoint, JobCheckpointSpan, OperatorCheckpointGroup, SubtaskCheckpointGroup,
 };
@@ -84,8 +83,8 @@ pub(crate) async fn create_job(
 
     let job_id = generate_id(IdTypes::JobConfig);
 
-    let env_vars_json = serde_json::to_value(&env_vars)
-        .map_err(|e| log_and_map(anyhow!("env_vars serialization failed: {e}")))?;
+    let env_vars_json =
+        serde_json::to_value(&env_vars).expect("HashMap<String, String> is always serializable");
 
     // TODO: handle chance of collision in ids
     api_queries::execute_create_job(
--- a/crates/arroyo-api/src/jobs.rs
+++ b/crates/arroyo-api/src/jobs.rs
@@ -1,4 +1,5 @@
 use crate::queries::api_queries::{DbCheckpoint, DbLogMessage, DbPipelineJob};
+use anyhow::anyhow;
 use arroyo_rpc::api_types::checkpoints::{
     Checkpoint, JobCheckpointSpan, OperatorCheckpointGroup, SubtaskCheckpointGroup,
 };
@@ -44,6 +45,7 @@ pub(crate) async fn create_job(
     preview: bool,
     auth: &AuthData,
     db: &DatabaseSource,
+    env_vars: HashMap<String, String>,
 ) -> Result<String, ErrorResp> {
     let checkpoint_interval = if preview {
         Duration::from_secs(24 * 60 * 60)
@@ -82,6 +84,9 @@ pub(crate) async fn create_job(
 
     let job_id = generate_id(IdTypes::JobConfig);
 
+    let env_vars_json = serde_json::to_value(&env_vars)
+        .map_err(|e| log_and_map(anyhow!("env_vars serialization failed: {e}")))?;
+
     // TODO: handle chance of collision in ids
     api_queries::execute_create_job(
         &db.client().await?,
@@ -96,6 +101,7 @@ pub(crate) async fn create_job(
         } else {
             None
         }),
+        &env_vars_json,
     )
     .await?;
 
--- a/crates/arroyo-api/src/pipelines.rs
+++ b/crates/arroyo-api/src/pipelines.rs
@@ -496,9 +496,6 @@ impl TryInto<Pipeline> for DbPipeline {
             StopMode::force => StopType::Force,
         };
 
-        let env_vars: HashMap<String, String> =
-            serde_json::from_value(self.env_vars).map_err(log_and_map)?;
-
         Ok(Pipeline {
             id: self.pub_id,
             name: self.name,
@@ -512,7 +509,7 @@ impl TryInto<Pipeline> for DbPipeline {
             action_text,
             action_in_progress,
             preview: self.ttl_micros.is_some(),
-            env_vars,
+            env_vars: self.env_vars,
         })
     }
 }
--- a/crates/arroyo-api/src/pipelines.rs
+++ b/crates/arroyo-api/src/pipelines.rs
@@ -305,6 +305,7 @@ pub(crate) async fn create_pipeline_int(
     pub_id: Option<String>,
     state_url: Option<String>,
     tags: HashMap<String, String>,
+    env_vars: HashMap<String, String>,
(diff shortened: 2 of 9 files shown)


crates/arroyo-api/src/pipelines.rs
```diff
@@ -667,6 +673,7 @@ async fn create_pipeline_inner(
         pub_id,
         pipeline_post.state_url,
         pipeline_post.tags.unwrap_or_default(),
+        pipeline_post.env_vars.unwrap_or_default(),
```
QnJ1c2kNCg, line 673: I wasn't sure if we wanted to do input validation. I opted to not for now, but there's a couple of validations that we could add:
* Have a list of reserved keys, for instance `ARROYO__CHECKPOINT_URL` to avoid clashing with our scheduling logic
* Check for length and number of env vars, etc

crates/arroyo-api/src/pipelines.rs
QnJ1c2kNCg, line 754: I didn't update the PATCH API, but maybe I should?

crates/arroyo-api/src/jobs.rs
```diff
@@ -82,6 +84,9 @@ pub(crate) async fn create_job(
 
     let job_id = generate_id(IdTypes::JobConfig);
 
+    let env_vars_json = serde_json::to_value(&env_vars)
```
mwylde: This should never fail, so should be an expect rather than returning an error

crates/arroyo-api/src/pipelines.rs
mwylde: I think it's fine to trust the user here

crates/arroyo-api/src/pipelines.rs
mwylde: This is the pipelines update method, which wouldn't update a job. We don't yet have a job update method, because modifying jobs isn't supported today.

crates/arroyo-controller/src/lib.rs
```diff
@@ -603,6 +605,17 @@ impl ControllerServer {
                 let res = queries::controller_queries::fetch_all_jobs(&client).await?;
                 for p in res {
                     let id = Arc::new(p.id);
+                    let env_vars: HashMap<String, String> =
+                        serde_json::from_value(p.env_vars.clone()).unwrap_or_else(|e| {
+                            error!(
```
mwylde: We probably don't want to error quite like this, as this is run in a loop every second leading to a constant stream of error logs.

It's also likely the wrong thing to do to just ignore the bad env vars — presumably the user set them intentionally, and the pipeline will not work as expected if they are not present, so the correct behavior is to move the pipeline to Failed.

mwylde (changes requested):

crates/arroyo-controller/src/lib.rs
QnJ1c2kNCg: Right make sense. I forgot to leave a comment on the PR asking what would be idiomatic here.

Moving the pipeline to Failed makes sense.

I first tried to move the pipeline to Failed or Failing directly from the updater loop but that seemed awkward (needing to call update_db, etc).

An approach that I think is _ok_ is to store the `env_vars` as a `Result` within the `JobConfig` struct (so we don't fail during the updater loop). During Scheduling, that Result will be `map_err` to a `fatal` error. This approach has the benefit of re-using the existing flow of fatal -> Failing -> Failed.

What do you think? I might have missed something more idiomatic.

mwylde: That could work. Another option could be to keep as a Value here and defer parsing until Scheduling.

QnJ1c2kNCg:
Addressed the comments:
* The `JobConfig` struct now stores the envs as a json `Value`
* We let the Scheduling state do the deserialization and call `fatal` if need be
* The API now returns a json `Value`, that way the caller can see their failed pipelines

**Testing**
I was able to test the error handling by corrupting the JSON in the DB myself.

Screenshot 2026-05-15 at 10 49 41 AM

And we can see the error reason and the invalid json through the APIs:
```
❯ curl -s http://localhost:5115/api/v1/pipelines/$PIPELINE_ID | jq .
{
  "id": "pl_eaxmIf6oUY",
  "name": "env-vars-failure-test",
  "query": "CREATE TABLE src WITH (connector = 'impulse', event_rate = '1', message_count = '1000000'); SELECT counter FROM src;",
  "udfs": [],
  "checkpoint_interval_micros": 10000000,
  "stop": "none",
  "created_at": 1778867299151182,
  "action": null,
  "action_text": "Failed",
  "action_in_progress": false,
  "graph": {
    "nodes": [
      {
        "node_id": 0,
        "operator": "impulse",
        "description": "BoundedImpulse<1 eps>",
        "parallelism": 1
      },
      {
        "node_id": 1,
        "operator": "chained_op",
        "description": "src -> watermark -> sink projection",
        "parallelism": 1
      },
      {
        "node_id": 4,
        "operator": "preview",
        "description": "PreviewSink",
        "parallelism": 1
      }
    ],
    "edges": [
      {
        "src_id": 0,
        "dest_id": 1,
        "key_type": "()",
        "value_type": "()",
        "edge_type": "Forward"
      },
      {
        "src_id": 1,
        "dest_id": 4,
        "key_type": "()",
        "value_type": "()",
        "edge_type": "Forward"
      }
    ]
  },
  "preview": false,
  "env_vars": {
    "TEST_ENV_VAR": 123
  }
}

~/gitlab/arroyo  broy/add-env_vars_to_job_configs [$]
❯ curl -s http://localhost:5115/api/v1/pipelines/$PIPELINE_ID/jobs | jq .
{
  "data": [
    {
      "id": "job_5Gx2dD46Bv",
      "running_desired": true,
      "state": "Failed",
      "run_id": 3,
      "start_time": 1778867299894289,
      "finish_time": 1778867365906956,
      "tasks": 3,
      "failure_reason": {
        "error": "failed to deserialize env_vars from job_configs: invalid type: number, expected a string",
        "domain": "internal"
      },
      "created_at": 1778867299155434
    }
  ]
}
```

mwylde (approved):
Looks good!

PR #1063, opened 2026-05-19
Support checkpoint status API for leader mode

mwylde:
The checkpoint status and details view of the web ui is backed by checkpoint metadata stored in the database. In leader mode, we can no longer store this in the DB, so we need a different approach.

This PR adds a new in-memory checkpoint metadata store in the leader and a new pair of RPC methods to expose the list of checkpoints and details of a particular checkpoint.

These are called by the API server, which looks up the leader information via the state_context field added in #1055.

I've also included a fix to the checkpoint timeline view, which could overlap with the labels if the there were events with times before the checkpoint start time (which can be off by a few ms).

### DIFF
--- a/crates/arroyo-api/src/jobs.rs
+++ b/crates/arroyo-api/src/jobs.rs
@@ -1,6 +1,8 @@
 use crate::queries::api_queries::{DbCheckpoint, DbLogMessage, DbPipelineJob};
+use anyhow::Context;
 use arroyo_rpc::api_types::checkpoints::{
-    Checkpoint, JobCheckpointSpan, OperatorCheckpointGroup, SubtaskCheckpointGroup,
+    Checkpoint, CheckpointEventSpan, JobCheckpointSpan, OperatorCheckpointGroup,
+    SubtaskCheckpointGroup,
 };
 use arroyo_rpc::api_types::pipelines::{JobLogLevel, JobLogMessage, OutputData, StopType};
 use arroyo_rpc::api_types::{
@@ -9,7 +11,7 @@ use arroyo_rpc::api_types::{
 };
 use arroyo_rpc::grpc::api::OperatorCheckpointDetail;
 use arroyo_rpc::public_ids::{IdTypes, generate_id};
-use arroyo_rpc::{get_event_spans, grpc};
+use arroyo_rpc::{LeaderContext, StateContext, get_event_spans, grpc, job_status_client};
 use axum::Json;
 use axum::extract::{Path, Query, State};
 use axum::response::sse::{Event, Sse};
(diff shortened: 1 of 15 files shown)


crates/arroyo-api/src/jobs.rs
```diff
@@ -35,7 +37,101 @@ use crate::{AuthData, queries::api_queries, to_micros, types::public};
...
+            Code::FailedPrecondition | Code::NotFound => bad_request(e.message().to_string()),
+            Code::Unavailable => ErrorResp {
+                status_code: StatusCode::BAD_GATEWAY,
+                message: "the leader is not currently available; try again later".to_string(),
+            },
+            _ => log_and_map(e),
+        })
+}
+
+fn checkpoint_from_leader(checkpoint: grpc::rpc::JobCheckpointMetadata) -> Checkpoint {
```
QnJ1c2kNCg: nit: `From` trait

crates/arroyo-rpc/src/config.rs
```diff
@@ -448,6 +448,9 @@ pub struct WorkerConfig {
     /// TLS configuration for worker TCP shuffling
     #[serde(default)]
     pub tls: Option<TlsConfig>,
+
+    /// Maximum number of checkpoints to keep in history for serving the checkpoint details APIs
+    pub checkpoint_details_to_keep: u32,
```
QnJ1c2kNCg, line 453: This config only applies in leader mode right? Is there a way to factor the config as an enum? ie. if we are leader mode we expect this, otherwise we don't?

crates/arroyo-worker/src/job_controller/mod.rs
```diff
@@ -22,9 +25,157 @@ use tracing::{error, info};
 // Re-export shared types from arroyo-rpc
 pub use arroyo_rpc::worker_types::{RunningMessage, TaskFailedEvent, WorkerContext};
 
-pub const CHECKPOINTS_TO_KEEP: u32 = 4;
+pub const CHECKPOINTS_TO_KEEP: u32 = 10;
```
QnJ1c2kNCg, line 28: Should we deprecate this in favor of the new `checkpoint_details_to_keep` configuration knob?

QnJ1c2kNCg (commented):
Since the checkpoint details are now only stored in-memory we lose a bit of debugability for pipelines that are in a Failed state right? Is that concerning?

crates/arroyo-api/src/jobs.rs
mwylde: Requires moving this to arroyo_rpc where the types are defined, but I think that's ok

crates/arroyo-rpc/src/config.rs
mwylde: I don't think that's desirable for a config. Configs should always be valid to set, whatever mode you're in; otherwise you have to be very careful about how you layer configs across your environments.

crates/arroyo-worker/src/job_controller/mod.rs
mwylde: This is controlling a different knob; this is the number of actual checkpoints (on disk) we're keeping around, which is separate from the number of checkpoints we're storing details for in memory.

The latter knob really exists so that if we are facing memory pressure from storing the details we can disable it easily.

mwylde:
> Since the checkpoint details are now only stored in-memory we lose a bit of debugability for pipelines that are in a Failed state right? Is that concerning?

I don't think this is very concerning; the checkpoint details are most useful to understand behavior of a running pipeline—particularly, why a checkpoint is taking longer than expected. We do also export general checkpoint metrics to understand the historical behavior.

PR #1064, opened 2026-05-21
Introduce `scheduler_config` column to the job_configs table

QnJ1c2kNCg:
Adds a per-job scheduler configuration that overlays the controller's global scheduler config at scheduling time.

This functionality is useful for multi-tenant controllers, where workers associated with different pipelines might want different configurations.
Only the Kubernetes scheduler supports this additional configuration.

Addressed the comments:
* The APIs now take an unstructured JSON payload
* It's the scheduler itself that will attempt to deserialize. Returning a `fatal` error on deser errors.
* Decided to try using figment directly. This gives a large amount of flexibility to the users and also makes the code simpler.

**Testing**
In addition to the new unit tests, I was able to test this on a local Kubernetes cluster.

I had Claude generate me a test script that:
- Waits for the controller to spawn the worker pod, then asserts on the resulting pod spec:
  - Per-job labels merged with chart labels (both present)
  - Per-job annotations merged with chart annotations (both present)
  - SMOKE_OVERRIDE=applied env var set on the worker container
  - ARROYO__WORKER__TASK_SLOTS=2 reflects the override
  - Replica count is ceil(parallelism / task_slots) = 1 pod (proves task_slots reached the scheduler's pod-count math)
- GETs /v1/pipelines/{id}/jobs and asserts scheduler_config round-trips through the API byte-for-byte

Output:
```
❯ ./scripts/test-scheduler-config-k8s.sh
== Preflight ==
✓ kubectl context: docker-desktop
✓ local registry has arroyo-worker and bitnami postgres
✓ containerd registry override present
== Installing helm release arroyo into arroyo-smoke ==
== Waiting for controller to become ready ==
✓ controller ready (took 33s)
== Port-forwarding controller HTTP on :5126 ==
✓ API ping OK
== POST pipeline with scheduler_config overrides ==
✓ pipeline created: pl_IGJspZJro6
== Waiting for worker pod (label smoke-test=overrides) ==
✓ worker pod: arroyo-worker-job-7rimljttyr-1-0
== Asserting per-job overrides reached the pod ==
✓ labels merged correctly (per-job + chart)
✓ annotations merged correctly (per-job + chart)
✓ env var SMOKE_OVERRIDE=applied set on worker container
✓ ARROYO__WORKER__TASK_SLOTS=2 (override applied)
✓ replica count respects task_slots override (1 pod)
== GET /v1/pipelines/pl_IGJspZJro6/jobs round-trip ==
✓ scheduler_config round-trips verbatim through the API
== Waiting for job to reach Running ==
✓ job state=Running
== All K8s smoke assertions passed ==
✓ per-job scheduler_config plumbing works end-to-end on K8s

== Cleanup: uninstalling helm release ==
```

**Example wire format**
```
POST /api/v1/pipelines
Content-Type: application/json
{
  "name": "my-pipeline",
  "query": "...",
  "parallelism": 2,
  "scheduler_config": {
    "worker": {
      "image": "my-registry/custom-worker:v2",
      "image-pull-policy": "Always",
      "task-slots": 2,
      "labels": { "team": "data" },
      "annotations": { "team.arroyo/cost-center": "data-platform" },
      "env": [ { "name": "MY_VAR", "value": "hello" } ]
    }
  }
}
```


crates/arroyo-rpc/src/api_types/jobs.rs
```diff
@@ -0,0 +1,214 @@
...
+    Kubernetes(KubernetesJobSchedulerConfig),
+}
+
+/// Per-job configuration for the Kubernetes scheduler.
+///
+/// All fields are optional: `None` means "use the global default from
+/// `kubernetes-scheduler.worker.*`".
+#[derive(Serialize, Deserialize, Clone, Debug, Default, PartialEq, ToSchema)]
+#[serde(rename_all = "snake_case", deny_unknown_fields)]
+pub struct KubernetesJobSchedulerConfig {
```
QnJ1c2kNCg: It feels slightly awkward to introduce a new struct that is basically a copy of our existing config but with `Option`. Let me know if you have ideas that would be cleaner.

Also, I wasn't sure on the name for this struct.

crates/arroyo-api/migrations/V31__add_job_config_scheduler_config.sql
```diff
@@ -0,0 +1,3 @@
+-- Optional per-job scheduler config.
+-- NULL means "no per-job config — use the controller's global scheduler config".
```
mwylde: I'm a bit confused by this — I'd expect the per-job config to layer on top of the global scheduler config

crates/arroyo-rpc/src/api_types/jobs.rs
```diff
@@ -0,0 +1,214 @@
...
+            ))
+        }
+    }
+}
+
+/// The scheduler-specific portion of a [`SchedulerConfig`]. Internally
+/// tagged on `type`.
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, ToSchema)]
+#[serde(tag = "type", rename_all = "snake_case")]
+pub enum SchedulerConfigSpec {
```
mwylde: I'm not sure this should be an enum. This makes details of particular schedulers part of the public API, and breaks the encapsulation of the scheduler abstraction. To be fair, we already started down this road in config, which might have been a mistake.

I would take the approach of treating these configs as unstructured throughout the API, and have the scheduler interpret them.

crates/arroyo-rpc/src/api_types/jobs.rs
mwylde: Beyond my high-level comment that this should probably be unstructured in the API and handled by the scheduler, I also wonder if it really makes sense to make everything overridable. Realistically, people are going to want to configure just a few of these (likely just image, maybe labels and annotations). 

If we do want everything to be overridable I think we'd want a system that allows us to share the config with the controller-wide config, rather than having each scheduler duplicate it here.

crates/arroyo-controller/src/schedulers/kubernetes/mod.rs
```diff
@@ -38,10 +39,99 @@ impl KubernetesScheduler {
         Self { client, config }
     }
 
-    fn make_pod(&self, req: &StartPipelineReq, number: usize, slots: usize) -> Pod {
-        let c = &self.config;
+    /// Return a [`KubernetesSchedulerConfig`] with the per-job
```
mwylde: This feels like a ton of code and complexity we're adding—a risk with using AI, as it makes it easy to add a ton of code.

Much of this logic is already encapsulated in our figment configuration. If we do want to support everything here, we should leverage what we already have for doing automated config merging rather than manually implementing it for every field.

crates/arroyo-api/src/jobs.rs
```diff
@@ -191,6 +205,27 @@ pub(crate) async fn create_job(
     Ok(job_id)
 }
 
+/// Reject `scheduler_config` payloads that:
+///   1. carry a `version` the current binary doesn't understand, or
+///   2. carry a `type` that doesn't match the controller's active
```
mwylde: This doesn't really work, because the API may not be the controller and may not even have the scheduler config set.

crates/arroyo-controller/src/lib.rs
```diff
@@ -75,6 +75,10 @@ pub struct JobConfig {
     ignore_state_before_epoch: Option<i32>,
     /// Per-job environment variables forwarded to workers at scheduling time.
     env_vars: serde_json::Value,
+    /// Per-job scheduler configuration as raw JSON. `None` means "no
+    /// per-job config — use the controller's global scheduler config
+    /// unchanged".
+    scheduler_config: Option<serde_json::Value>,
```
mwylde: Does it make sense to have None and `{}`? Those seem to me to be the semantics

QnJ1c2kNCg: I removed the `Option`.

crates/arroyo-api/src/jobs.rs
QnJ1c2kNCg: Good point, I forgot that API != Controller.

crates/arroyo-controller/src/schedulers/kubernetes/mod.rs
QnJ1c2kNCg: I updated the PR to use figment. It is indeed cleaner and handles the merge for us.

crates/arroyo-rpc/src/api_types/jobs.rs
QnJ1c2kNCg: Ack. I moved the API to unstructured JSON.

crates/arroyo-api/migrations/V31__add_job_config_scheduler_config.sql
QnJ1c2kNCg: I updated the comment and removed the NULL-ability, hopefully it's clearer now.

crates/arroyo-rpc/src/api_types/jobs.rs
QnJ1c2kNCg: I decided to go with the figment approach. It gives us a high degree of flexibility and the implementation is pretty clean.

It does, however, mean that every field in the configuration is overridable.

QnJ1c2kNCg:
Thanks @mwylde for the review. I've addressed all the comments:
* The APIs now take an unstructured JSON payload
* It's the scheduler itself that will attempt to deserialize. Returning a fatal error on deser errors.
* Decided to try using figment directly. This gives a large amount of flexibility to the users and also makes the code simpler.

mwylde (approved):

PR #1069, opened 2026-05-28
Set finish time on checkpoint details in leader mode when committing

mwylde:
This PR fixes an issue introduced in #1063, where checkpoints where not correctly marked as finished in the details metadata when they included a commit step. This didn't affect the actual checkpoints, but did lead to the webui showing checkpoints failing to finish.

In addition to that fix, this also includes some improvements to the checkpoint details page in the webui, which now caches checkpoint details that the user has viewed so that they don't disappear when trimmed in history or when the leader shuts down.

---

<a href="https://app.devin.ai/review/arroyosystems/arroyo/pull/1069" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/crates/arroyo-worker/src/job_controller/committing_state.rs
+++ b/crates/arroyo-worker/src/job_controller/committing_state.rs
@@ -7,7 +7,7 @@ pub enum CheckpointIdOrRef {
     // used by database-backed mode, checkpoint metadata v0
     CheckpointId(String),
     // used by checkpoint metadata v1
-    CheckpointRef(CommitPermit, String),
+    CheckpointIdAndRef(String, CommitPermit),
 }
 
 pub struct CommittingState {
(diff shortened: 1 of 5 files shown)


crates/arroyo-worker/src/job_controller/committing_state.rs
```diff
@@ -7,7 +7,7 @@ pub enum CheckpointIdOrRef {
     // used by database-backed mode, checkpoint metadata v0
     CheckpointId(String),
     // used by checkpoint metadata v1
-    CheckpointRef(CommitPermit),
+    CheckpointRef(CommitPermit, String),
```
QnJ1c2kNCg: nit: I don't find it super obvious that the `String` here is also the ID. Maybe we keep we rename to something like `CheckpointIdAndRef(String, CommitPermit)`

QnJ1c2kNCg (approved):

PR #1072, opened 2026-05-28
Keep leader alive after checkpoint stopping

mwylde:
There is currently a race condition for checkpoint stopping in leader mode:
1. The controller sends `CheckpointReq { then_stop: true, ... }` to the leader
2. The leader initiates final checkpoint then monitors progress
3. The checkpoint completes and the leader shuts down
4. The controller polls for leader status (waiting for JobStopped) but this fails, because the leader has already shut down

The fix in this PR is to not have the leader shut down on final checkpoint; it hangs around until it gets an actual stop request from the controller. We're also no longer triggering a fatal error if final checkpointing fails, so that we are able to recover the pipeline from the last successful checkpoint.

Also includes a new "manual" scheduler which is useful for testing these sorts of scenarios. 

---

<a href="https://app.devin.ai/review/arroyosystems/arroyo/pull/1072" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/crates/arroyo-worker/src/lib.rs
+++ b/crates/arroyo-worker/src/lib.rs
@@ -1015,17 +1015,32 @@ impl WorkerGrpc for WorkerServer {
         &self,
         _request: Request<JobFinishedReq>,
     ) -> Result<Response<JobFinishedResp>, Status> {
+        let is_worker_leader = self.state.job_controller_tx.get().is_some();
+
         let mut phase = self.state.phase.lock().unwrap();
         if let WorkerExecutionPhase::Running(engine_state) = &*phase {
             engine_state.shutdown_guard.cancel();
         }
         *phase = WorkerExecutionPhase::Idle;
+        drop(phase);
 
-        let token = self.shutdown_guard.token();
-        tokio::task::spawn(async move {
-            tokio::time::sleep(Duration::from_secs(1)).await;
-            token.cancel();
-        });
+        if is_worker_leader {
+            // Keep the worker leader reachable so the controller can observe the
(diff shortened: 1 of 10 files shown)


crates/arroyo-worker/src/lib.rs
```diff
@@ -1015,17 +1015,32 @@ impl WorkerGrpc for WorkerServer {
...
+        if is_worker_leader {
+            // Keep the worker leader reachable so the controller can observe the
+            // terminal job status before explicitly stopping the worker process.
+            info!(
+                message =
+                    "job finished on worker leader; waiting for controller to stop worker process",
+                worker_id = self.state.worker_context.worker_id.0,
+                job_id = *self.state.worker_context.job_id,
+                generation = self.state.worker_context.generation,
+            );
```
QnJ1c2kNCg, line 1036: Are there concerns if the controller never sends the stop message? I'm aware that we have a fencing mechanism in the checkpointing protocol so I don't think there is a risk, but I wanted to double check your thinking around this.

(Would it be prudent to move the worker to an explicit state here?)

mwylde: The controller should always send a stop worker command to the scheduler before starting a new cluster, which in a container environment will delete the actual container.

That may not happen or may not work, which will lead to orphaned resources. That's a risk that existed before this change (since the process stopping doesn't delete the container anyways), but in any case just leads to wasted resources and not any actual impact.

The solution to orphaned resources is an asynchronous process that finds them and cleans them up.

PR #1090, opened 2026-07-01
Support updating `env_vars` and `scheduler_config` on pipelines

cmackenzie1:
In #1060 and #1064, support for specifying per-pipeline `env_vars` and `scheduler_config` was added to the pipeline creation endpoint. This commit now allows modifying these properties using `PATCH` and the assumption each pipeline only ever has one job today. The `PATCH` logic does a entire field replacement and does not attempt to merge the provided values with what may already exist for the job. Users should fetch the existing values first from the jobs API and update the value as needed before making the corresponding `PATCH`.

I'll admit this is a little awkward though given `env_vars` and `scheduler_config` belong to specific job instances and not the pipeline.

To test this, I ran `arroyo cluster`, setup connections and then created a Pipeline

```bash
curl --request POST \
--url http://localhost:5115/api/v1/pipelines \
--header 'content-type: application/json' \
--data '{
"name": "test",
"query": "select * from impulse",
"parallelism": 1
}'
```

After the Pipeline was running, I then issued PATCH requests to modify `env_vars` and watched the pipeline transition from Scheduling to Running.

```bash
curl --request PATCH \
--url http://localhost:5115/api/v1/pipelines/pl_bBwbpsJp2T \
--header 'content-type: application/json' \
--data '{
"env_vars": {
"HELLO": "WORLD"
}
}'
```

### DIFF
--- a/crates/arroyo-api/queries/api_queries.sql
+++ b/crates/arroyo-api/queries/api_queries.sql
@@ -165,7 +165,7 @@ WHERE pub_id = :pub_id AND organization_id = :organization_id;
 
 ----------- jobs -----------------------
 
---! update_job(checkpoint_interval_micros?, stop?, parallelism_overrides?)
+--! update_job(checkpoint_interval_micros?, stop?, parallelism_overrides?, env_vars?, scheduler_config?)
 UPDATE job_configs
 SET
    updated_at = :updated_at,
@@ -173,7 +173,9 @@ SET
 
    stop = COALESCE(:stop, stop),
    checkpoint_interval_micros = COALESCE(:checkpoint_interval_micros, checkpoint_interval_micros),
-   parallelism_overrides = COALESCE(:parallelism_overrides, parallelism_overrides)
+   parallelism_overrides = COALESCE(:parallelism_overrides, parallelism_overrides),
+   env_vars = COALESCE(:env_vars, env_vars),
+   scheduler_config = COALESCE(:scheduler_config, scheduler_config)
 WHERE id = :job_id AND organization_id = :organization_id;
 
 --! restart_job(mode, ignore_state_before_epoch?)
--- a/crates/arroyo-controller/src/states/running.rs
+++ b/crates/arroyo-controller/src/states/running.rs
@@ -11,6 +11,7 @@ use crate::states::recovering::Recovering;
 use crate::states::rescaling::Rescaling;
 use crate::states::restarting::Restarting;
 use crate::states::stop_if_desired_running;
+use crate::types::public::RestartMode;
 use crate::{job_controller::ControllerProgress, states::StateError};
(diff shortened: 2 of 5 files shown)


crates/arroyo-api/queries/api_queries.sql
```diff
@@ -165,15 +165,21 @@ WHERE pub_id = :pub_id AND organization_id = :organization_id;
...
    updated_at = :updated_at,
    updated_by = :updated_by,
 
    stop = COALESCE(:stop, stop),
    checkpoint_interval_micros = COALESCE(:checkpoint_interval_micros, checkpoint_interval_micros),
-   parallelism_overrides = COALESCE(:parallelism_overrides, parallelism_overrides)
+   parallelism_overrides = COALESCE(:parallelism_overrides, parallelism_overrides),
+   env_vars = COALESCE(:env_vars, env_vars),
+   scheduler_config = COALESCE(:scheduler_config, scheduler_config),
+   restart_nonce = CASE WHEN :env_vars IS NOT NULL OR :scheduler_config IS NOT NULL
```
cmackenzie1: Bumping this triggers the rescheduling

mwylde: The idea of restart_nonce is that it triggers a restart in a case where no config has changed (we want a pure side-effect), so we need a kind of hack for that. But we don't need that here—the scheduler config has changed, and so we should be reconciling based on that.

In other words, each scheduler has a function F(Config) that produces an outcome O (for kubernetes, for example, this is a set of deployments). If C becomes C', the reconciler should be able to tell that F(C') doesn't match what's currently deployed, and use that to trigger a redeploy with the new config.

cmackenzie1: Moved the detection and restart state transition to the controller state machine.

crates/arroyo-controller/src/states/running.rs
```diff
@@ -59,6 +59,16 @@ impl State for Running {
...
                             }
 
+                            // env_vars and scheduler_config are only applied when workers are
+                            // (re)scheduled, so a change to either while the job is running
+                            // requires a restart to take effect.
+                            if c.scheduler_config != ctx.config.scheduler_config
+                                || c.env_vars != ctx.config.env_vars {
+                                return Ok(Transition::next(*self, Restarting {
+                                    mode: c.restart_mode
+                                }));
```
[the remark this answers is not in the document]
mwylde: The bot is correct — should be safe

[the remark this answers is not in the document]
cmackenzie1: good bot

QnJ1c2kNCg (approved):

mwylde (approved):

PR #1099, opened 2026-07-08
fix(kafka-sink): return error instead of silent Ok on missing commit restore state

kotwal-itpro:
## Problem

The exactly-once Kafka sink's `handle_commit` path silently returned `Ok(())` when `producer_to_complete` was `None` — a state that indicates the operator restarted between `handle_checkpoint` and `handle_commit` and lost the in-memory reference to the transactional producer. In that state the pending Kafka transaction may have already committed on the broker, aborted, or still be open; the sink has no way to tell without querying transaction state, and checkpoint restore during the two-phase-commit window is not yet implemented (as the pre-existing `error!("... Restoring from commit phase not yet implemented")` on that branch already noted).

Silently returning `Ok(())` told the checkpoint coordinator the commit succeeded when its outcome was in fact unknown, breaking the exactly-once contract.

## Fix

Return a `DataflowError::ConnectorError { domain: Internal, retry: NoRetry, ... }` on that branch so the pipeline fails loudly rather than proceeding with unknown consistency. Uses the existing `connector_err!` macro.

## Test

Adds `test_handle_commit_returns_error_when_producer_state_missing` in `kafka/sink/test.rs` that constructs a `KafkaSinkFunc` in `ExactlyOnce { producer_to_complete: None }` mode and asserts `handle_commit` returns `Err`. The test does not require a live Kafka broker because the `None` branch returns before any producer call.

Existing tests unchanged.

---

<a href="https://app.devin.ai/review/arroyosystems/arroyo/pull/1099" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/crates/arroyo-connectors/src/kafka/sink/mod.rs
+++ b/crates/arroyo-connectors/src/kafka/sink/mod.rs
@@ -1,6 +1,7 @@
 use anyhow::Result;
 use std::borrow::Cow;
 
+use arroyo_rpc::connector_err;
 use arroyo_rpc::errors::DataflowResult;
 use arroyo_rpc::grpc::rpc::{GlobalKeyedTableConfig, TableConfig, TableEnum};
 use arroyo_rpc::{CheckpointEvent, ControlResp};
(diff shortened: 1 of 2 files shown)


kotwal-itpro:
Hi maintainers — following up on this PR. The automated review (Devin AI) came back clean, and I've included a unit test covering the missing-producer-state case without requiring a live Kafka broker. Happy to make any adjustments needed — just let me know if there's anything blocking review, or if this should be routed to someone specific. Thanks for your time!

kotwal-itpro:
Hey team, just checking in on this one — rebased against latest master today to keep it current (was 10 commits behind after the recent datafusion/postgres upgrades). CI should re-run.

Happy to make changes if the approach here isn't quite what you'd want — the fix returns an explicit error on missing commit-state instead of the current silent Ok, but I'm open to alternatives (e.g. logging + treating as first-run) if that lines up better with how you think about restore semantics for the exactly-once Kafka sink.

Thanks!

kotwal-itpro:
Rebased onto latest master (was 35 commits behind, including the recent job-controller removal in #1146) to keep this current and mergeable. No changes to the fix itself — diff is identical, just replayed cleanly. CI should re-run on the new head.

PR #1102, opened 2026-07-10
Add bytes-based flushing to source collectors

46bit:
## What

Sources read data in and collect it up into batches. Batches are flushed based on configurable criteria: until now this has been the number of records collected and how long they have been waiting.

This commit adds another criteria for flushing, based on the memory usage of the records. `source-batch-max-bytes`, disabled by default, allows triggering flushing when a certain amount of data has been collected.

## Why

This new feature enables two useful behaviours:

1) Large records: reduce the risk of OOMs when memory-constrained
2) Small records: create larger, more-efficient batches

## How

Memory usage is estimated based on input bytes, rather than by instrumenting Arrow buffer memory.

Ideally we would implement this based on Arrow buffer usage. However, Arrow's `ArrayBuilder` trait doesn't expose memory size and we would need to maintain a lot of code ourselves to make it work. I think it's reasonable to use input bytes as an approximation.

---

<a href="https://app.devin.ai/review/arroyosystems/arroyo/pull/1102" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/crates/arroyo-connectors/src/nexmark/operator.rs
+++ b/crates/arroyo-connectors/src/nexmark/operator.rs
@@ -281,7 +281,9 @@ impl SourceOperator for NexmarkSourceFunc {
             next_event.bid.as_ref().write_into(&mut bid_builder);
             timestamp_builder.append_value(to_nanos(next_event.event_timetamp) as i64);
 
-            if should_flush(records, flush_time) {
+            // Nexmark generates events synthetically with no raw input bytes.
+            // Pass 0 for bytes buffered.
+            if should_flush(records, 0, flush_time) {
                 collector
                     .collect(RecordBatch::try_new(
                         ctx.out_schema.schema.clone(),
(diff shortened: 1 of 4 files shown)


crates/arroyo-rpc/default.toml
```diff
@@ -5,6 +5,7 @@ job-controller = "controller"
 [pipeline]
 source-batch-size = 512
 source-batch-linger = "100ms"
+# source-batch-max-bytes = 10485760  # flush after ~10MB of input; disabled by default
```
mwylde: default.toml is the defaults that we bake into the binary — there isn't a reason to include a commented out value here

mwylde (approved):

crates/arroyo-rpc/default.toml
46bit: Done

mwylde (approved):

PR #1108, opened 2026-07-18
Ensure that we don't send massive error message over gRPC

QnJ1c2kNCg:
The gRPC servers are configured with a max size of 4MB, this patch adds more resiliency for potentially large error message that would bust the gRPC message size limit.

---

<a href="https://app.devin.ai/review/arroyosystems/arroyo/pull/1108" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/crates/arroyo-worker/src/lib.rs
+++ b/crates/arroyo-worker/src/lib.rs
@@ -44,7 +44,7 @@ use tonic::{Request, Response, Status};
 use tracing::{debug, error, info, warn};
 
 use crate::job_controller::controller::WorkerJobController;
-use crate::utils::to_d2;
+use crate::utils::{MAX_TASK_ERROR_FIELD_BYTES, maybe_truncate, to_d2};
 use arroyo_datastream::logical::LogicalProgram;
 use arroyo_planner::physical::new_registry;
 use arroyo_rpc::config::config;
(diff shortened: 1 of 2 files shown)


crates/arroyo-worker/src/lib.rs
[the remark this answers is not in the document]
QnJ1c2kNCg: Fixed

crates/arroyo-worker/src/utils.rs
```diff
@@ -135,3 +136,55 @@ pub async fn to_d2(logical: &LogicalProgram) -> anyhow::Result<String> {
...
+    }
+
+    value.truncate(end);
+    value.push_str(&suffix);
+
+    warn!(
+        "Truncated oversized String from {} bytes to {} bytes: {}",
+        original_bytes,
+        value.len(),
+        value
```
cmackenzie1, line 161: If I read this right, these logs will be ~64kB of unknown data. Is that something we want to include in the log?

cmackenzie1 (approved):

crates/arroyo-worker/src/utils.rs
QnJ1c2kNCg: I decided to log them as they can be useful to understand the root cause. In theory (🤞), we shouldn't really have this truncation happen often, if at all.

mwylde (approved):

PR #1114, opened 2026-07-27
Export job state metrics for Prometheus

QnJ1c2kNCg:
Emit job state Prometheus metrics as part of the controller updater loop.

---

<a href="https://app.devin.ai/review/arroyosystems/arroyo/pull/1114" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/crates/arroyo-controller/src/lib.rs
+++ b/crates/arroyo-controller/src/lib.rs
@@ -28,6 +28,8 @@ use arroyo_server_common::wrap_start;
 use arroyo_types::{MachineId, PipelineId, WorkerId, from_micros};
 use arroyo_worker::job_controller::job_metrics::JobMetrics;
 use cornucopia_async::DatabaseSource;
+use lazy_static::lazy_static;
+use prometheus::{IntGaugeVec, register_int_gauge_vec};
 use states::{Created, State, StateMachine};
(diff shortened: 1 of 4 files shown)


mwylde (approved):

crates/arroyo-controller/src/lib.rs
```diff
@@ -51,6 +53,63 @@ pub mod schedulers;
...
+lazy_static! {
+    static ref JOBS_BY_STATE: IntGaugeVec = register_int_gauge_vec!(
+        "arroyo_controller_jobs",
+        "Current number of jobs by operational state",
+        &["state"]
+    )
+    .unwrap();
+}
+
+fn metric_job_state(
```
cmackenzie1: Is there a specific reason we re-map all states instead of exposing the states themselves? There is a lot of chance for drift between here and what states may actually exist.

QnJ1c2kNCg: No strong reason, I initially thought having larger "buckets" would be nicer. However, I agree about the drift risk and it's easy to create "buckets" in dashboards/alerts.

I'll update the PR.

cmackenzie1 (approved):

PR #1115, opened 2026-07-28
Add gRPC connect timeouts, prevent global locking of job_state

cmackenzie1:
---

<a href="https://app.devin.ai/review/arroyosystems/arroyo/pull/1115" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/crates/arroyo-controller/src/job_controller/leader_manager.rs
+++ b/crates/arroyo-controller/src/job_controller/leader_manager.rs
@@ -18,6 +18,8 @@ use tonic::codegen::InterceptedService;
 use tonic::transport::Channel;
 use tracing::{info, warn};
 
+const LEADER_CONNECT_TIMEOUT: Duration = Duration::from_secs(10);
+
 pub struct LeaderManager {
     leader_client: JobStatusGrpcClient<InterceptedService<Channel, InjectWorkerId>>,
     pub job_id: JobId,
@@ -35,13 +37,23 @@ impl LeaderManager {
         address: String,
     ) -> anyhow::Result<Self> {
         let leader_client = retry!(
-            job_status_client(
-                "controller",
-                &config().worker.tls,
-                worker_id,
-                address.clone()
+            match tokio::time::timeout(
+                LEADER_CONNECT_TIMEOUT,
+                job_status_client(
+                    "controller",
+                    &config().worker.tls,
+                    worker_id,
+                    address.clone(),
+                ),
             )
-            .await,
+            .await
+            {
+                Ok(result) => result,
+                Err(_) => Err(anyhow!(
+                    "timed out connecting to worker leader after {:?}",
+                    LEADER_CONNECT_TIMEOUT
+                )),
+            },
             5,
             Duration::from_millis(100),
             Duration::from_secs(2),
--- a/crates/arroyo-controller/src/job_controller/leader_manager.rs
+++ b/crates/arroyo-controller/src/job_controller/leader_manager.rs
@@ -18,8 +18,6 @@ use tonic::codegen::InterceptedService;
 use tonic::transport::Channel;
 use tracing::{info, warn};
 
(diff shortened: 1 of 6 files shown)


crates/arroyo-rpc/src/config.rs
```diff
@@ -1073,6 +1084,30 @@ mod tests {
         }
     }
 
+    #[test]
+    #[allow(clippy::result_large_err)]
+    fn grpc_connect_timeout_defaults_to_ten_seconds() {
```
mwylde: Do we need these tests? They seem to just be testing the generic config functionality.

crates/arroyo-rpc/src/config.rs
```diff
@@ -193,6 +193,9 @@ fn load_config(paths: &[PathBuf]) -> Figment {
 #[derive(Debug, Deserialize, Serialize, Clone)]
 #[serde(rename_all = "kebab-case", deny_unknown_fields)]
 pub struct Config {
+    /// gRPC client configuration
+    pub grpc: GrpcConfig,
```
mwylde: I would prefer to have this config just on the leader connection, rather than a generic timeout for all gRPC — that has a huge surface area

crates/arroyo-rpc/src/config.rs
```diff
@@ -275,6 +278,13 @@ pub struct Config {
     pub disable_telemetry: bool,
 }
 
+#[derive(Debug, Deserialize, Serialize, Clone)]
+#[serde(rename_all = "kebab-case", deny_unknown_fields)]
+pub struct GrpcConfig {
+    /// Maximum time to establish a gRPC connection
+    pub connect_timeout: HumanReadableDuration,
```
mwylde: This should be optional, so we can preserve the existing behavior of no timeout

crates/arroyo-rpc/default.toml
```diff
@@ -2,6 +2,9 @@ checkpoint-url = "/tmp/arroyo/checkpoints"
 default-checkpoint-interval = "10s"
 job-controller = "controller"
 
+[grpc]
+connect-timeout = "10s"
```
mwylde: I'm not sure we want to change the default here — at least not until we've validated it

crates/arroyo-controller/src/lib.rs
```diff
@@ -651,22 +651,28 @@ impl ControllerServer {
...
-                Ok(())
-            }
-        } else {
-            warn!(message = "Received message for unknown job id", job_id);
-            Err(Status::failed_precondition(format!(
-                "No job with id {job_id}"
-            )))
-        }
+        // Keep per-job backpressure from holding the global job map lock.
+        let tx = {
```
mwylde, line 655: this is good catch

mwylde (changes requested):

crates/arroyo-rpc/src/config.rs
cmackenzie1: Removed

crates/arroyo-rpc/src/config.rs
cmackenzie1: Limited to only leader connects for now

crates/arroyo-controller/src/job_controller/leader_manager.rs
```diff
@@ -18,6 +18,8 @@ use tonic::codegen::InterceptedService;
 use tonic::transport::Channel;
 use tracing::{info, warn};
 
+const LEADER_CONNECT_TIMEOUT: Duration = Duration::from_secs(10);
```
mwylde: Wait, we probably still want this to be configurable? Also, 10 seconds might be too low

cmackenzie1: Made it configurable again, default of None and we can adjust it to find the right value on the deployment.

cmackenzie1:
Replaced with #1117 #1118

PR #1131, opened 2026-08-13
ci: update action versions in ci.yml to current majors

katsugtgz:
Updates the CI workflow to use current action versions.

Changes in .github/workflows/ci.yml:
- actions/checkout@v3 -> actions/checkout@v7 (3 occurrences)
- actions/setup-java@v3 -> actions/setup-java@v6
- actions/setup-python@v5 -> actions/setup-python@v7
- actions/cache@v4 -> actions/cache@v6.1.0

The workflow does not use the pip-install input, which setup-python v7 removed, and the setup-java step uses only distribution and java-version, which v6 keeps, so all bumps are drop-in.

Validation:
- workflow YAML parses after every change
- CI, Semgrep, lint-console, build-rust, and build-console checks pass on the branch

Scope: .github/workflows/ci.yml only. Other workflow files (binaries.yml, docker.yaml, semgrep.yml) still reference older action versions and can be follow-ups if wanted.

Commit author katsugtgz, unsigned.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,7 +36,7 @@ jobs:
         with:
           components: clippy, rustfmt
           toolchain: 1.95
-      - uses: actions/cache@v5
+      - uses: actions/cache@v6.1.0
         with:
           path: |
             ~/.cargo/bin/
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,7 +27,7 @@ jobs:
     steps:
       - name: Check out
         uses: actions/checkout@v7
-      - uses: actions/setup-java@v5
+      - uses: actions/setup-java@v6
         with:
           distribution: "temurin"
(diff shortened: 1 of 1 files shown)


katsugtgz:
Sorry for the extra push here. I completed the remaining version updates in the same workflow file so the migration is consistent, rather than leaving some actions on older majors.

katsugtgz:
Sorry for the churn on this branch. Follow-up to the CI update: actions/cache in ci.yml was still on v5, this pushes it to v6.1.0 to finish the set.

katsugtgz:
Rebased onto current master (branch had fallen behind). Same changes, no new edits. Workflow checks are re-running now.

katsugtgz:
Sorry for the extra churn on this branch. setup-java v6 is the current major and this workflow only uses distribution and java-version inputs, which are unchanged, so the bump is drop-in.

PR #1143, opened 2026-09-02
Prevent single-record batches from `ContextBuffer`

46bit:
`ContextBuffer` has been measuring batch linger times based on its creation time. Once the first linger period elapsed, every record has been getting flushed as its own batch.
    
We could do with more thorough tests around batching behaviour, but for now I've added a minimal test that reproduced the issue.

### DIFF
--- a/crates/arroyo-formats/src/de.rs
+++ b/crates/arroyo-formats/src/de.rs
@@ -39,7 +39,7 @@ struct ContextBuffer {
     /// Accumulated raw input bytes since last flush.
     /// Currently measured as pre-deserialization input size rather than Arrow buffer memory.
     buffered_bytes: usize,
-    created: Instant,
+    buffered_since: Option<Instant>,
 }
 
 impl ContextBuffer {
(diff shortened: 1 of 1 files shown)


crates/arroyo-formats/src/de.rs
```diff
@@ -36,7 +36,7 @@ pub enum FieldValueType<'a> {
 
 struct ContextBuffer {
     buffer: Vec<Box<dyn ArrayBuilder>>,
-    created: Instant,
+    buffered_since: Instant,
```
mwylde: I think what happened here is that originally ContextBuffer was used only once, and consumed by the flush call, then recreated with a new created timestamp. Then at some point we started reusing it and missed updating this timestamp.

So I think the most correct fix here is to have this be Option<Instant> and set to None when we flush, then initialized on the next write. That achieves the same behavior we had before, which is how much time have we been waiting to flush a piece of data, not how long since we last flushed.

mwylde (changes requested):

crates/arroyo-formats/src/de.rs
46bit: Agreed. I didn't want to overcomplicate it but that is a better design.

46bit: Updated

mwylde (approved):

PR #1147, opened 2026-09-10
Introduce connection table versioning

cmackenzie1:
Connections cannot currently be updated. Supporting updates requires preserving the connection definitions used to compile existing pipelines. One possible approach to supporting this is to add immutable connection versions and record the selected version for each pipeline. 

```sql
CREATE TABLE connection_table_versions (
    id BIGSERIAL PRIMARY KEY,
    connection_table_id BIGINT NOT NULL
        REFERENCES connection_tables(id) ON DELETE CASCADE,
    version INTEGER NOT NULL,
    config JSONB,
    schema JSONB,
    created_at TIMESTAMPTZ NOT NULL DEFAULT CURRENT_TIMESTAMP,
    created_by VARCHAR NOT NULL,
    UNIQUE (connection_table_id, version)
);
```

The existing table `connection_tables` will remain as the stable identity of connection, while `connection_table_versions` stores its configuration history. Queries that selected `config` and `schema` from `connection_tables` now get their values from an inner join on the new `connection_table_versions` table. 

```sql
INNER JOIN connection_table_versions
    ON connection_table_versions.connection_table_id = connection_tables.id
    AND connection_table_versions.version = connection_tables.current_version
```

I did consider the approach of putting versions directly in `connection_tables` with a `(pub_id, version)` unique constraint, but that would duplicate identity and metadata across revisions and complicate references to a specific connection independently of a version.

This establishes the database structure and pins newly compiled pipelines to specific connection versions. A later change will expose API operations for creating and selecting new versions.

---

<a href="https://app.devin.ai/review/arroyosystems/arroyo/pull/1147" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-devin-review-dark.svg?v=3">
    Devin Review
  </picture>
</a>


crates/arroyo-api/migrations/V33__connection_table_versions.sql
```diff
@@ -0,0 +1,43 @@
...
+BEGIN
+    INSERT INTO connection_table_versions
+        (connection_table_id, version, config, schema, created_at, created_by)
+    VALUES
+        (NEW.id, NEW.current_version, NEW.config, NEW.schema, NEW.created_at, NEW.created_by);
+    RETURN NEW;
+END;
+$$ LANGUAGE plpgsql;
+
+CREATE TRIGGER connection_tables_initial_version
```
mwylde: Should we be dropping the trigger after the migration?

crates/arroyo-api/migrations/V33__connection_table_versions.sql
```diff
@@ -0,0 +1,43 @@
+ALTER TABLE connection_tables
+ADD COLUMN current_version INTEGER NOT NULL DEFAULT 1;
```
mwylde: Why do we need a current_version field here? Shouldn't it always be the latest?

crates/arroyo-api/migrations/V33__connection_table_versions.sql
```diff
@@ -0,0 +1,43 @@
...
+
+CREATE TABLE connection_table_versions (
+    id BIGSERIAL PRIMARY KEY,
+    connection_table_id BIGINT NOT NULL
+        REFERENCES connection_tables(id) ON DELETE CASCADE,
+    version INTEGER NOT NULL,
+    config JSONB,
```
mwylde, line 6: I'm a bit concerned that we're duplicating the connection table config across this table and connection_tables -- should we drop the overlapping columns from connection_table?

crates/arroyo-api/migrations/V33__connection_table_versions.sql
cmackenzie1: Moved away from triggers and pulled the PATCH API into this PR.

crates/arroyo-api/migrations/V33__connection_table_versions.sql
cmackenzie1: It was to avoid subqueries for MAX(version). Removed this column.

crates/arroyo-api/migrations/V33__connection_table_versions.sql
cmackenzie1: Dropping the old columns now in the migration.

crates/arroyo-api/sqlite_migrations/V11__connection_table_versions.sql
```diff
@@ -0,0 +1,49 @@
...
+ALTER TABLE connection_tables
+DROP COLUMN config;
+
+ALTER TABLE connection_tables
+DROP COLUMN schema;
+
+ALTER TABLE connection_table_pipelines
+ADD COLUMN connection_version INTEGER NOT NULL DEFAULT 1;
+
+CREATE TRIGGER connection_table_pipeline_version_insert
```
mwylde: Do we also want to drop the trigger here?

cmackenzie1: Switched to a create and rename.

mwylde (approved):

PR #1149, opened 2026-09-11
Support Avro format in FileSystem source connector

### LINKED ISSUE #1040: Support reading Avro in FileSystem connector
As per latest code https://github.com/ArroyoSystems/arroyo/blob/master/crates/arroyo-connectors/src/filesystem/source.rs#L281 , current avro format is not supported by FileSystem connector. 

It would be helpful for users who have source data coming in avro format in localFs or HDFS

veenaypatil:
## Summary
- Fills in the `Format::Avro(_) => todo!()` stub in `FileSystemSourceFunc::read_file`
- Reads each Avro file as a single self-describing object container file (OCF) and decodes it via the existing embedded-schema path in `arroyo-formats` (`avro_messages`) with one `collector.deserialize_slice()` call — no changes needed outside this connector
- Gets Avro's built-in block codecs (null/deflate/snappy/zstd) for free via `apache_avro::Reader`

## Design notes / trade-offs
Checkpointing and resume happen at whole-file granularity, unlike JSON (per-line) or Parquet (per-batch): a checkpoint can't land mid-file, and a restart re-reads any file that was in progress from the start. This is correct (no duplicate or lost records), just coarser-grained. A fully streaming per-record Avro decode would need changes to `arroyo-formats` itself, which seemed out of scope for this issue.

## Testing
Built from source and ran with `arroyo run` against an 800k-record dataset written identically to both Avro files (MinIO) and a Kafka topic (each message a self-contained one-record OCF blob, so both sources exercise the same decode path). Compared FileSystem-Avro output against Kafka-Avro output over the overlapping committed range: **exact match, 0 differences across all 18 columns.**

Also verified: multiple files in one source, a deflate-compressed Avro file, an empty (0-record) file, and a truncated/corrupt file — all handled cleanly with no errors or panics.

While testing I ran into a few pre-existing gaps in the shared Avro decode path (unrelated to this change — they reproduce identically on the Kafka Avro source) and a checkpoint/commit behavior for bounded batch pipelines with two-phase-commit sinks. Filing those separately rather than folding them into this PR.

Fixes #1040

---

<a href="https://app.devin.ai/review/arroyosystems/arroyo/pull/1149" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-devin-review-dark.svg?v=3">
    Devin Review
  </picture>
</a>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -629,6 +629,7 @@ dependencies = [
  "tokio-rustls 0.25.0",
  "tokio-stream",
  "tokio-tungstenite",
+ "tokio-util",
  "tonic",
  "tracing",
  "typify 0.0.13",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -101,6 +101,7 @@ thiserror = "2"
 tokio = { version = "1", features = ["full"] }
 tokio-postgres = { version = "0.7", features = ["with-serde_json-1", "with-time-0_3", "with-uuid-1"] }
 tokio-stream = { version = "0.1", features = ["full"] }
+tokio-util = { version = "0.7", features = ["io", "io-util"] }
 toml = "0.8.13"
 tracing = "0.1"
 tracing-subscriber = { version = "0.3.20", features = ["env-filter", "json"] }
--- a/crates/arroyo-connectors/Cargo.toml
+++ b/crates/arroyo-connectors/Cargo.toml
@@ -27,6 +27,7 @@ serde = { workspace = true }
 serde_json = { workspace = true }
 tokio = { workspace = true }
 tokio-stream = { workspace = true }
+tokio-util = { workspace = true }
 typify = { workspace = true }
 prost = { workspace = true }
 tonic = { workspace = true }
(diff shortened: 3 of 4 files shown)


veenaypatil:
Thanks for the review! All three addressed in [6e85a45](https://github.com/ArroyoSystems/arroyo/commit/6e85a452619052e147602f3594c63717346e613d):

- Memory: read_avro_file now streams the object (via tokio_util::io::SyncIoBridge bridging the async byte stream to a sync Read) and decodes on a blocking task that emits bounded 8192-record batches, instead of buffering the whole file + every decoded record.
- Checkpoint blocking: the select! loop now services ctx.control_rx between every batch, not just after a whole file finishes decoding.
- Compression: SourceFileCompressionFormat (gzip/zstd) is now applied to the stream before Avro OCF parsing.

Re-verified after the fix: reran the 800k-record FileSystem-vs-Kafka parity test (still exact match), and added a new gzip-wrapped multi-file test — it now decodes correctly, and the checkpoint-stop landed at exactly 40,960 rows (5 × 8192), confirming checkpoints now interleave mid-file at batch boundaries rather than only between whole files. Empty-file and truncated-file edge cases still pass cleanly.

PR #1163, opened 2026-09-18
Port Iceberg connector to DF54-compatible Iceberg

QnJ1c2kNCg:
Fourth PR in the DataFusion 54 upgrade.
- Upgrade iceberg-rust from 0.6 to 0.10.1.
- Adapt the REST catalog to the new builder and pluggable storage APIs.
- To build the new `RestCatalogBuilder`, we need async. I had a few options, but I decided to go with making our internal trait's `make_operator` async. To me, that seemed like the more idiomatic thing to do.
- Add support for temporary S3 credentials through session tokens.

CI build is expected to fail, but the Iceberg test are passing locally.
```
cargo check -p arroyo-connectors --all-targets
cargo test -p arroyo-connectors iceberg --lib
```

---

<a href="https://app.devin.ai/review/arroyosystems/arroyo/pull/1163" target="_blank"><picture><source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-devin-review-dark.svg?v=4">Devin Review</picture></a>


crates/arroyo-connectors/src/filesystem/sink/iceberg/storage.rs
```diff
@@ -0,0 +1,96 @@
...
+    fn build(&self, config: &StorageConfig) -> Result<Arc<dyn Storage>> {
+        let mut props = config.props().clone();
+        // Set the fallback only after the catalog and user properties have been
+        // merged. Supplying it as a user property would override catalog config.
+        props
+            .entry("s3.path-style-access".into())
+            .or_insert_with(|| "true".into());
+        let config = StorageConfig::from_props(props);
+        Ok(Arc::new(ArroyoStorage {
+            local: OpenDalStorageFactory::Fs.build(&config)?,
```
mwylde, line 27: We don't have to do this now, but would be nice to replace this with our standard object_store based storage now that it's possible to plug in — we currently have an undesirable property that data writes are done via object_store but metadata is done via opendal, leading to subtly different behavior.

mwylde (approved):

crates/arroyo-connectors/src/filesystem/sink/iceberg/storage.rs
QnJ1c2kNCg: I'll note this down!

